PR #7921, opened 2016-05-24
TMS-2944 server: implement csv team invitation

cihangir:
API usage:
- send post request to `-/teams/invite-by-csv`
- content type should be `multipart/form-data`
- form field name is ignored, you can use whatever you want.
- it does not return any actionable response (page refresh is required) in regards of 
  - csv parsing ignores invalid records
  - csv parsing ignores new lines 
  - it ignores already invited users
  - it ignores non-email containing lines
- only the first column is required, others can be empty
- delimiter is "comma" 
- authorization and authentication is handled
- it doesnt have any rate limiting, throttling
- i benchmarked on my MBP with up to 2K emails - after that test timesout.

https://koding.atlassian.net/browse/TMS-2944


sinan:
@cihangir why do we need a page refresh here?

cihangir:
The provided endpoint does not return the processed invites, to be able to list the new invitations, refresh of the list is required since we do not send realtime updates

PR #8338, opened 2016-07-07
Client with es-2015 & webpack

usirin:
Since I will not be working on this anymore, I am opening this PR just as a reference. Somebody can pick it up from where it is, or we can close this.

I just wanted to put it here.

/cc @koding/frontend


usirin:
closing this, i might have some improvements later.

Issue #8471, opened 2016-07-19
Different teams/organizations might have different period of free trials


usirin:
### Overview
We want to be able to provide a way to change the period of free trials if they register/create the team with a promotional code.

### Goals
- Koding admins (mainly marketing/business related people) can create a promotional event.
- This event will have a specific code/token, and amount of time that it extends the trial.
- This token must be present in the team creation process to make use of this extended trial.
- Teams created with this token, will have extended amount of trial.

PR #8538, opened 2016-07-22
Add support for dynamic group plan

usirin:
This is the implementation required for current system to work with dynamic group plans.

I am not sure if this is 100% correct, i might have to revisit what we are doing [here](https://github.com/koding/koding/blob/master/workers/social/lib/social/models/computeproviders/helpers.coffee#L61). Here is the simple flow:
- Save given `config.plan` to group no matter what.
- While reading a group's plan:
  - First try to find it in `teamplans.coffee`, if it exists then return it.
  - Second, try to see if there is a `JGroupPlan` instance, if it exists then return it.
  - Third, return default plan in `teamplans.coffee`

Let me know if i need to do something more.

/cc @szkl @cihangir 

Related with #8471

### DIFF
--- a/servers/lib/server/handlers/createteam.coffee
+++ b/servers/lib/server/handlers/createteam.coffee
@@ -168,6 +168,8 @@ createGroupKallback = (client, req, res, body) ->
       domains
       # username or email of the user, can be already registered or a new one for creation
       username
+      # title of team's plan
+      plan
     } = body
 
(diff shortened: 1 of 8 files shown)


gokmen:
@usirin restarted tests but currently has conflicts with `master` can u check it out?

usirin:
Closing this as we merged this one already in #8628

Issue #8452, opened 2016-07-15
Update design for new 'Build Stack' modal


diclee:

diclee:
Verify passed on sandbox.

PR #8593, opened 2016-07-26
Resource state modal: design updates

alex-ionochkin:
@gokmen, can you please review these changes?
Mostly they are style, markup and text changes according to updated design of the modal

Here is the demo - http://recordit.co/BuEYkoIly3

https://github.com/koding/koding/issues/8452


alex-ionochkin:
I have to close this PR since I can't add new commits here after koding repository became public. I'll create a new one with the same commits from my new fork

PR #8615, opened 2016-07-27
koding: switch to go1.6 / glide

rjeczalik:
PR split into second one containing only code changes for review purpose - https://github.com/koding/koding/pull/8616

Recreated after fork update - https://github.com/koding/koding/pull/8375

PR #8616, opened 2016-07-27
[review] koding: switch to go1.6 / glide

rjeczalik:
parent: https://github.com/koding/koding/pull/8615


go/src/socialapi/workers/oauth/api/api.go
```diff
@@ -1,100 +0,0 @@
-package main
```
cihangir: why did you delete these?

rjeczalik: I did not find any reference to it, thus I assumed it's not being used.

Moreover it does not build - so it could not be used in current form:

```
$ pwd
/Volumes/rjk.io/src/github.com/koding/koding/go/src/socialapi/workers/oauth/api
$ go build .
can't load package: package socialapi/workers/oauth/api: found packages main (api.go) and api (handlers.go) in /Volumes/rjk.io/src/github.com/koding/koding/go/src/socialapi/workers/oauth/api
```

cihangir: that was a wip feature that got de-prioritized and i wanted all the things in the main repo, that is why it is in current state. On the other hand why do we build it if there is not any ref in the repo?

rjeczalik: Glide complained about it when reading imports / updating deps. Instead of fixing build / deps I just removed it as it was not used.

Do you want it back?

fatih: I think best way is to add a build tag, so the source code is still there but it will not pick up by any tools:

```go
// +build wip

package main
```

rjeczalik: https://github.com/koding/koding/pull/8615/commits/62ea7b532da6d63e25af29e760a1329e2a943bfb - already done

Hey @fatih!

rjeczalik:
Closing. Original PR - https://github.com/koding/koding/pull/8633

PR #8635, opened 2016-07-28
computeproviders: make sure next operation is being called

usirin:


usirin:
post mortem of: #8628

PR #8641, opened 2016-07-28
vagrant: add heartbeats / retries

rjeczalik:
When internet connection on user host (e.g. laptop) goes off and
a vagrant.up operation was in progress, eventual result of the
operation may be lost  due to kite ignoring send errors.

For now adding heartbeats for terraformplugin/vagrant <-> klient/vagrant
communication as it is a blanket cover for all networking edge-cases
that are not currently handled with kite.

The proper fix would be in the longer term to make kite guarantee
any successfful call is always sent (e.g. by queueing messages
during outages).

### DIFF
--- a/go/src/koding/klient/vagrant/handlers.go
+++ b/go/src/koding/klient/vagrant/handlers.go
@@ -9,10 +9,12 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"koding/klient/tunnel/tlsproxy/pem"
 
 	"github.com/boltdb/bolt"
+	"github.com/cenkalti/backoff"
 	"github.com/hashicorp/go-multierror"
 	"github.com/koding/kite"
 	"github.com/koding/kite/dnode"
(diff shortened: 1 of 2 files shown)


go/src/koding/klient/vagrant/handlers.go
```diff
@@ -522,8 +549,20 @@ func (h *Handlers) watchCommand(r *kite.Request, filePath string, fn func() (<-c
...
 	}()
 
 	return true, nil
 }
+
+func retry(op func() error) {
+	retry := backoff.NewExponentialBackOff()
+	retry.MaxElapsedTime = 2 * time.Minute
+	retry.MaxInterval = 10 * time.Second
+	retry.Reset()
```
cihangir: i think Reset is not mandatory here

rjeczalik:
You're right, it's not, only needed if InitialInterval was changed.

go/src/koding/klient/vagrant/handlers.go
rjeczalik: You're right, it's not, only needed if InitialInterval was changed.

PR #8592, opened 2016-07-26
klient: migrate dev->prod a klient connected to prod kontrol

rjeczalik:

PR #8658, opened 2016-07-29
klient: migrate dev->prod a klient connected to prod kontrol

rjeczalik:
resent https://github.com/koding/koding/pull/8592


leeola:
LGTM

PR #8659, opened 2016-07-31
Loading animation should be displayed until the KD script is loaded on Koding Utilities page

### LINKED ISSUE #8528: Loading animation should be displayed until the KD script is loaded on Koding Utilities page.
gif
http://recordit.co/vJGGEDh8Gz

hakankaradis:
fixes #8528 
http://recordit.co/x45lIUCmyP
tested with firefox and chrome


client/home/lib/utilities/components/kdcli/container.coffee
```diff
@@ -1,40 +1,28 @@
...
-      else
-        if globals.config.environment in ['dev', 'default', 'sandbox']
-          "export KONTROLURL=#{KodingKontrol.getKontrolUrl()}; curl -sL https://sandbox.kodi.ng/c/d/kd | bash -s #{token}"
-        else "curl -sL https://kodi.ng/c/p/kd | bash -s #{token}"
+    super props
 
-      @setState
-        key : key
-        cmd : cmd
+    @state=
```
usirin: one space before `equal` sign

client/app/lib/components/reactspinner/index.coffee
```diff
@@ -0,0 +1,7 @@
+kd = require 'kd'
+React = require 'kd-react'
+Spinner = require 'react-spinner'
+
+module.exports = class ReactSpinner extends React.Component
```
usirin: Let's rename this component to `Spinner`

client/app/lib/flux/teams/actions.coffee
```diff
@@ -347,6 +348,21 @@ fetchCurrentStateOfApiAccess = ->
   state = team.isApiEnabled is yes
   reactor.dispatch actions.SET_API_ACCESS_STATE, { state }
 
+fetchOtaToken = ->
```
usirin: `loadOtaToken`

gokmen: we need to write this down into somewhere because I'm seeing it's mixed in this file already. https://github.com/koding/koding/pull/8659/files#diff-a4b7c9664d2e74d36a97a8137571b70eR390 Btw, I agree with `loadOtaToken` for this case, because even it's fetching something it's main purpose here to load some state into some store, we're not actually using this method to fetch the data.

usirin: We talked the exact same thing and decided to make all the functions renamed to `load` prefix instead of `fetch` prefix. We will do it in another PR.

PR #8660, opened 2016-07-31
klient: lazy heartbeats

rjeczalik:
This PR makes vagrant.\* commands start sending heartbeats after
the vagrant command execution is successful (in progress).

Also fixes a bug, when heartbeats were stopped right after early return
of the async watchCommand func.

/cc @cihangir


leeola:
LGTM

PR #8669, opened 2016-08-01
terraformer: use configured secret key

rjeczalik:


config/credentials.default.coffee
```diff
@@ -1,3 +1,5 @@
+secret = () -> Math.random().toString(36).slice(-16)
```
cihangir: it is better have this secret a fixed string, since each call the configure will result in different keys for "secrets" and that could break clustering of kite-user services, since each one will have diff secret for jwt

rjeczalik: the rationale for this is that this secrets should be different for different open-source users, so one user won't be able to access different user kloud / terraformer services with default configuration - so it's safe by default; if someone wanted to have it fixed - one would provide his own config

wdyt?

rjeczalik: the rationale for this is that this secrets should be different for different open-source users, so one user won't be able to access different user kloud / terraformer services with default configuration - so it's safe by default; if someone wanted to have it fixed - one would provide his own config

wdyt?

szkl: Let's write it into `generated/` directory.

cihangir: Since credentials.default.coffee would be editted by deployers, i dont think it is necessary and we dont need to worry about. With the same logic, we would need to generate each credential (postgres, mongo username/pass etc) for each deployment. Our default credential config is not safe by default like every other default config.

config/credentials.default.coffee
```diff
@@ -1,6 +1,33 @@
+generate = () -> Math.random().toString(36).slice(-16)
+
+readOrGenerateSecrets = (file) ->
+  fs = require('fs')
+  secrets = {}
+
+  if fs.exists(file)
```
cihangir: prob you wanted to use fs.existsSync

rjeczalik: done

config/credentials.default.coffee
```diff
@@ -1,6 +1,33 @@
...
+      throw err unless err.code is 'EEXIST'
+
+    fs.writeFileSync(file, JSON.stringify(secrets))
+
+  return secrets
+
 module.exports = (options) ->
   kiteHome = "$KONFIG_PROJECTROOT/generated/kite_home/koding"
 
+  secrets = readOrGenerateSecrets("./generated/secrets.json")
```
szkl: I think secrets file should be placed in `../generated`.

rjeczalik: it's relative to PWD when `./configure` was called, using `../generated` would create that dir outside the repo

rjeczalik: changed require -> readFileSync, now it's relative to CWD which is always top level dir of the repo

config/credentials.default.coffee
cihangir: I would vote for removing since deployers will need to manage two files for configuration of their deployment. But if majority of the votes say "lets have it" i would want it to be moved to main.default.coffee and have them passed with `options` into credential.dev.coffee

On the other hand we could make credentials.* files configurable easily since `./configure` passes all command line options to downstream

config/credentials.default.coffee
rjeczalik: I dropped the patch.

config/credentials.default.coffee
rjeczalik: @cihangir Ok, so let's do one thing at once - this PR fixes terraformer, the secret generation we can track in https://github.com/koding/koding/issues/8672.

I've removed the generation from this PR.

Issue #8687, opened 2016-08-02
Replace Nuclear.js with Redux


usirin:
Since we are going through a cleaning phase after solo reduction, it's the best time to make this switch. 
- [x] Introduce redux to clientside. (#8688)
- [ ] Replace all `nuclear-js` modules with [`redux` modules](https://github.com/erikras/ducks-modular-redux).
- [ ] Do not write any more `nuclear-js` code, instead create a redux module.
- [ ] Restructure all global state we are using in clientside into different modules so that those can be part of the `redux` store.
##

PR #8688, opened 2016-08-02
Introduce Redux

usirin:
This PR only introduces the `store` singleton (which is a redux store).

No functionality improvements will be introduced with this PR.


usirin:
Completes the first item in #8687

gokmen:
@usirin I've merged this in missed the `wip` flag, we can revert it if needed but it seems it's ok to keep it as is for now.

gokmen:
@usirin reverted, there was a weird compile issue on client. we can look at that together.

Issue #8549, opened 2016-07-23
kloud: stop vm on klient dial failure  


rjeczalik:
Stop vm if it provisioned successfully, but dialing klient failed. Instead of making the stack build fail.

rjeczalik:
ref: https://github.com/koding/koding/pull/8732#issuecomment-237858620

PR #8732, opened 2016-08-05
kloud: improve dialing klients during stack build

rjeczalik:


rjeczalik:
Decided that stopping vms after failed dial would break the `Installing software` modal that currently is being developed - we won't know whether dial failed only due to connectivity problems and what's state of the provisioning - for that we would need a way to refresh a stack state (`terraform refresh` + `provision-file-exec` which will reinstall klient).

I'm going to drop https://github.com/koding/koding/issues/8549 and instead I'm sending reworked klient dial part of the kloud.

/cc @leeola

leeola:
LGTM

PR #8737, opened 2016-08-05
provider: destroy terraform resources if provisionning failed

### LINKED ISSUE #8550: kloud: destroy stack if build failed
Make it possible to keep the vms if stack failed to provision.

rjeczalik:
Closes #8550.

### DIFF
--- a/go/src/koding/kites/kloud/provider/builder.go
+++ b/go/src/koding/kites/kloud/provider/builder.go
@@ -311,21 +311,25 @@ func (bs *BaseStack) applyAsync(ctx context.Context, req *kloud.ApplyRequest) er
 		return err
 	}
 
-	// at this point resources have been built successfully on the terraform side;
-	// if updating stack or provisioning klient on remote hosts fails,
-	// we need to cleanup the stack
-	defer func() {
-		if err != nil {
-			_, e := tfKite.Destroy(&tf.TerraformRequest{
-				ContentID: contentID,
-				TraceID:   bs.TraceID,
-			})
+	// At this point resources have been built successfully on the terraform side.
+	// If updating stack or provisioning klient on remote hosts fails,
+	// we need to cleanup the stack; if debug: true was set in the stack template
+	// for any of the instance resources, the stack won't be destroyed
+	// so it's possible to troubleshoot it post-mortem.
+	if !bs.Debug {
+		defer func() {
+			if err != nil {
+				_, e := tfKite.Destroy(&tf.TerraformRequest{
+					ContentID: contentID,
+					TraceID:   bs.TraceID,
+				})
 
-			if e != nil {
-				bs.Log.Warning("error destroying stack after build failure: %s", e)
+				if e != nil {
+					bs.Log.Warning("error destroying stack after build failure: %s", e)
+				}
 			}
-		}
-	}()
+		}()
+	}
 
 	bs.Eventer.Push(&eventer.Event{
 		Message:    "Checking VM connections",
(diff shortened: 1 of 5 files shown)


leeola:
LGTM

go/src/koding/kites/kloud/provider/builder.go
```diff
@@ -311,13 +311,33 @@ func (bs *BaseStack) applyAsync(ctx context.Context, req *kloud.ApplyRequest) er
...
 
+	// At this point resources have been built successfully on the terraform side.
+	// If updating stack or provisioning klient on remote hosts fails,
+	// we need to cleanup the stack; if debug: true was set in the stack template
+	// for any of the instance resources, the stack won't be destroyed
+	// so it's possible to troubleshoot it post-mortem.
+	if !bs.Debug {
+		defer func() {
+			if err != nil {
+				_, e := tfKite.Destroy(&tf.TerraformRequest{
```
cihangir: isnt this problematic if the admin is in the progress of incremential updates? how can we decide on behalf of user to destroy the stack template? even terraform does not automatically destroy the created resources and warns the user to manually edit created resources, i see many problems with this approach and advise to revert it. Is there anything that i am missing?

rjeczalik: Terraformer kite does not do incremental updates (if you meant retrying apply if state got partially transitioned to the expected one).

This is best-effort attempt of fixing e.g. leaked vms, when apply partially failed.

At this point I am not sure whether not updating state file on failure was intentional design for terraformer or not. But taking into consideration fact, that we do not store klient connectivity result in a state file [0] the apply was most likely mean to be one-shot. Correct me if I'm wrong.

[0] So the vm resource has a chance to be recreated when klient for whatever reason does not connect to kontrol.

rjeczalik: Context: https://koding.slack.com/archives/bugs/p1468868457000002

cihangir: > At this point I am not sure whether not updating state file on failure was intentional design for terraformer or not. 

It was not intentional AFAIR.

Just checked the code again and after this [commit](https://github.com/koding/koding/commit/3c9ae8c3e0cd7a20a7319d8daa67dd5f990f6240#diff-5ce92a83686bcbe1f62e027cee0f5f82L30) we broke that support.

Do you see any problem on adding it back to system? Could we solve the problem with a proper error message and some UI operations? (eg: your stack build failed but you can retry building it - build button chages to re-build and we can add destroy button next to it?) <- this part is just a thought we can talk with product team

rjeczalik: > Just checked the code again and after this commit we broke that support.

This commit did not break it, as it did not work in the first place - the goroutine was executed before / during terraform process, so when state file was not even created (it failed with `state.tf: no such file or directory` error).

Right now `Try building again` button always creates stack from scratch. Making it rebuild instead (do an incremental apply) should be safe with one caveat: if building stack fails on `checking connectivity` then rebuild would be a dummy operation.

I agree and I lean towardds reverting this and fixing the terraformer to keep the state file on failures.

cihangir: > This commit did not break it, as it did not work in the first place

It seems i forgot the `defer` and didnt realize even at the second look

rjeczalik: 👍 I'll bring it back sir.

PR #8741, opened 2016-08-05
klientctl: subprocess update

leeola:
This allows kd update to execute the new code, during the update
process. Ensuring migrations and etc can be done on immediately on
update.

Related: #8517

### DIFF
--- a/go/src/koding/klientctl/main.go
+++ b/go/src/koding/klientctl/main.go
@@ -261,6 +261,12 @@ func main() {
 					Name:  "force",
 					Usage: "Updates kd & klient to latest available version.",
 				},
+				cli.BoolFlag{
+					Name: "continue",
+					// TODO(leeola): Update to the latest cli package, and use the Hidden flag
+					// for this bool flag.
+					Usage: "An internal flag, not intended for use.",
+				},
 			},
 		},
 		cli.Command{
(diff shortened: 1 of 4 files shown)


leeola:
@rjeczalik ping

also, the tests didn't run on this and a few MRs.. not sure what's up with that. Thoughts?

go/src/koding/klientctl/main.go
```diff
@@ -261,6 +261,12 @@ func main() {
 					Name:  "force",
 					Usage: "Updates kd & klient to latest available version.",
 				},
+				cli.BoolFlag{
+					Name: "continue",
+					// TODO(leeola): Update to the latest cli package, and use the Hidden flag
+					// for this bool flag.
+					Usage: "An internal flag, not intended for use.",
```
rjeczalik: `Internal use only.` - make it consistent with metrics cmd

After `go1.6` branch is merged we should update deps and fix this, the `kd help` output gets hurt by this.

go/src/koding/klientctl/update.go
```diff
@@ -285,3 +309,32 @@ func checkUpdate() (bool, error) {
...
+	}
+
+	return flags
+}
+
+// isStringZeroValue checks if the given string is a zero value in string form,
+// as would be returned by Context.String("flagName").
+func isStringZeroValue(s string) bool {
+	switch s {
+	case "":
```
rjeczalik: you can fold it:

```go
switch s {
case "", "0", "false":
    return true
default:
    return false
}
```

rjeczalik:
@leeola LGTM 👍 Waiting for update of desc for `continue` cmd, since it'd be visible when deployed.

Issue #8647, opened 2016-07-29
Time-out error shouldn't be displayed for stack builds that really need longer times than normal builds.


diclee:
If a user clicks on 'Try Building Again' while the process is still continuing without a problem, that will ruins the process.
We should ping the machine if the build process still continues or something is really broken.

cc/ @gokmen for further details.

![image](https://cloud.githubusercontent.com/assets/8261892/17233402/68b8d4a2-54e4-11e6-973b-c878f860fe7c.png)
note: the above screenshot is from the design document taken in order to give an idea about the modal appearing.
##

gokmen:
@alex-ionochkin for some stack templates this process may take an hour to finalize and with those builds I'm seeing `try building again` which is not correct and may cause to corrupt stack build process. Instead we need to make this based on the connection status, we can try to send `ping` to kite and based on the response we can decide to show that state, and maybe instead of `try building again` we can show `reload page` button instead.

diclee:
Verify passed on latest.

PR #8756, opened 2016-08-08
Resource state modal: timeouts during software installation

alex-ionochkin:
## Description

In case software installation process takes more time than expected (5 min by default or duration specified in `koding`  `buildDuration` section of stack template) we should not show regular build stack error screen with `Try Building Again` button since click on this button can break correct work of stack build. Instead, we can `ping` machine to check if it works fine and let user start using machine looking into the installation state in logs. To achieve it I introduce a new `Timeout` screen with `Start using your VM` button which should be shown when such timeout happens
Also, I added a couple of minor fixes:
- tag name of alien icon is changed to `figure`
- when current machine state is requested in `MachineFlowController`, wrong property `Status` was used to get machine status from response, it was fixed to use correct property `State`
## Motivation and Context

This change is intended to avoid breaking stack installation process when it takes a lot of time 
https://github.com/koding/koding/issues/8647
## How Has This Been Tested?

Manually

http://take.ms/Un3zM
http://recordit.co/q156BtjFVZ (see the beginning and end of the video)
## Types of changes
## Checklist:

### DIFF
--- a/client/app/lib/providers/resourcestatemodal/controllers/timeoutchecker.coffee
+++ b/client/app/lib/providers/resourcestatemodal/controllers/timeoutchecker.coffee
@@ -21,4 +21,8 @@ module.exports = class TimeoutChecker extends kd.Object
     { duration } = @getOptions()
     @timer = kd.utils.wait duration * 1000, @lazyBound 'emit', 'Timeout'
 
-  stop: -> kd.utils.killWait @timer
+
+  stop: ->
+
+    kd.utils.killWait @timer
+    @timer = null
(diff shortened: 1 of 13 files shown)


client/app/lib/providers/resourcestatemodal/controllers/timeoutchecker.coffee
```diff
@@ -21,4 +21,8 @@ module.exports = class TimeoutChecker extends kd.Object
     { duration } = @getOptions()
     @timer = kd.utils.wait duration * 1000, @lazyBound 'emit', 'Timeout'
 
-  stop: -> kd.utils.killWait @timer
```
sinan: `stop: -> @timer = kd.utils.killWait @timer`

sinan:
👍

Issue #8757, opened 2016-08-08
kd: Integrate kdopen into kd


leeola:
- [ ] kdopen integrated into kd
- [ ] If no subs found, use system open command.

PR #8758, opened 2016-08-08
klientctl: Added kdopen

leeola:
Implements item 1 on #8757

### DIFF
--- a/go/src/koding/klientctl/open/command.go
+++ b/go/src/koding/klientctl/open/command.go
@@ -1,0 +1,175 @@
+package open
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"koding/klient/kiteerrortypes"
+	"koding/klientctl/ctlcli"
+	"koding/klientctl/klient"
+	"koding/klientctl/klientctlerrors"
+	"koding/klientctl/util"
+	"strings"
+
+	"github.com/koding/logging"
+)
+
+// CLI options for the given command.
+type Options struct {
+	Debug     bool
+	Filepaths []string
+}
+
+// Init contains various instances required for a Command instance to be initialized.
+type Init struct {
+	Stdout io.Writer
+	Log    logging.Logger
+
+	// The options to use if this struct needs to dial Klient.
+	//
+	// Note! These will be ignored if c.Klient is already defined before Run() is
+	// called.
+	KlientOptions klient.KlientOptions
+
+	// The klient instance this struct will use.
+	Klient interface {
+		LocalOpenFiles(...string) error
+	}
+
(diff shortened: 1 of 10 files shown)


leeola:
@rjeczalik ping

go/src/koding/klientctl/open/command.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (c *Command) Run() (exit int, err error) {
+	defer func() {
+		if err != nil && !c.dontPrintErr {
+			c.Stdout.Printlnf(err.Error())
+		}
+	}()
+
+	// if err := c.handleOptions(); err != nil {
```
rjeczalik: please remove

go/src/koding/klientctl/open/util.go
```diff
@@ -0,0 +1,61 @@
...
+				if err != nil {
+					return err
+				}
+				f.Chmod(perm)
+				f.Close()
+			} else {
+				// If we don't know what to do with the error, it might be
+				// permissions/etc. Return it.
+				return err
+			}
```
rjeczalik: I assume the `paths` are never sorted. If a slice of paths contains one elements that could not be created, the Mkfiles method is not deterministic - it will fail on that element successfully creating different number of files, depending on position of the broken path within the slice.

What if we try to create all the paths instead and don't early return? Also what happens, when path is a directory at line 26, do we care?

Code-style nitpick - make the code to have less branches, improves readability:

```go
		_, err := os.Stat(path)
		if os.IsNotExist(err) {
 			os.MkdirAll(filepath.Dir(path), perm)
 			// The file doesn't exist, create it.
 			f, err := os.Create(path)
 			if err != nil {
 				return err
 			}
 			f.Chmod(perm)
 			f.Close()
 		} else if err != nil {
 			// If we don't know what to do with the error, it might be
 			// permissions/etc. Return it.
 			return err
 		}
```

leeola: > What if we try to create all the paths instead and don't early return? Also what happens, when path is a directory at line 26, do we care?

Imo, it's similar to `MkdirAll`, if a failure is encountered the process stops. No cleanup is done. I'm also unsure of the usefulness of creating all the files, minus any that failed.

@rjeczalik I don't prefer creating all and returning a multi-error, but you'd prefer it to not return early on failures i'll add it. Your call :)

Note: i added the codestyle improvement, appreciated 👍

leeola: @rjeczalik code style fixed btw, if you'd like to change the function let me know, otherwise this is good to merge i believe.

rjeczalik: @leeola it's good 👍

PR #8787, opened 2016-08-11
[WIP] Introduce redux

hakankaradis:
It is going to be updated when it is done.


hakankaradis:
## Motivation
https://github.com/erikras/ducks-modular-redux

To handle our reducers, we will need to structure 
So every reducer will have that information `{actionTypes, actions, reducer}` 
I created `team` reducer under `app/redux/modules` as an example to create convention for our reducers
@usirin

hakankaradis:
## Motivation of Using `Seamless-Immutable` instead of `immutableJS`

It will prevent accidental updates of the store’s data 

With `seamless-immutable` 
We will have no longer `partially` immutable objects
No more converting from `immutable` object to vanilla JS object (fromJS() and toJS())
And one of the most important one, We will have access to object's prototypes 
We will not need to revive the object to use its methods

Here is the immutable team object 
`seamless-immutable` add some methods to data. (the worn ones)
And as you see it is JGroup instance so we will have access to call its methods

PR #8777, opened 2016-08-09
'Rename VM' feature is missing on vm settings modal

hakankaradis:
User will be able to update machine name from virtual machine tab.
## Description

When user focus on any machine name on virtual machine, we will give user ability to update machine label and after user blur input box or hit enter we will update the machine name.
## Motivation and Context

#8643
## How Has This Been Tested?

Go to virtual machine tab, edit any machine name then blur input box or hit enter. You will see updated machine name from both sidebar and the virtual machine tab.
## Screenshots (if appropriate):

http://recordit.co/rIBYZ2eZsL
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #8776, opened 2016-08-09
Replace bant with Webpack

usirin:
Replace bant/browserify with `webpack`. Details will be here **soon**.

It's backwards compatible replacement, so there shouldn't be any changes in build process. All necessary changes (`Makefile`, lots of stuff to get the code working, etc) are already in. It compiles with `node v0.10`.

Let the review begins. 🎊 

/cc @koding/frontend 
## Motivation and Context

Lots of motivation, _but not context yet.._
## How Has This Been Tested?

This PR is open to trigger auto builds.
## Checklist:
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #8796, opened 2016-08-13
stack-editor/editor: dont show Edit Name button if you can't edit stack

hakankaradis:
WIP

### DIFF
--- a/client/app/lib/providers/computecontroller.ui.coffee
+++ b/client/app/lib/providers/computecontroller.ui.coffee
@@ -161,7 +161,7 @@ module.exports = class ComputeControllerUI
       Save       :
         title    : saveButtonTitle
         type     : 'submit'
-        style    : 'solid green medium save-btn'
+        style    : 'solid primary medium save-btn'
         loader   : { color : '#444444' }
         callback : -> @hideLoader()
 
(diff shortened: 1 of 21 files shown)


sinan:
why wip?

hakankaradis:
@sinan I was planning to fix old green button consistency problem in this pr. 
And also fix comments of #8777 
It will be ready soon.

client/app/lib/providers/computecontroller.ui.coffee
```diff
@@ -182,11 +182,15 @@ module.exports = class ComputeControllerUI
 
 
       buttons['Advanced Mode'] =
+        title    : 'Advanced Mode'
         style    : 'solid medium advanced-mode-btn'
         type     : 'button'
         callback : ->
           form.toggleClass 'in-advanced-mode'
-          @toggleClass 'green'
+          if @buttonTitle is 'Advanced Mode'
```
hakankaradis: @sinan 
I think `Advanced Mode` button should be renamed instead of changing its class because it is still secondary action but not sure for `Basic Mode` title

hakankaradis:
Based on this pr #8776 We will not have some folder anymore so I didn't check for old `green` buttons.

PR #8801, opened 2016-08-15
third-party: update koding/terraform

rjeczalik:


cihangir:
is it from https://github.com/koding/terraform or hashicorp/terraform?

rjeczalik:
Still from github.com/koding/terraform - it contains a change that our frontend relies on (removed retries). If that behaviour is kept using upstream only, we could get rid of a fork.

cihangir:
i asked because title was saying from hashicorp/terraform instead of koding/terraform,  updated it.

rjeczalik:
:+1: Thanks!

PR #8816, opened 2016-08-15
klient: removed unix size check

leeola:
Path size checking in KD was using unix tools, blocking windows support.

### DIFF
--- a/go/src/koding/klient/fs/fs.go
+++ b/go/src/koding/klient/fs/fs.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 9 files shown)


go/src/koding/klient/fs/fs.go
```diff
@@ -37,6 +38,14 @@ type ReadDirectoryOptions struct {
 	IgnoreFolders []string
 }
 
+type GetInfoOptions struct {
```
rjeczalik: 👍

go/src/koding/klient/fs/fs.go
```diff
@@ -397,3 +404,22 @@ func GetDiskInfo(r *kite.Request) (interface{}, error) {
 
 	return di, nil
 }
+
+func GetPathSize(r *kite.Request) (interface{}, error) {
+	var opts GetFolderSizeOptions
+
+	if r.Args.One().Unmarshal(&opts) != nil || opts.Path == "" {
```
rjeczalik: Return meaningful error:

```go
if err := r.Args.One().Unmarshal(&opts); err != nil {
    return nil, err
}

if opts.Path == "" {
    return nil, errors.New("invalid empty path")
}
```

rjeczalik:
This LGTM 👍

PR #8853, opened 2016-08-17
Logging in with a JWT

cihangir:
Added a new handler for logging users with JWT
## Description

Add new handler
Add test cases for the handler
Add endpoint in webserver
## Motivation and Context

When people creates team from hubspot we want to log them in automatically
## How Has This Been Tested?

There is integrations test. After hubspot integration is done one should be logged in automatically to their team after creation of the team.
## Types of changes
## Checklist:

## How to use this feature?

Team creation returns a "token" in the response, hit the new endpoint with that token.
`"#{protocol}//#{groupName}.#{host}/-/loginwithtoken?token=#{data.token}"` also you can add optional `redirectTo` query parameter.


sinan:
thx @cihangir, @usirin can you take care of the wiring and frontend part?

usirin:
Will do once this gets merged. Thx @cihangir

PR #8874, opened 2016-08-19
Pricing/Payment Frontend

usirin:


usirin:
It's ready to be code reviewed. I will create a stack in koding team, and will prepare a showcase-like place that we can test the states (expired, expiring, trialing, etc.)

/cc @sibblegp @cihangir

sibblegp:
@usirin I'm going to create a list of things that need to be tested.

gokmen (changes requested):
Requires rebase.

client/app/lib/redux/dispatchInitialActions.coffee
```diff
@@ -0,0 +1,77 @@
...
+
+  { _id } = globals.currentGroup
+
+  state = getState()
+
+  group = bongo.byId('JGroup', _id)(state)
+
+  if group.payment?.subscription
+  then dispatch(loadSubscription())
+  else dispatch(createSubscription(state.customer.id, 'p_up_to_10'))
```
gokmen: better to get this `p_up_to_10` from config/constants

client/app/lib/redux/helper.coffee
```diff
@@ -0,0 +1,45 @@
...
+
+first = (obj) -> _.values(obj)[0]
+
+exports.defineSchema = (name, definitions) ->
+  isArray = Array.isArray name
+
+  [name] = name  if isArray
+
+  rootSchema = new normalizr.Schema name
+  definitions and rootSchema.define definitions
```
gokmen: what is this doing?

client/app/lib/redux/helper.coffee
```diff
@@ -0,0 +1,45 @@
...
+
+  return if isArray then normalizr.arrayOf(rootSchema) else rootSchema
```
gokmen: even it's coming as an `Array` you're getting the first value in it on line 40, then how `rootSchema` makes difference?

client/app/lib/redux/modules/bongo.coffee
```diff
@@ -0,0 +1,77 @@
...
+
+  switch action.type
+    when LOAD.SUCCESS
+      { result } = action
+      result = [result]  unless Array.isArray result
+      result.forEach (res) ->
+        state = state.set res.constructor.name, immutable {}  unless state[res.constructor.name]
+        state = state.update res.constructor.name, (collection) ->
+          collection = collection.set res._id, immutable res
+          return collection
```
gokmen:
```coffee
collection.set res._id, immutable res
```

would be enough

client/app/lib/redux/modules/bongo.coffee
```diff
@@ -0,0 +1,77 @@
...
+      { result } = action
+      result = [result]  unless Array.isArray result
+      result.forEach (res) ->
+        { _id, constructor: { name: constructorName } } = res
+        state = state.set constructorName, state[constructorName].without(_id)
+
+      return state
+
+    else
+      return state
```
gokmen: remove all `return state` and remove `else` part then just put one `return state` at the end.

client/app/lib/redux/modules/payment/creditcard.coffee
```diff
@@ -0,0 +1,72 @@
...
+      if c.default_source
+        return normalized.entities.sources[c.default_source]
+
+      return null
+
+    when REMOVE.SUCCESS, customer.REMOVE.SUCCESS
+      return null
+
+    else
+      return state
```
gokmen: same here no need for `else` just put `return state` and remove other `return`s except the one on line 22

client/app/lib/redux/modules/payment/subscription.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+
+endsAt = (state) ->
+  if sub = state.subscription
+  then Number(sub.current_period_end) * 1000
+  else Date.now()
+
+pricePerSeat = (state) ->
+
+  if p = plan(state) then p.amount / 100 else 0
```
gokmen: code convention,`then` in new line

client/app/lib/redux/modules/payment/subscription.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+trialDays = (state) -> plan(state)?.trial_period_days
```
gokmen: it can be `false` when `plan` not returns anything, will it be ok for the rest of the code?

client/app/lib/util/appendHeadElement.coffee
```diff
@@ -1,7 +1,7 @@
-kd = require 'kd'
-
 module.exports = ({ type, url }, callback) ->
 
+  kd = require 'kd'
+
```
gokmen: why do we need this change?

client/app/lib/util/getGroupStatus.coffee
```diff
@@ -0,0 +1,14 @@
+
+module.exports = getGroupStatus = (group) ->
+
+  # return 'unlimited'  if global.config.environment is 'default'
+
+  return 'no payment'  unless group.payment
+
+  return 'no subscription'  unless group.payment
```
gokmen: logically illogical :) there is no way to get `no subscription`

client/app/lib/util/getGroupStatus.coffee
```diff
@@ -0,0 +1,14 @@
...
+
+  return 'no status'  unless group.payment.subscription
```
gokmen: I think you missed the checks here, I think whole file should be like this;

```coffee
module.exports = getGroupStatus = (group) ->

  # return 'unlimited'  if global.config.environment is 'default'
  return 'no payment'       unless group.payment
  return 'no subscription'  unless group.payment.subscription
  return 'no status'        unless group.payment.subscription.status

  return group.payment.subscription.status

```

client/app/lib/util/isGroupDisabled.coffee
```diff
@@ -0,0 +1,9 @@
+getGroupStatus = require './getGroupStatus'
+
+expiredStatuses = ['expired', 'past_due', 'canceled']
+
+module.exports = isGroupDisabled = (group) ->
+
+  status = getGroupStatus group
```
gokmen: we can do the same for this `status` check again in remote-extension of `JGroup`

client/app/lib/util/makeHttpClient.coffee
```diff
@@ -0,0 +1,50 @@
+{ assign } = require 'lodash'
+axios = require 'axios'
```
gokmen: what is this package?

client/app/lib/util/routeHandler.coffee
```diff
@@ -50,10 +50,51 @@ module.exports = (options) ->
...
+
+        { appManager } = kd.singletons
+
+        return handleDisabled role, status  if frontApp = appManager.getFrontApp()
+
+        appManager.once 'AppIsBeingShown', -> handleDisabled role, status
+
+        appManager.open 'IDE', { forceNew: yes }, (app) ->
+          app.amIHost = yes
+          appManager.tell 'IDE', 'showNoMachineState'
```
gokmen: it might not enough to prevent IDE access, we may need to put this check on kloud and klient kite implementations for front-end.

client/home/lib/billing/components/creditcardcontainer.coffee
```diff
@@ -0,0 +1,123 @@
...
+
+  { nickname } = account.profile
+  [ userId ] = Object.keys(users).filter (id) ->
+    users[id].username is nickname
+
+  return if userId then users[userId] else null
+
+
+accountUser = createSelector(
+  bongo.all('JUser')
```
gokmen: what is this doing? only `JUser` you can fetch on FE for the active `JAccount`, session owner.

client/home/lib/billing/components/subscriptioncontainer.coffee
```diff
@@ -0,0 +1,71 @@
...
+subscriptionTitle = createSelector(
+  subscription.pricePerSeat
+  subscription.isTrial
+  subscription.trialDays
+  teamSize
+  (pricePerSeat, isTrial, trialDays, size) ->
+    if isTrial
+      if trialDays is 7
+      then "Koding Basic Trial (1 Week)"
+      else "Koding Basic Trial (1 Month)"
```
gokmen: maybe we should move all constants to someplace else.

gokmen (changes requested):
Looks good in general but we need to fix the issues addressed in review.

client/app/lib/components/headermessage/container.coffee
```diff
@@ -0,0 +1,133 @@
...
+    expiring: 'Notify My Admin'
+    past_due: 'Notify My Admin'
+
+makeHeaderMessages = (daysLeft) ->
+  admin:
+    expiring:
+      type: 'danger'
+      title: 'Your trial is about to expire.'
+      description: "
+        Your 1 week trial is about to expire. You have only #{daysLeft} days
```
usirin: 👍

client/app/lib/components/headermessage/container.coffee
```diff
@@ -0,0 +1,133 @@
...
+
+
+module.exports = connect(
+  mapStateToProps
+  null
+  mergeProps
+)(HeaderMessage)
+
+
+assign = (args...) -> _.assign {}, args...
```
sinan: why? because typing `_.` is too hard 😄

client/app/lib/providers/computecontroller.coffee
```diff
@@ -78,6 +84,17 @@ module.exports = class ComputeController extends KDController
...
+  bindGroupStatusEvents: ->
+
+    { groupsController } = kd.singletons
+
+    # /cc @cihangir: not sure if this is the right way to bind the event.
+    groupsController.on 'payment_status_changed', ({ oldStatus, newStatus }) =>
+      wasDisabled = @disabled
+      @disabled = isGroupDisabled getGroup()
+
+      do @reset  unless @disabled isnt wasDisabled
```
sinan: `do @reset  if @disabled is wasDisabled` would be more logically logical maybe :)

client/home/lib/billing/components/creditcardcontainer.coffee
usirin: It's only reading from `JUser` values in `bongo` reducer.

client/app/lib/util/makeHttpClient.coffee
usirin: It's an easy, `http-client` library that returns promises for each request, and also has the capability of creating individual instances that is bound to specific requests. for example creating a client with a base url of `/payment` after that each request for that instance will be prefixed with that url, more info is [here](https://github.com/mzabriskie/axios/#creating-an-instance)

client/app/lib/util/appendHeadElement.coffee
usirin: hmm, to be able to mock `appendHeadElement` without mocking `kd`, it's probably not required anymore, but not sure.

client/app/lib/redux/modules/payment/subscription.coffee
usirin: it's gonna be `undefined`, rather than `false`.

only time this can become `undefined` is when there is no `subscription` for a group. And all those cases where we need to write `trialDays` is being guarded by subscription existence checks.

client/app/lib/redux/modules/bongo.coffee
usirin: Yep, probably i made a debugging session to see the result of `collection` and then forgot to remove the explicit return. Will fix.

client/app/lib/components/headermessage/container.coffee
usirin: `_.assign` is mutative, `assign` isn't :)

client/app/lib/util/routeHandler.coffee
usirin: We can do this in the next PR, this one is big enough already.

client/app/lib/redux/modules/payment/creditcard.coffee
usirin: This one is a little bit tricky. Because the removed state itself is `null`. But you're right about `else` statement, i am fixin' it.

usirin:
@gokmen Updated.

gokmen (approved):

sibblegp:
Woooooooo!

PR #8802, opened 2016-08-15
koding: switch to go1.7.1

rjeczalik:
/cc @cihangir

PR #8880, opened 2016-08-19
go/stripe: update stripe package

cihangir:
## Description

Update go package of stripe
## Motivation and Context

New payment implementation will use new api


cihangir:
it does not have the glide.yaml and vendors folder updated, fyi

rjeczalik:
I've updated stripe/stripe-go in https://github.com/koding/koding/pull/8802

cihangir:
👍  thanks

PR #8882, opened 2016-08-21
Payment Backend V1

cihangir:
## Description

Koding is billed monthly, prior to services being provided (e.g.  we bill the day that the free trial ends) based upon how many team members were a part of the team in the past month. Obviously, open source is still free.  However, for the initial charge after their free trial, we should charge them for the current number of developers in their account since they may have added developers that they didn’t intend to pay for.

The pricing tiers are:
- Up to 10 users:  $49.97 per developer per month
- Up to 50 users:  $39.97 per developer per month
- Over 50 users:  $34.97 per developer per month

These price breaks apply to all users in a team.  So if a team has 11 developers, the team pays $39.97 per month per developer.

If a credit card is declined, all accounts connected to the team are immediately suspended.  The card should be retried every 24 hours.  If a new card is entered, it should be attempted immediately.
- [x] create endpoint for customer operations ( CRUD ) ( go - payment worker )
- [x] create endpoint for invoice operations ( CRUD ) ( go - payment worker )
- [x] create endpoint for assigning users to a plan (subscriptions) ( CRUD ) ( go - payment worker )
- [x] create endpoint for registering user's credit cards. ( CRUD )
- [x] auto create plans on stripe 
- [x] listen for stripe webhooks 
  - [x] "charge.succeeded"
  - [x] "charge.failed"
  - [x] "customer.subscription.created"
  - [x] "customer.subscription.deleted"
  - [x] "customer.subscription.updated"
  - [x] "customer.subscription.trial_will_end"
  - [x] "invoice.created"
  - [x] "invoice.payment_failed"
  - [x] "invoice.payment_succeeded"


go/src/koding/db/models/group.go
```diff
@@ -23,4 +23,30 @@ type Group struct {
...
+}
+
+// DeletedMember holds information about deleted members from a group.
+type DeletedMember struct {
+	Id        bson.ObjectId `bson:"_id" json:"-"`
+	AccountID bson.ObjectId `bson:"accountId" json:"accountId"`
+	GroupID   bson.ObjectId `bson:"groupId" json:"groupId"`
+	// these will be read only during query/modify time, otherwise do not access
+	// SubscriptionID string
+	// UpdatedAt time.Time
```
rjeczalik: dead code

go/src/koding/db/mongodb/modelhelper/deletedmember.go
```diff
@@ -0,0 +1,78 @@
+package modelhelper
+
+import (
+	"koding/db/models"
```
rjeczalik: Keep the imports groups ordered - stdlib, internal, external; e.g.:

```go
import (
	"time"

	"koding/db/models"
 
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 )
```

go/src/koding/db/mongodb/modelhelper/group.go
```diff
@@ -142,3 +143,17 @@ func CreateGroup(m *models.Group) error {
...
+
+func MakeAdmin(accountId, groupId bson.ObjectId) error {
+	r := &models.Relationship{
+		Id:         bson.NewObjectId(),
+		TargetId:   accountId,
+		TargetName: "JAccount",
+		SourceId:   groupId,
+		SourceName: "JGroup",
+		As:         "admin",
+		TimeStamp:  time.Now().UTC(),
```
rjeczalik: Here time in UTC is used, in `CalculateAndApplyDeletedMembers` - in local timezone time. Not sure if intentional.

go/src/koding/db/mongodb/modelhelper/deletedmember.go
cihangir: i dont manage the imports, that is what comes with goimports, what are you using instead?

go/src/koding/db/models/group.go
cihangir: left intentionally 
see line 49

go/src/koding/db/mongodb/modelhelper/group.go
cihangir: CalculateAndApplyDeletedMembers is wrong, thanks for pointing

go/src/koding/workers/janitor/exempt.go
```diff
@@ -35,16 +34,7 @@ func NewChecker(name string, fn func(*models.User, *Warning) (bool, error)) *Exe
...
-	if err == mgo.ErrNotFound {
-		return false, nil
-	}
-
-	if err != nil {
-		return false, err
-	}
-
-	return paymentapi.New("").IsPaidAccount(account)
+	return false, nil
```
rjeczalik: Please add a `TODO` for this function. (it shouldn't have`Fn` suffix, but I see it would conflict with global IsUserPaid checker).

go/src/socialapi/workers/common/mux/mux.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"socialapi/models"
 	"socialapi/workers/common/handler"
```
rjeczalik: The same nitpick as above - imports order.

go/src/socialapi/workers/common/mux/mux.go
```diff
@@ -37,6 +38,9 @@ type Mux struct {
 	config *Config
 	log    logging.Logger
 	redis  *redis.RedisSession
+
+	closing bool
+	closeMu sync.RWMutex
```
rjeczalik: `RWMutex` is not needed here, we're not expecting multiple readers and we shouldn't care - `sync.Mutex` is enough

go/src/socialapi/workers/payment/api/api.go
```diff
@@ -1,171 +1,38 @@
...
-		req.GetCreditCard(),
-	)
-}
-
-func GroupCreditCardRequest(u *url.URL, h http.Header, _ interface{}) (int, http.Header, interface{}, error) {
-	req := &payment.GroupRequest{
-		GroupId: u.Query().Get("groupId"),
+	return nil
+	//
+	// for now disable confirmation checking
```
rjeczalik:
```
// TODO(cihangir): for now disable confirmation checking, going to be enabled again when XXX happens
```

go/src/socialapi/workers/payment/api/card_test.go
```diff
@@ -0,0 +1,166 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"koding/db/mongodb/modelhelper"
+	"socialapi/models"
+	"socialapi/rest"
+	"socialapi/workers/common/tests"
+	"testing"
```
rjeczalik: import order - sometimes `goimport` gets it right and format correctly: https://github.com/koding/koding/pull/8882/files#diff-2be48698b8c33592e1f90e3fa58e88cfR3

go/src/socialapi/workers/payment/api/server_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/koding/runner"
+	"github.com/stripe/stripe-go"
+)
+
+func withTestServer(t *testing.T, f func(url string)) {
+	const workerName = "paymentwebhook"
+
+	r := runner.New(workerName)
+	if err := r.Init(); err != nil {
+		t.Fatal(err.Error())
```
rjeczalik: just t.Fatal(err)

go/src/socialapi/workers/payment/customer.go
```diff
@@ -0,0 +1,354 @@
...
+	usage.Plan = plan
+	usage.Due = uint64(usage.User.Total) * plan.Amount
+
+	return usage, nil
+
+}
+
+func fetchParallelizableeUsageItems(group *models.Group) (*Usage, error) {
+	var infoErr error
+	var errMu sync.RWMutex
```
rjeczalik: sync.Mutex is enough

go/src/socialapi/workers/payment/plan.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return Plans[Over50Users].ID
+}
+
+// CreateDefaultPlans creates predefined default plans. This is meant to be run
+// when the worker starts to be sure the plans are there.
+func CreateDefaultPlans() error {
+	for _, planParams := range Plans {
+		p, err := stripeplan.Get(planParams.ID, nil)
```
rjeczalik: err is not handled here

go/src/socialapi/workers/payment/tests.sh
```diff
@@ -3,11 +3,6 @@
 echo "running payment tests"
 go test -c socialapi/workers/payment/ && ./payment.test -c $(dirname $0)/../../config/dev.toml -test.v=true
 
-echo "running stripe tests"
-go test -c socialapi/workers/payment/stripe/ && ./stripe.test -c $(dirname $0)/../../config/dev.toml -test.v=true
+echo "running api tests"
+go test -c socialapi/workers/payment/api && ./api.test -c $(dirname $0)/../../config/dev.toml -test.v=true
```
rjeczalik: More elegant imo is to make test read flags params after `--`:

```
go test -v -race socialapi/workers/payment/api -- -c $(dirname $0)/../../config/dev.toml
```

in tests:

```go
var testArgs []strong

func init() {
    for i, arg := range os.Args {
        if arg == "--" {
            testArgs = os.Args[i+1:]
            break
        }
    }
}

...
flag.CommandLine.Parse(testArgs)
...
```

Just an idea.

rjeczalik: And please add `-test.race`.

go/src/socialapi/workers/payment/webhook.go
```diff
@@ -0,0 +1,303 @@
...
+
+	"invoice.created":           invoiceCreatedHandler,
+	"invoice.payment_failed":    invoicePaymentFailedHandler,
+	"invoice.payment_succeeded": invoicePaymentFailedHandler,
+}
+
+// GetHandler returns the registered handler for stripe webhooks if registered
+func GetHandler(name string) (StripeHandler, error) {
+	action, ok := stripeActions[name]
+	if ok {
```
rjeczalik: Can we handle exceptional branch instead?

```go
action, ok := stripeActions[name]
if !ok {
    return nil, errors.New("handler not found")
}

return action, nil
```

More idiomatic™.

go/src/socialapi/workers/payment/webhook.go
```diff
@@ -0,0 +1,303 @@
...
+	cus, err := customer.Get(invoice.Customer.ID, nil)
+	if err != nil {
+		return err
+	}
+
+	group, err := modelhelper.GetGroup(cus.Meta["groupName"])
+	// we might get events from other environments where we might not have the
+	// group in this env.
+	if err == mgo.ErrNotFound {
+		return nil
```
rjeczalik: Wouldn't it be good to log such cases?

go/src/koding/db/models/group.go
rjeczalik: Can we prefix it with a `// TODO(cihangir):`, so it's easy to e.g. `grep -r TODO go/src/koding`?

go/src/koding/db/mongodb/modelhelper/deletedmember.go
rjeczalik: `goimports` usually fail to sort it right, it think it's caused that both `koding/db/models` and  `gopkg.in/mgo.v2` look like external depedency.

I think in 1.6 it may sort vendored imports correctly, but well I'm groupping them by hand, correcting after goimports when it fails to do it for me.

rjeczalik:
@cihangir LGTM 👍 I made only a few minor code style nitpicks.

go/src/koding/db/models/group.go
cihangir: it seems my inline doc is not clear as i wanted it to be, what i meant actually "those fields are in the database, know they are there, but do not access them in the app directly." so it is not a TODO item, will re-word the sentence

go/src/koding/db/mongodb/modelhelper/deletedmember.go
cihangir: I dont remember when i last compiled goimports :) After upgrading Go in our repo, i could do a pass in our app directory with new binary

go/src/socialapi/workers/payment/plan.go
cihangir: it seems you didnt like my err == nil handling there :) used another approach

go/src/socialapi/workers/payment/tests.sh
cihangir: Interesting suggestion, i will experiment with it to make it compatible with other workers/tests - currently "runner" does the flag registration/parsing part

go/src/socialapi/workers/payment/webhook.go
cihangir: when i first wrote it it was returning a default handler, after refactoring, forgot the change the order, fixed it 👍

go/src/socialapi/workers/payment/webhook.go
cihangir: since we have one testing account all the event will go the sandbox endpoint and it will log many unnecessary lines, if you think we could catch some issues with logging them, i could add "if config.Env == "sandbox" do not log etc, i didnt see a reason tho

go/src/koding/workers/janitor/main.go
```diff
@@ -40,6 +40,7 @@ type janitor struct {
```
cihangir: @rjeczalik  addressed your comments under this commit, PTAL

go/src/socialapi/workers/payment/webhook.go
rjeczalik: I think we can leave it for now, I thought in places where we make assumptions (like here - the other environments causing code to branch here) it'd be nice to have a way of testing that assumption.

My comment was rather about defensive log statement there. I think it's the right way to have conditional logging (`if config.Env == "sandbox"`), as it's logger responsibility to filter messages and decide whether to display them or not. There's no such mechanism in the logging library so I think it's not worth it.

go/src/koding/workers/janitor/main.go
rjeczalik: @cihangir Looks good :+1:

gokmen:
LGTM 👍  I've cleaned up remaining payment related parts of Compute also moved old plan functionality to a limit functionality for future requirements.

PR #8695, opened 2016-08-03
provisionklient: do not check usernames

rjeczalik:
reference: https://koding.slack.com/archives/support/p1470174389000380

PR #8884, opened 2016-08-21
nginx: reverse proxy s3 urls

rjeczalik:
ref: https://github.com/koding/koding/pull/8695#discussion_r73578313

After the deploy urls in klient are going to be switch to use the revproxied ones.

### DIFF
--- a/deployment/nginx.coffee
+++ b/deployment/nginx.coffee
@@ -365,6 +365,11 @@ module.exports.create = (KONFIG, environment)->
         resolver 8.8.8.8;
       }
 
+      # redirect /s3/:bucket/:key to :bucket.s3.amazonaws.com/:key
+      location ~^/s3/([^/]*)/(.*)$ {
+        return 302 https://$1.s3.amazonaws.com/$2;
+      }
+
       # Hackathon2014 is the old hackathon page and served via webserver
       # todo(cihangir) remove after hubspot integration
       location = /Hackathon2014 {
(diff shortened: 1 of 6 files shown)


deployment/nginx.coffee
```diff
@@ -365,6 +365,17 @@ module.exports.create = (KONFIG, environment)->
         resolver 8.8.8.8;
       }
 
+      # redirect /s3/:bucket/:key to :bucket.s3.amazonaws.com/:key
+      location ~^/s3/([^/]*)/(.*)$ {
```
cihangir: thanks for the very generalized solution but this might give opportunities to malicious users eg they can abuse our bandwidth with big files  

What do you think of using smth like `~*(^\/s3\/koding-(klient|softlayer|provision))/(.*)$` to make it arguably less prone?

ps: didnt test the regex 😇

rjeczalik: good point, will do

deployment/nginx.coffee
```diff
@@ -365,6 +365,17 @@ module.exports.create = (KONFIG, environment)->
...
+        proxy_pass            "https://$1.s3.amazonaws.com/$2";
```
szkl: A trivial note. I'd prefer `https://s3.amazon.aws.com/$1/$2` to avoid resolving subdomain name.

deployment/nginx.coffee
rjeczalik: @cihangir on a second thought, wouldn't it be better to redirect instead of revproxy?

cihangir: awesome 👍  (302)

deployment/nginx.coffee
cihangir: see #kd channel, aws started redirecting with invalid headers

deployment/nginx.coffee
rjeczalik: done

cihangir:
👍

PR #8898, opened 2016-08-24
Redux-Bongo instances

hakankaradis:
This PR introduces a new way on how we can store bongo instances by using **bongo middleware** and **bongo reducer** into redux store.

Every dispatched action with bongo key goes to bongo middleware. Bongo middleware (like any other middleware) dispatch another action with promise key and three action types (`BEGIN`, `SUCCESS`, `FAIL`).  After it dispatches related `SUCCESS` action, `bongo` reducer listens to these changes, and cache the data.

One other thing we are using bongo middleware is that binding `update` and `delete` event to each bongo instance, so it will automatically dispatch actions to update or delete your instances.

This is a guaranteed way of making sure that any bongo instance we may want to use, will always up-to-date once it's loaded through the `bongo middleware`. So you don't need to worry about that your store has most recent instance or not.

### DIFF
--- a/client/app/lib/redux/middleware/bongo.coffee
+++ b/client/app/lib/redux/middleware/bongo.coffee
@@ -1,25 +1,28 @@
 _ = require 'lodash'
 { LOAD, REMOVE } = require 'app/redux/modules/bongo'
 
-
 module.exports = bongoMiddleware = (remote) -> (store) -> (next) -> (action) ->
+
   return next action  unless action.bongo
+
   { bongo, type, types } = action
   rest = _.omit action, ['bongo', 'type', 'types']
 
   next
     types: types or generateTypes(type)
     promise: ->
+
       bongo(remote, { getState: store.getState }).then (results) ->
-        results = [results]  unless Array.isArray results
+
+        unless Array.isArray results
+          results = [results]
+
         results.forEach (result) ->
           result.on 'update', ->
             store.dispatch { type: LOAD.SUCCESS, result }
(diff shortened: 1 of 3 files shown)


client/component-lab/SidebarMachinesListItem/SidebarMachinesListItem.coffee
```diff
@@ -0,0 +1,28 @@
...
+  @propTypes =
+    machine: React.PropTypes.object
+
+  @defaultProps =
+    machine : {}
+
+
+  render: ->
+
+    return  unless @props.machine
```
usirin: This should not be a concern of this component. If `machine` not present, it shouldn't even be rendered.

client/app/lib/redux/selectors/sidebar.coffee
```diff
@@ -0,0 +1,82 @@
...
+
+stacksWithTemplates = (stacks, stackTemplates) ->
+
+  return null if not stacks or not stackTemplates
+
+  stacksWithTemplates = {}
+  Object.values(stacks).forEach (stack) ->
+    baseStackId = stack.baseStackId
+    if baseStackId
+      stacksWithTemplates[stack._id] = stackTemplates["#{baseStackId}"]
```
usirin:
```coffee
stacksWithTemplates[stack._id] = stackTemplates[baseStackId]
```

usirin:
Looks great man! Merging.

PR #8904, opened 2016-08-25
gateway: kite gateway to s3 for klient

rjeczalik:
This PR adds a "gateway.auth" method to kloud, which gives a temporary token for a klient with a write permission only to s3://koding-publiclogs/:username/logs.

Please take a look.


go/src/koding/kites/gateway/gateway.go
```diff
@@ -0,0 +1,315 @@
+package gateway
```
cihangir: first of all i am not good at naming things, what i have seen in wild with naming things "gateway" after, is mostly when they take actions at the backend with the passed request, on other words, they act as an entrance to internal services/networks etc. The functionality for this kite does not look like it :) what about changing it to something like s3keygen?

rjeczalik: Can it be just `keygen`? In future it will handout tokens for any external service that klient uses if possible, e.g. for datadog or whatever will collect metrics on klient side.

cihangir: why not

rjeczalik: done in https://github.com/koding/koding/pull/8947/commits/32b6175acc2aa249fa3de8c4d20b0d3054123c5e

rjeczalik:
@leeola @cihangir gentle ping

cihangir:
i was done with my review, thanks for the PR 👍

rjeczalik:
@leeola Please take a look, I'll address your comments in separate PR.

leeola:
:+1:, looking now

PR #8812, opened 2016-08-15
Participated Team list feature

alex-ionochkin:
## Description

This PR adds:
- a new landing page which sends email with a list of teams user participates or is invited to (for logged out user)
- a new modal showing a list of user teams (for logged in user)

Also, PR includes routing fixes to `/Teams` landing pages
## Motivation and Context

https://github.com/koding/koding/issues/8332
## How Has This Been Tested?

`./run nodetestfiles ./servers/models/account.test.coffee`
`./run nodetestfiles ./servers/lib/server/handlers/findteam.test.coffee`
## Screenshots (if appropriate):

http://recordit.co/Orv6gfNOE5
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #8332, opened 2016-07-06
Participated Team list feature set


gokmen:
- [x] Provide list of teams that I'm part of once I logged in to any of my team
  - [x] Backend requirement (https://github.com/koding/koding/pull/7883)
  - [x] Tests needs to be written
  - [x] Design
  - [x] UI implementation

---

- [x] Provide list of teams that I'm part of by email when I provided it from landing page
  - [x] Backend requirement
  - [x] Tests needs to be written
  - [x] Email Design
  - [x] Landing page implementation
 

ref: https://koding.atlassian.net/browse/TMS-3044

diclee:
Email with the list of participated teams cannot be received. @alex-ionochkin can you please look at it to see if anything is wrong with backend requirements or it's a problem about the email service.

alex-ionochkin:
@diclee, it works fine for me on sandbox. I was able to receive correct email after requesting it on https://sandbox.koding.com/Teams/FindTeam. Can you check it once again and make sure that email address is correct? If it still doesn't work for you, I'll check logs to understand the reason of the problem

diclee:
@alex-ionochkin I've checked it again and it worked on latest but not on sandbox. I'm closing the issue since it's working on latest (will work on production as well). Can you please check sandbox again when you have time?

PR #8905, opened 2016-08-25
Team list feature: addition fixes

alex-ionochkin:
## Description

This PR contains a few fixes in addition to https://github.com/koding/koding/pull/8812:
- sprite images with hyphens are renamed to use underscores instead
- `nib` is used for linear gradients styles
- fix for a bug of missing bottom border of the list when it has the only item
- style fixes for Safari
## Motivation and Context

https://github.com/koding/koding/issues/8332


client/app/lib/changeteam/controller.coffee
```diff
@@ -16,5 +16,9 @@ module.exports = class ChangeTeamController extends KodingListController
         return  if showError err
         groups = groups.filter (group) -> group.slug isnt 'koding'
         callback null, groups
+    options.noItemFoundWidget = new kd.CustomHTMLView
+      tagName  : 'p'
```
alex-ionochkin: I changed tag name of `noItemFoundWidget` to `p` in order to have `div`s for list items only. In this case I can use `:first-of-type` and `:last-of-type` to specify styles of first and last item in css

package.json
```diff
@@ -69,7 +69,7 @@
     "minimist": "0.0.8",
     "mkdirp": "0.3.2",
     "mongodb": "1.4.14",
-    "nib": "0.5.0",
+    "nib": "1.1.0",
```
alex-ionochkin: Old version of `nib` throws exception for `linear-gradient(-180deg,#3c3c57 11%,#8598be 100%)` since it uses degree as argument. Newer version has a fix of this bug

PR #8906, opened 2016-08-25
klientctl: added --fuse option

leeola:
To simplify the UX of turning off "smart mount", a --fuse option has
been added. This will force the usage of Fuse regardless of the remote
folders size or containing data. Further modification of the fuse
settings can of course still use --prefetch-all and etc.


rjeczalik:
@leeola: 👍

PR #8913, opened 2016-08-29
Recover kontrol database schema script

szkl:
Signed-off-by: Sonmez Kartal sonmez@koding.com


rjeczalik:
:+1:

These were lost when switching to glide tool. Also fixed in [go1.6 branch](https://github.com/koding/koding/pull/8802).

PR #8914, opened 2016-08-29
MD: add initial contributing doc

cihangir:


cihangir:
cc/ @sibblegp @devrim

PR #8947, opened 2016-09-05
klient: rework logging

rjeczalik:
~~deps: https://github.com/koding/koding/pull/8904~~

This PR reworks logging in klient, making it possible to upload arbitrary log files from remote machines using keygen and logrotate packages. Please take a look.

This PR also adds "log.upload" method to klient kite, which allows for adding e.g. `kd upload` command if needed. /cc @leeola


go/src/koding/kites/keygen/keygentest/gatewaytest.go
```diff
@@ -1,18 +1,60 @@
-package gateway_test
```
cihangir: file name still "gateway" intentional?

go/src/koding/kites/keygen/keygentest/gatewaytest.go
```diff
@@ -114,6 +156,29 @@ func (d *Driver) keyPair() *kitetest.KeyPair {
 // ParseFlags uses multiconfig to fill the given v with matching
 // flag values read from tags, environment and command line.
 func ParseFlags(v interface{}) error {
+	type parentFlags interface {
+		Underlying() *Flags
+	}
+
+	envPrefix := "gateway"
```
cihangir: why do you choose gateway for env vars instead of the service/worker name?

go/src/koding/klient/uploader/uploader.go
```diff
@@ -0,0 +1,275 @@
...
+type Options struct {
+	KeygenURL string      // required
+	Kite      *kite.Kite  // required
+	Bucket    string      // required
+	Region    string      // required
+	DB        *bolt.DB    // optional; in-memory store if nil
+	Log       kite.Logger // optional; defaultLog if nil
+}
+
+// Logger is a wrapper for kite.Logger that implements logging.Logger.
```
cihangir: i didnt understand the requirement here, could you please explain? AFAIK kite.logger is already implements koding.logger as we can see here https://github.com/koding/kite/blob/02ecbfc2262fb6dd7be61a0ff23b7a79431a02fe/logger.go#L80

config/generateKonfig.coffee
```diff
@@ -138,6 +138,10 @@ module.exports = (options, credentials) ->
     userPublicKey: credentials.kloud.userPublicKey
     userPrivateKey: credentials.kloud.userPrivateKey
 
+    keygenAccessKey: credentials.kloud.gatewayAccessKey
```
rjeczalik: done

go/src/koding/kites/keygen/keygentest/gatewaytest.go
rjeczalik: done, renamed to keygentest.go, I missed this any many others 😇

go/src/koding/kites/keygen/keygentest/gatewaytest.go
rjeczalik: done, renamed to "keygen"

go/src/koding/klient/uploader/uploader.go
rjeczalik: The code you linked creates kite.Logger from logging.Logger. Here's the opposite - creating logging.Logger from kite.Logger.

go/src/koding/klient/uploader/uploader.go
```diff
@@ -0,0 +1,275 @@
...
+			1: {Dir: reflect.SelectRecv, Chan: reflect.ValueOf(up.req)},
+		}
+		prefix = up.cfg.Kite.Config.Id
+	)
+
+	for {
+		var req *UploadRequest
+
+		switch n, v, _ := reflect.Select(cases); n {
+		case 0:
```
cihangir: why we need to use reflect here?

rjeczalik: Do be able to select on dynamic number of channels.

Alternative would be something like:

```go
tickers := []*time.Ticker

loop:
    for {
        for _, i := range rand.Perm(len(tickers)) {
            switch {
            case <-tickers[i].C:
                // ...
                // handling here
                // ...
                continue loop
            default:
                continue
            }
        }
    }
```

I find the reflect more readable & simpler.

go/src/koding/klient/uploader/uploader.go
rjeczalik: Now I got it on a second read - the kite.Logger interface can be upgraded to logging.Logger interface, since the underlying value implements logging.Logger when created by kite.

Done, I've removed the wrapper in https://github.com/koding/koding/pull/8947/commits/71eef548cb053017890af6c96fce5f2520b6d159.

go/src/koding/klient/uploader/uploader.go
cihangir: i think another way to do that could be

```
	for {
		select {
		case <-up.close:
			return
		case req := <-up.req:
			go func() {
				t := time.NewTicker(req.Interval)
				for {
					select {
					case <-up.close:
						t.Stop()
						return
					case <-t.C:
						// do your magic here
					}
				}
			}()
			// do your magic here, for the first time
		}
	}
```

cihangir: Also please check if the files are already listened/watched by another ticker, multiple requests to kite.upload logs could result in multiple tickers

rjeczalik: Also would work, but I don't like starting additional goroutines for something, that could be done in one. In my opinion it's easier to reason about concurrency and do not worry about leaking goroutines.

Your call, if you find it easier to read I can rework reflect.Select to multiple goroutines. :-)

rjeczalik: > Also please check if the files are already listened/watched by another ticker

done in https://github.com/koding/koding/pull/8947/commits/ee069993f0a3e7b6e11b92b87ff62e35502f9e6c

cihangir: I think we have different takes on readability :) 

Since you said  "your call" lets change it to non-reflected version not to fiddle with the runtime memory 🙋

rjeczalik: done in https://github.com/koding/koding/pull/8947/commits/d59fe509b2e3437bb1d7be208ac35878799f9c56

leeola:
LGTM

rjeczalik:
@leeola @cihangir Added one more - https://github.com/koding/koding/commit/d37869569c89391602334e48fda86c38b551f9bc#diff-18cc3d0abd4e3e71be2ae47707313dfeR42. Please take a look.

leeola:
:+1:, LGTM

Issue #8791, opened 2016-08-11
If you close 'share stack template' modal without selecting an option, you're stuck on stack editor even though the stack has been saved successfully.


diclee:
A suggestion as a solution: "The button 'Initialize' should stay there to initialize stack and maybe we can dim the “make team default” button to give the sense that this was done successfully."

alex-ionochkin:
I agree that `Initialize` button should stay there but I would keep `Make Team Default` button active since user may change their mind and make the stack template default for the team.

diclee:
Verify passed on sandbox.

PR #8952, opened 2016-09-06
StackEditor: editor updates after button actions

alex-ionochkin:
## Description

This PR contains a few fixes in stack editor:
- it leaves `Make Team Default` button visible (and disabled) on the editor after user made stack template default for the team
- it returned back reloading editor on `Reload` event (previously it was broken)
## Motivation and Context

https://github.com/koding/koding/issues/8791
https://github.com/koding/koding/issues/8969

http://recordit.co/sX3fC8Es1M
## Types of changes

### DIFF
--- a/client/stack-editor/lib/index.coffee
+++ b/client/stack-editor/lib/index.coffee
@@ -23,6 +23,13 @@ module.exports = class StackEditorAppController extends AppController
 
     @selectedEditor = null
 
+    { router } = kd.singletons
+    router.on 'RouteInfoHandled', (routeInfo) =>
+      return  unless @selectedEditor
+
+      { data, reloadRequired } = @selectedEditor
+      @removeEditor data.stackTemplate._id  if reloadRequired
+
 
   openEditor: (stackTemplateId) ->
 
@@ -140,9 +147,8 @@ module.exports = class StackEditorAppController extends AppController
 
     { computeController } = kd.singletons
 
-    EnvironmentFlux.actions.fetchAndUpdateStackTemplate templateId, (template) =>
-      @removeEditor template._id
-      @showView template
+    EnvironmentFlux.actions.fetchAndUpdateStackTemplate(templateId).then (template) =>
+      @editors[template._id]?.reloadRequired = yes
 
 
   createEditor: (stackTemplate) ->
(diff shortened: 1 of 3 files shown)


sinan:
thanks @alex-ionochkin but this doesn't seem to address the issue reported. if you click the overlay of that success modal the editor remains in a weird state. Does this cover that case?

client/stack-editor/lib/index.coffee
```diff
@@ -140,9 +147,8 @@ module.exports = class StackEditorAppController extends AppController
 
     { computeController } = kd.singletons
 
-    EnvironmentFlux.actions.fetchAndUpdateStackTemplate templateId, (template) =>
-      @removeEditor template._id
-      @showView template
+    EnvironmentFlux.actions.fetchAndUpdateStackTemplate(templateId).then (template) =>
+      @editors[template._id]?.reloadRequired = yes
```
usirin: Let's not modify the internal state of the editor here, instead you can write a method like `setReloadRequired` and it would handle binding necessary state variables. e.g a simple JS object (`@shouldReloadMap`), which has the id as key, and value as `true`(which is a dummy variable, simply if you have a key, just reload those editors, when needed). 

That way editors will still be **just** editors, and the state stays external. In the `RouteInfoHandled` event handler we can simply remove all the editors that should be removed (which we are reading it from `@shouldReloadMap`, or if you still want to make sure that we are only removing the selectedEditor instance, you can still keep the `return  unless @selectedEditor` line in the changes.

alex-ionochkin: good catch 👍

alex-ionochkin:
You're right, @sinan. It's my bad :) I understood issue incorrectly. I thought the problem is that user is stuck on the editor after they performed `Make Team Default` action since there is no `Initialize`/`Re-initialize` button on the editor after it so user is not able to initialize the stack.
I'll add a commit for the original issue but let's keep my previous changes in this PR since they are also relevant

alex-ionochkin:
Here is a new screencast - http://recordit.co/Wdie6oByxb
Please also see my comment https://github.com/koding/koding/issues/8791#issuecomment-245252050 about `Set Team Default` button

alex-ionochkin:
UPDATE: I've added a link to https://github.com/koding/koding/issues/8969 since this PR fixes that issue

PR #8966, opened 2016-09-07
Payment Backend - Custom Plan support 

cihangir:
depends on https://github.com/koding/koding/pull/8882


cihangir:
will open against master branch

PR #8977, opened 2016-09-08
Verify user email if it wasn't verified while fetching otaToken

### LINKED ISSUE #8910: Show proper error message for KD Cli section
Show proper error message if user email is not verified yet

## Expected Behavior
In utilities kdcli section, if user email is not verified 
User should see this error message  `Sorry, you need to confirm your email address first.`

## Possible Solution
Check that user email is verified

1. Create team with not verified email
2. Go to utilities kd cli page

hakankaradis:
For users from solo account with no verified email, verify their emails while fetching otaToken.
## Motivation and Context

fixes #8910 
## How Has This Been Tested?

It doesn't break anything for other users.
Needs an account with not verified email and will see otaToken in Cli section in utilities

### DIFF
--- a/workers/social/lib/social/models/account.coffee
+++ b/workers/social/lib/social/models/account.coffee
@@ -1365,26 +1365,18 @@ module.exports = class JAccount extends jraphical.Module
 
     return errorCallback()  unless sessionToken
 
-    @isEmailVerified (err, isVerified) ->
+    JSession         = require './session'
+    { v4: createId } = require 'node-uuid'
 
-      return callback err  if err
+    JSession.one { clientId: sessionToken }, (err, session) ->
 
-      if not isVerified
-        message = 'Sorry, you need to confirm your email address first.'
-        return callback new KodingError message
+      if err or not session
+        return errorCallback()
 
-      JSession         = require './session'
-      { v4: createId } = require 'node-uuid'
-
-      JSession.one { clientId: sessionToken }, (err, session) ->
-
-        if err or not session
-          return errorCallback()
-
-        [otaToken] = createId().split '-'
-        session.update { $set: { otaToken } }, (err) ->
-          if err then errorCallback()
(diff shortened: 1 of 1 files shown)


usirin:
@gokmen can you please take a look?

workers/social/lib/social/models/account.coffee
```diff
@@ -1358,33 +1361,40 @@ module.exports = class JAccount extends jraphical.Module
...
 
-    @isEmailVerified (err, isVerified) ->
+    @isEmailVerified (err, isVerified) =>
 
       return callback err  if err
 
       if not isVerified
-        message = 'Sorry, you need to confirm your email address first.'
-        return callback new KodingError message
+        @findUserAndConfirmEmail username, (err) =>
```
gokmen: instead of verifying it in here automatically let's remove this check block from here completely. we'll update old data for existing users on production.

workers/social/lib/social/models/account.coffee
```diff
@@ -585,13 +585,16 @@ module.exports = class JAccount extends jraphical.Module
...
+      @findUserAndConfirmEmail username, callback
+
+  findUserAndConfirmEmail: (username, callback) ->
+    JUser = require './user'
+    JUser.one { username }, (err, user) ->
+      return  callback err if err
+      return  callback new Error 'User is not found' unless user
+      user.confirmEmail (err) ->
+        return callback new Error 'An error occurred while confirming email' if err
+        callback null, yes
```
gokmen: let's revert this change as well after: https://github.com/koding/koding/pull/8977/files#r78067181

hakankaradis:
@gokmen  I removed isEmailVerified check, we will not need this check anymore since we verify all emails for team users. So I think this pr will be merge-able after we verify solo users email

gokmen:
LGTM 👍  restarted tests

PR #8984, opened 2016-09-08
stacktemplateitem: member will edit his own stack from stacks page

hakankaradis:
Member will be able to edit his/her own stack template from Dashboard/Stacks page.


gokmen:
LGTM 👍

Issue #8971, opened 2016-09-07
'Stack Updated' modal is disappearing automatically for the admin updating the stack.


diclee:
## Expected Behavior
It should not disappear.
## Current Behavior
![gif](http://g.recordit.co/MkElyqlVk4.gif)
http://recordit.co/MkElyqlVk4

diclee:
Verify passed on sandbox.

PR #8997, opened 2016-09-12
StackUpdatedWidget: make visible when stack unread count increases

alex-ionochkin:
## Description

The main goal of this PR is to show `StackUpdatedWidget` when stack unread count increases.
Also, I removed the state of `StackUpdatedWidget` since it strongly correlates with the state of parent component. I replaced `isShown` state prop with simple prop `visible` which is set by parent. Parent knows when to show `StackUpdatedWidget` and it listens to widget's `onClose` callback in order to hide it. Visibility of `StackUpdatedWidget` is set by parent's `state.showWidget` value.
## Motivation and Context

https://github.com/koding/koding/issues/8971

http://recordit.co/sWsMY6JOvf
## Types of changes

### DIFF
--- a/client/app/lib/components/sidebarstacksection/index.coffee
+++ b/client/app/lib/components/sidebarstacksection/index.coffee
@@ -33,10 +33,16 @@ module.exports = class SidebarStackSection extends React.Component
 
 
   getDataBindings: ->
+
     selectedTemplateId: EnvironmentFlux.getters.selectedTemplateId
 
 
-  componentWillReceiveProps: -> @setCoordinates()
+  componentWillReceiveProps: (nextProps) ->
+
+    nextStackUnreadCount = @getStackUnreadCount nextProps.stack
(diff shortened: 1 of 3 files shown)


client/app/lib/components/sidebarstacksection/index.coffee
```diff
@@ -33,10 +33,16 @@ module.exports = class SidebarStackSection extends React.Component
...
   getDataBindings: ->
+
     selectedTemplateId: EnvironmentFlux.getters.selectedTemplateId
 
 
-  componentWillReceiveProps: -> @setCoordinates()
+  componentWillReceiveProps: (nextProps) ->
+
+    nextStackUnreadCount = @getStackUnreadCount nextProps.stack
+    @setState { showWidget : yes }  if nextStackUnreadCount > @getStackUnreadCount()
```
usirin: :+1:

PR #9004, opened 2016-09-13
kloud: cleanup providers / enable providers as plugins

rjeczalik:
This PR cleans up providers and introduces common interface for a kloud provider (vs previous `interface{}`). This is to enable internal providers and documenting writing a custom one.


go/src/koding/kites/kloud/provider/provider.go
```diff
@@ -21,6 +21,10 @@ import (
 	"gopkg.in/mgo.v2/bson"
 )
 
+// All is a global lookup map used ny kloud to register
+// available stack providers.
+var All = make(map[string]func(*BaseProvider) stack.Provider)
```
cihangir: instead of exposing the var, i would expose a function 
smth like

```
type Stacker interface {
  Provide(*provider.BaseProvider) stack.Provider
}

func MustRegister(name string, s Stacker) {
  if _, ok := provider.All[name]; ok {
    panic("...")
  }

  all[name] = s  
}
```

it is not a problem, just a taste of my being on the safe side :)

go/src/koding/kites/kloud/kloud/kloud.go
```diff
@@ -462,118 +411,7 @@ func newSession(conf *Config, k *kite.Kite) (*session.Session, error) {
 	return sess, nil
 }
 
-// Providers ctors. If a provider has a dependencty on a service
-// defined in *session.Session, and the service is disabled (nil)
-// due to missing configuration or any other reason, we return
-// disabled provider that rejects all requests with 411 *kite.Error.
-
-func newKodingProvider(sess *session.Session, conf *Config, authUsers map[string]string) stack.Provider {
```
cihangir: as you know i already removed these :)

go/build.sh
```diff
@@ -1,131 +1,110 @@
...
 )
 
-tldflags="-X main.GitCommit${LINK_OPERATOR}${version:0:8}"
-`which go` install -v -ldflags "$tldflags"  "${terraformservices[@]}"
+# source configuration for kloud providers
+for provider in $KODING_REPO/go/src/koding/kites/kloud/provider/*; do
+	if [[ -d "${provider}/build.sh.d" ]]; then
+		for src in ${provider}/build.sh.d/*; do
+			if [[ -f "$src" ]]; then
+				source "$src"
```
cihangir: in general cases a `if [ -r "$src" ]` check would be a good practice but i think we can omit it here since we will manage the build scripts.

go/src/koding/kites/kloud/provider/provider.go
rjeczalik: The thinking behind `provider.All` is it's meant to be modified in packages' init functions, so access is always synchronised. Sometimes (especially in tests, which I'm planning to add) I'd want to instrument existing providers - using functions instead of global map I'd need to also add `Get*` and `Delete*`.

I'm going to think about this more and eventually update in separate PR.

go/src/koding/kites/kloud/kloud/kloud.go
rjeczalik: Let's merge https://github.com/koding/koding/pull/9042 first and then I'll rebase my PR.

rjeczalik:
> Review-Reviewed

@cihangir: rebased, merging

cihangir:
👍

PR #9019, opened 2016-09-14
ResourceStateModal: show machine name on the modal when turning VM on

alex-ionochkin:
It shows selected machine's name on the modal with `TURN VM ON` button (as khaled requested)

vm-name-on-modal

### DIFF
--- a/client/app/lib/providers/resourcestatemodal/views/machineflow/startmachinepageview.coffee
+++ b/client/app/lib/providers/resourcestatemodal/views/machineflow/startmachinepageview.coffee
@@ -15,16 +15,19 @@ module.exports = class StartMachinePageView extends JView
 
   pistachio: ->
 
-    '''
+    machine = @getData()
+    title   = machine.jMachine.label
+
+    """
       <div class="start-machine-flow start-machine-page">
         <header>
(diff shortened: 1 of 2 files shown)


sinan:
👍

client/app/lib/providers/resourcestatemodal/views/machineflow/startmachinepageview.coffee
```diff
@@ -15,16 +15,19 @@ module.exports = class StartMachinePageView extends JView
...
+
+    """
       <div class="start-machine-flow start-machine-page">
         <header>
           <h1>Boot Virtual Machine</h1>
         </header>
         <section class="main">
           <div class="background"></div>
-          <h2>Let's Boot up your VM</h2>
+          <h2>Let's Boot up #{title}</h2>
```
sinan: actually @gokmen reminded me that, this way of using pistachio breaks the realtime updates when data is changed, please use it as:

```
<h2>Let's Boot up {{ #(jMachine.label)}}</h2>
```

see more here: https://github.com/koding/koding/blob/master/node_modules_koding/pistachio/browser/pistachio.js#L34

PR #9034, opened 2016-09-15
ResourceStateModal: pistachio cleanup in different views

alex-ionochkin:
@sinan, this is a change you mentioned in https://github.com/koding/koding/pull/9019#pullrequestreview-48038.
Thanks for suggestion!


alex-ionochkin:
Also, I added the similar pistachio fixes in other `ResourceStateModal` views

gokmen (commented):
💯

Issue #9015, opened 2016-09-14
Admin cannot access disabled user's stack.


diclee:
## Current Behavior
![gif](http://g.recordit.co/0SREGlNBOG.gif)
http://recordit.co/0SREGlNBOG

After trying to fix the permissions, admin gets 'Something went wrong' error message displayed and disabled user's stack is not loaded as a result.

diclee:
It fails when 'Stack Creation' is disabled for team members.

alex-ionochkin:
Real server error is:
```
KiteError {message: "runtime error: invalid memory address or nil pointer dereference", name: "KiteError", type: "genericError", code: "", isOperational: true}
```
It happens when kloud's `addAdmin()` method is called - https://github.com/koding/koding/blob/master/client/app/lib/providers/computecontroller.coffee#L1128
Here is a call error stack of error in VM terminal:
```
goroutine 42975 [running]:
runtime/debug.Stack(0x40f40c, 0x1b7d840, 0xa568350b3f00)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x79
runtime/debug.PrintStack()
        /usr/local/go/src/runtime/debug/stack.go:16 +0x22
vendor/github.com/koding/kite.(*Client).runMethod.func1(0xc420a07f28, 0xc4204af400, 0xc420a07f30)
        /home/aionochkin/koding/go/src/vendor/github.com/koding/kite/request.go:68 +0x6e
panic(0x120e480, 0xc42000c0a0)
        /usr/local/go/src/runtime/panic.go:458 +0x243
koding/kites/kloud/klient.(*Klient).AddUser(0x0, 0xc420a08b90, 0x4, 0x0, 0x0)
        /home/aionochkin/koding/go/src/koding/kites/kloud/klient/klient.go:262 +0x37
koding/kites/kloud/stack.(*Kloud).AdminAdd(0xc42018c420, 0xc420a44000, 0xc420ca7dc8, 0x4417eb, 0x0, 0x5368d9)
        /home/aionochkin/koding/go/src/koding/kites/kloud/stack/admin.go:28 +0x71
koding/kites/kloud/stack.(*Kloud).AdminAdd-fm(0xc420a44000, 0xc4204af470, 0xc4ffffffff, 0x0, 0xc4202dc1c0)
        /home/aionochkin/koding/go/src/koding/kites/kloud/kloud/kloud.go:345 +0x34
[...]

PR #9037, opened 2016-09-15
kloud/klient: fix NewWithTimeout

rjeczalik:


gokmen:
@rjeczalik can u provide a brief description? what is this for, I missed the context.

rjeczalik:
@gokmen It's a fix for #9015. I broke the implementation with https://github.com/koding/koding/commit/b5b8f645609e82622ae2cc2e831778f4f2a80417#diff-bba21813d7b81c119f739eab0bbb45faR195 - when DefaultInterval == timeout, then both return values were nil.

PR #9048, opened 2016-09-18
Tracking update

sinan:
aff/link tracking for koding.com
## Description

updates both node and go backend to send updated events to segment.
## Motivation and Context

it enables to reward 3rd parties for their affiliate efforts and provides koding.com a better traffic intelligence.
## How Has This Been Tested?

through server logs on both segment and hasoffers admin ui's.
## Types of changes
## Checklist:


cihangir:
This patch should be applied to make context work in system

https://gist.github.com/cihangir/54c24506e12a5f9bdf7914e0925939f9

cc/ @sinan

gokmen:
@sinan what is the status about this? there is a conflict currently, fyi.

sinan:
we are not going to do it with hasoffers, but there are some nice changes here that I can cherry pick. Let's keep it a little bit more as a reminder to me.

PR #9049, opened 2016-09-19
kloud: new plugin api

rjeczalik:
~~deps: https://github.com/koding/koding/pull/9004~~


rjeczalik:
moved to https://github.com/koding/koding/pull/9052

PR #9058, opened 2016-09-19
Replace selenium test stack with Rainforest QA

szkl:


gokmen:
Instead of removing existing tests we can disable them can't we?

szkl:
Yes, we can. Initial attempt was to get rid of them 😄

usirin:
👍 for removing them completely, after some time they are gonna become obsolete and outdated anyway.

PR #9066, opened 2016-09-20
Payment small fixes

cihangir:


cihangir:
cc/ @usirin @gokmen

client/app/lib/redux/services/payment.coffee
```diff
@@ -36,6 +36,8 @@ exports.fetchSubscription = fetchSubscription = pickData ->
   client.get Endpoints.SubscriptionGet
 
 exports.createSubscription = createSubscription = pickData (params = {}) ->
+  if not params.trialEnd
+    params.trialEnd = (new Date()).getTime() + (30 * 24 * 60 * 60 * 1000)
```
usirin: minor

we can shorten this with using `Date.now()` which is a shorter version of `(new Date()).getTime()`

usirin:
:metal: 👍

client/app/lib/redux/services/payment.coffee
cihangir: i know js fu +(new Date) works too ;)

go/src/socialapi/workers/payment/customer.go
```diff
@@ -58,7 +62,9 @@ func DeleteCustomerForGroup(groupName string) error {
 	return modelhelper.UpdateGroupPartial(
 		modelhelper.Selector{"_id": group.Id},
 		modelhelper.Selector{
-			"$unset": modelhelper.Selector{"payment.customer.id": ""},
+			// deleting customer deletes everything belong to that customer in stripe,
+			// so say we all
+			"$unset": modelhelper.Selector{"payment": ""},
```
rjeczalik: out of curiosity - why `modelhelper.Selector` and not just `bson.M`?

client/app/lib/redux/services/payment.coffee
usirin: Whoa, i had no idea about this :D I guess it first applies `toString` and then `+` makes it a number?

But my comment was mainly because I felt that `Date.now()` is more readable than `(new Date()).getTime()`

rjeczalik:
@cihangir 👍

go/src/socialapi/workers/payment/customer.go
cihangir: we had two motivations back in time 
1) we wanted one less import line and added Selector 
2) bongo in nodejs side has "selector" and "options" https://github.com/koding/koding/blob/5956d6ca5f804e58452fe2aea548328fdb8acff8/node_modules_koding/jraphical/lib/module.coffee#L19

client/app/lib/redux/services/payment.coffee
usirin: TIL `+(variable)` is how [asm.js](https://github.com/dherman/asm.js/) casts everything into numbers.

PR #9067, opened 2016-09-20
Change team modal: minor updates

alex-ionochkin:
This PR includes the following changes on `Change Team` modal:
- it fixes a lack of bottom indent if the list is scrollable
- it puts the current team at the top of the list, other teams are sorted alphabetically

/cc @ftasdelen

### DIFF
--- a/client/app/lib/changeteam/controller.coffee
+++ b/client/app/lib/changeteam/controller.coffee
@@ -15,6 +15,16 @@ module.exports = class ChangeTeamController extends KodingListController
       whoami().fetchRelativeGroups (err, groups) ->
         return  if showError err
         groups = groups.filter (group) -> group.slug isnt 'koding'
+
+        { groupsController } = kd.singletons
+        currentGroup = groupsController.getCurrentGroup()
+        groups = groups.sort (group1, group2) ->
+          return -1  if group1.slug is currentGroup.slug
+          return 1  if group2.slug is currentGroup.slug
+          return -1  if group1.slug < group2.slug
+          return 1  if group1.slug > group2.slug
+          return 0
+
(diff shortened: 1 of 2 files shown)


client/app/lib/changeteam/controller.coffee
```diff
@@ -15,6 +15,16 @@ module.exports = class ChangeTeamController extends KodingListController
...
         groups = groups.filter (group) -> group.slug isnt 'koding'
+
+        { groupsController } = kd.singletons
+        currentGroup = groupsController.getCurrentGroup()
+        groups = groups.sort (group1, group2) ->
+          return -1  if group1.slug is currentGroup.slug
+          return 1  if group2.slug is currentGroup.slug
+          return -1  if group1.slug < group2.slug
+          return 1  if group1.slug > group2.slug
+          return 0
```
usirin: This can be shortened:

```coffee
switch
  when group1.slug <= group2.slug then -1
  when group1.slug > group2.slug then 1
  else 0
```

gokmen: or `_.sortBy`

```coffee
groups = _.sortBy groups, ['slug']
```

usirin: @gokmen even better. @alex-ionochkin discard my comment, let's go with `lodash`

sinan:
Looks good to me! if you fix that sort we can take this in. thx,

alex-ionochkin:
PR is updated

gokmen:
can we make this overlay darker to make the change modal more visible? like this:

image

cc/ @alex-ionochkin @ftasdelen @sinan

PR #9069, opened 2016-09-20
New Sidebar

hakankaradis:
This pr is wip right now.


client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+  waterfall([
+    (next) ->
+      dispatch {
+        types: [LOAD.BEGIN, LOAD.SUCCESS, LOAD.FAIL]
+        bongo: -> template.generateStack().then (result) ->
+          { stack, results: { machines } } = result
+          instances = []
+          instances.push stack
+          machines.forEach (machine) -> instances.push machine.obj
```
usirin:
```coffee
instances = machines.map (machine) -> machine.obj
```

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+          next null, machines[0].obj
+          reactor.dispatch 'GENERATE_STACK_SUCCESS', { template, stack }
```
usirin: While doing this change, I think we need to remove all dependent actions into redux from `nuclear`. So ideally you wouldn't have to do this, but instead you would handle this via redux.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+          return instances
+        }
+
+    (machine, next) ->
+      console.log '???? ', machine, ' should be null amk', machine.label
```
usirin: remove

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+      new kd.NotificationView { title: 'Stack generated successfully' }
+      reloadIDE machine.label
+      next()
+  ])
+  console.log 'something'
```
usirin: remove

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+
+openOnGitlab = (stack) ->
```
usirin: This can be a key for `mapDispatchToProps`. This doesn't have to do anything with this module. Instead move this function into the container component.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+  remoteUrl = stack.getAt ['config', 'remoteDetails', 'originalUrl']
+  kd.singletons.linkController.openOrFocus remoteUrl
+
+
+handleRoute = (route) ->
```
usirin: Same here, this doesn't belong here. Ideally we should have a module fore handling router state, but right now using `kd.singletons.router.handleRoute` in the container components should be good enough.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+  kd.singletons.router.handleRoute route
+
+
+setAccess = (template) ->
+
+  return  unless template
+
+  return {
+    types: 'UPDATE'
+    promise: -> new Promise (resolve, reject) ->
```
usirin: Again, why do you need another promise to wrap a bongo call?

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+            machines.forEach (machine) -> result.push machine
+            handleRoute '/IDE'
+            return resolve result
+
+          , followEvents = no
+  }
+
+
+reloadIDE = (machineSlug) ->
```
usirin: This doesn't belong here either. Move this to container.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+  console.log 'machine Slug ', machineSlug
+  { computeController } = kd.singletons
+  computeController.reloadIDE machineSlug
+
+
+reinitStack = (stack, machines, template) -> (dispatch) ->
```
usirin: I don't line changing the global `action` signature. Maybe we should pass the `dispatch` function with `getState` in promise middleware [here].(https://github.com/koding/koding/blob/master/client/app/lib/redux/middleware/promise.coffee#L30)

Something like `promise({ dispatch: store.dispatch, getState: store.getState })`

So that our `promise` keys can get `dispatch` and `getState` injected.

That way what we would do is:
```coffee
reinitStack = (stack, machines, template) ->
  return {
    types: [REINIT.BEGIN, REINIT.SUCCESS, REINIT.FAIL]
    promise: ({ dispatch }) ->
      dispatch(destroyStack stack, machines, 'reinitStack').then ->
        dispatch(initializeStack(template) # this requires a change to `initializeStack` as well.
  }
```

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+  return  unless template or not stack
```
usirin: If you want to trigger catch handlers you need to throw here:

```coffee
throw new Error "template or stack can't be `undefined`"  unless template or not stack
```

Returning will be result in hitting the success handler, not the rejection handler.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+
+stackProvider = (stack) ->
+
+  { config, credentials } = stack
+  return 'managedVM'  unless config
+  { requiredProviders } = config
+  for selectedProvider in requiredProviders
+    break  if selectedProvider in ['aws', 'vagrant']
+
+  selectedProvider ?= (Object.keys credentials ? { aws: yes }).first
```
usirin: `? {aws.yes}` part will never work since `Object.keys` will return an array no matter what. So checking the first element would be a better approach here:

```coffee
[key] = Object.keys credentials
selectedProvider ?= key ? 'aws'
```

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+  selectedProvider ?= 'aws'
+
+  return selectedProvider
+
+
+stacksRevisionStatus = (state) ->
+
+  state.stacksAndRevisions
```
usirin: one line please.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+  sidebarStacks
+  stacksRevisionStatus
+  (stacks, status) ->
+
+    return unless stacks and status
+
+    return _.mapValues stacks, (stack, key) ->
+      menuItems = {}
+
+      if stack.bongo_.constructorName is 'JComputeStack'
```
usirin: Can you move this functionality into its own function

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+  (machines) ->
+
+    _.values(machines).filter (machine) ->
+      return no  unless machine.users.length > 1
+
+      # dont show sharedVMs to its owner
+      { profile: { nickname } } = whoami()
+      ownerNickname = getMachineOwner machine, yes
+      nickname isnt ownerNickname
+)
```
usirin: Ideally we want to use `bongo.byId 'JAccount', whoami()._id` here, but since `whoami` already has the data, it makes sense to use it here. 

One thing about that, the instance returned from `whoami` might not be the actual `JAccount` instance but rather a plain object.

Right now it's ok, but let's find a better solution without requiring to use additional utils for these kind of stuff.

client/app/lib/redux/modules/sidebar/stacks.coffee
```diff
@@ -0,0 +1,306 @@
...
+destroyStack = (stack, machines, type = 'deleteStack') ->
+
+  return  unless stack
+
+  { computeController, appManager } = kd.singletons
+
+  return {
+    types: [REMOVE.BEGIN, REMOVE.SUCCESS, REMOVE.FAIL]
+    promise: -> new Promise (resolve, reject) ->
+      computeController.ui.askFor type, {}, (status) ->
```
usirin: I know this is probably a copy/paste from somewhere else, but this kind of heavy implementations should be behind its own class/provider/service etc. I don't like doing `appManager` specific things under here, or even showing a `ui` from `computeController`. Actions should stay as pure as possible by accepting injected arguments via special keys. `promise, bongo, stripe, payment, etc.`. Let's try to figure out how we can move these operations into somewhere but use it here.

client/app/lib/util/getMachineOwner.coffee
```diff
@@ -1,13 +1,14 @@
...
-module.exports = (machine) ->
+module.exports = (machine, revive = no) ->
 
-  switch machine.get 'provider'
+  provider = if revive then machine.provider else machine.get 'provider'
+
+  switch provider
     when 'koding', 'managed'
-      return machine.get('credential')
+      return if revive then machine.credential else machine.get 'credential'
```
usirin: Why? I mean what does this decision have to do with `revive` argument?

usirin (changes requested):
Let's fix these changes first, then I will make another review pass.

client/app/lib/redux/modules/sidebar/stacks.coffee
gokmen: I promise I won't put a promise on top of the existing promises if there is no need to promise on that promise. (and a kitten died somewhere)

sibblegp: I promise I will promise to promise something if I promise to promise it.

gokmen: and one more dead...

hakankaradis: I was just checking how promise middleware handle a promise inside of promise. 
Forgot to remove it

client/app/lib/providers/computecontroller.coffee
```diff
@@ -1245,6 +1249,28 @@ module.exports = class ComputeController extends KDController
           callback()
     else showError 'Failed to share credential'
 
+  makeTeamDefault2: (template, credential, callback) ->
```
hakankaradis: from wip comment

client/app/lib/redux/modules/sidebar/stacks.coffee
hakankaradis: 👍

usirin (approved):

Issue #9084, opened 2016-09-21
vendor: update koding/terraform


rjeczalik:
Branch - https://github.com/koding/koding/tree/tf-update. Reworked terraformer to use go-plugin, one more change is required to make it work:

Terraformer kite needs to be reworked, looks like binary plugins are not meant to be executed directly, it worked so far, but with new version the plugins explicitly exit with:

```
2016/09/21 15:51:17 [DEBUG] plugin: waiting for RPC address for: /home/rjeczalik/koding/go/bin/terraform-provider-null
2016/09/21 15:51:17 [DEBUG] plugin: terraform-provider-null: This binary is a plugin. These are not meant to be executed directly.
2016/09/21 15:51:17 [DEBUG] plugin: terraform-provider-null: Please execute the program that consumes these plugins, which will
2016/09/21 15:51:17 [DEBUG] plugin: terraform-provider-null: load any plugins automatically
2016/09/21 15:51:17 [DEBUG] plugin: /home/rjeczalik/koding/go/bin/terraform-provider-null: plugin process exited
2016/09/21 15:51:17 [ERROR] root: eval: *terraform.EvalInitProvider, err: plugin exited before we could connect
```

cihangir:
reminder here: magic cookie of terraform

https://github.com/koding/koding/blob/ba0fa3efcbcb26eb1ce7aa5612dc405db37e32a8/go/src/vendor/github.com/hashicorp/terraform/plugin/server.go#L26
https://github.com/koding/koding/blob/ba0fa3efcbcb26eb1ce7aa5612dc405db37e32a8/go/src/vendor/github.com/hashicorp/terraform/plugin/client.go

PR #9086, opened 2016-09-21
terraformer: enable all providers

rjeczalik:
This PR enables all providers for terraformer.

Context: https://koding.slack.com/files/devrim/F2DFJQY5S/pasted_image_at_2016_09_19_21_12.png

@szkl @cihangir Building all providers increases size of /opt/koding/go/bin from 855M to 1.3G, just FYI as I remember we were running close to the limit of current production block device size.

Related work is to update terraform/terraformer, currently it's blocked by:

https://github.com/koding/koding/issues/9084

Which is going to be done separately to this PR.

### DIFF
--- a/go/build.sh
+++ b/go/build.sh
@@ -75,12 +75,7 @@ export COMMANDS=(
 )
 
 export TERRAFORM_COMMANDS=(
-	vendor/github.com/hashicorp/terraform/builtin/bins/provider-aws
-	vendor/github.com/hashicorp/terraform/builtin/bins/provider-terraform
-	vendor/github.com/hashicorp/terraform/builtin/bins/provider-null
-	vendor/github.com/hashicorp/terraform/builtin/bins/provisioner-file
-	vendor/github.com/hashicorp/terraform/builtin/bins/provisioner-local-exec
-	vendor/github.com/hashicorp/terraform/builtin/bins/provisioner-remote-exec
+	vendor/github.com/hashicorp/terraform/builtin/bins/...
 	koding/kites/cmd/provider-vagrant
 	vendor/github.com/koding/terraform-provider-github/cmd/provider-github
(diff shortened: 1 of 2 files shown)


cihangir:
I think the context of that convo was just to enable non-cloud ones since our UI does not support them ATM. Am i correct @devrim @gokmen ?

cihangir (changes requested):
Thanks for the reminder @rjeczalik  
Lets update the AMI of EB first then merge this cc/ @szkl

gokmen:
@cihangir yes, any resources which is not known by client-side won't be visible from UI but all those resource will work as is.

rjeczalik:
@gokmen Do we want all providers enabled or enable only the non-cloud ones? I think we wanted them all, as in:

```
Koding supports everything Terraform supports.
```

http://www.koding.com/docs/what-is-a-stack

gokmen:
@rjeczalik they will work for sure, but we need to handle them somehow in the ui, like a db resource. we were planning some improvements on that but postponed it for now.

rjeczalik:
So what about we enable all of them now, and roll out incrementaly ui
support?

devrim:
+1




- devrim yasar, ceo co-f @koding




On Thu, Sep 29, 2016 at 11:06 AM -0700, "Rafal Jeczalik" <notifications@github.com> wrote:










So what about we enable all of them now, and roll out incrementaly ui

support?



On Thursday, 29 September 2016, Gokmen Goksel <notifications@github.com>

wrote:

gokmen:
@rjeczalik said that 7 days ago :) https://github.com/koding/koding/pull/9086#issuecomment-248978512

cihangir (approved):

cihangir:
🎉

PR #9087, opened 2016-09-21
Daily Presence System

cihangir:
This PR adds presence info on daily granularity. It is not near-realtime.


client/app/lib/index.coffee
```diff
@@ -15,6 +15,7 @@ localStorage           = require './localstorage'
 isStarted = false
 
 require './styl/require-styles'
+require './util/ping'
```
cihangir: cc/ @sinan

sinan: 👍

sinan (commented):
Looks good to me!

gokmen:
@cihangir not sure if there are other plans but instead of doing this presence check from application can't we use `kontrol` requests for presence checks? Since you need to make a call to kontrol in any case we can keep track of presence data based on those calls which will also include activity information of `kd` users as well.

rjeczalik:
@gokmen I think these are two distinct use-cases here:

- socialapi/postgre stores historical presence information
- kontrol knows current presence status

I'd also want to have kontrol serve presence information, but I think we could discuss that separately outside this PR. Wdyt?

cihangir:
https://github.com/koding/koding/pull/9087#issuecomment-248775226

Thanks for the comment @gokmen @rjeczalik 

Like @rjeczalik mentioned we still need the functionality of this PR (persistent presence system) but we can always change and tweak the pinging system. https://github.com/koding/koding/pull/9087/files#diff-cbfd311ca6240cdeb99f188fe2d58f8dR34

And technically there are some blockers on using kloud/kontrol requests for presence, like the scope of the request, we dont know in which group the action is taken, yet.

gokmen:
@rjeczalik ah I meant to use `kontrol` calls for keeping track of activity instead of pinging from client app.

@cihangir sure, we need to keep this information somewhere but all I was saying not to make another call to check if user is active or not since we already have a continuous call point there with `kontrol` but as you pointed out group information is must there. Can't we get the group from the action requested resource? When one queries their klient kites with the queryId, we'll have the information if resource found (`JMachine` has group info etc.)

cihangir:
> Can't we get the group from the action requested resource? When one queries their klient kites with the queryId, we'll have the information if resource found (JMachine has group info etc.)


Yep, i am planing to do smth like that, we are talking with @rjeczalik on possible solutions. As a first step to start charging users, i added a simple endpoint, eventually i may not remove that endpoint too since we also need to charge for users that are on koding - not necessarily they need to build vms, they can join collab without having a VM

cihangir:
@rjeczalik if you merge this i will fix my other PR

go/src/socialapi/workers/cmd/presence/main.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+var (
+	name = "Presence"
+)
+
+func main() {
+	r := runner.New(name)
+	if err := r.Init(); err != nil {
+		fmt.Println(err)
```
rjeczalik: shouldn't it exit with non-success code when init fails? `log.Fatal` or `os.Exit`

rjeczalik (approved):

rjeczalik:
```
# socialapi/workers/payment
go/src/socialapi/workers/payment/customer.go:160: undefined: modelhelper.GetDeletedMemberCountByGroupId
```

@cihangir ^

go/src/socialapi/workers/cmd/presence/main.go
cihangir: fixed all occurences

rjeczalik:
Build failed due to ongoing issue with papertrail:

```
Uncaught AssertionError: expected 500 to equal 200
      + expected - actual

      +200
      -500
      
    at Request._callback (/opt/koding/servers/lib/server/handlers/api/logs.test.coffee:160:36)
    at _stream_readable.js:944:16
  
```

@cihangir I'm merging this, the above will need to be fixed separately.

PR #9095, opened 2016-09-22
Fix coffeelint problems after node/npm upgrade

usirin:
I had to switch to ignore for `eol_last` and `newlines_after_classes`. I couldn't find a proper solution as the problem appeared unconclusive (mainly due to the throwing errors while there is no error)

This needs a further exploration but not right now.


usirin:
Will open when i got back to this.

PR #9096, opened 2016-09-22
Create plan on team create

usirin:
Right now team plan is not created at team creation, this PR makes sure that team has a plan if the config is not `default`
## How Has This Been Tested?

Manually and with unit tests
## Types of changes

## Checklist:

/cc @gokmen @sinan @szkl

### DIFF
--- a/servers/lib/server/handlers/createteam.coffee
+++ b/servers/lib/server/handlers/createteam.coffee
@@ -9,6 +9,7 @@ KONFIG                                  = require 'koding-config-manager'
 { hostname, environment }               = KONFIG
 { getClientId, handleClientIdNotFound } = require './../helpers'
 { validateTeamDomain }                  = require '../../../../workers/social/lib/social/models/user/validators'
+createPaymentPlan                       = require './createpaymentplan'
 
 module.exports = (req, res, next) ->
 
@@ -214,7 +215,7 @@ createGroupKallback = (client, req, res, body) ->
 afterGroupCreateKallback = (res, params) ->
 
   { JUser, JTeamInvitation, Tracker } = koding.models
-  { body : { slug, invitees }, client,  username } = params
+  { body : { slug, invitees, coupon }, client,  username } = params
 
   return (err, group) ->
     if err or not group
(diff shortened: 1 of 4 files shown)


servers/lib/server/handlers/createteam.coffee
```diff
@@ -224,10 +225,18 @@ afterGroupCreateKallback = (res, params) ->
...
       # add other parallel operations here
-      (fin) -> createInvitations client, invitees, (err) ->
+      (fin) ->
+        createInvitations client, invitees, (err) ->
           console.error 'Err while creating invitations', err  if err
           fin()
 
+      (fin) ->
+        params = { sessionToken: client.sessionToken }
+        params.coupon = coupon  if coupon
```
cihangir: thanks for adding coupon support 👍

servers/lib/server/handlers/createteam.test.coffee
```diff
@@ -14,6 +14,7 @@ KONFIG                              = require 'koding-config-manager'
 { testCsrfToken }                   = require '../../../testhelper/handler'
 { generateRegisterRequestParams }   = require '../../../testhelper/handler/registerhelper'
 { generateCreateTeamRequestParams } = require '../../../testhelper/handler/teamhelper'
+{ Status }                          = require '../../../../client/app/lib/redux/modules/payment/constants'
```
cihangir: woo–hoo using client side code on backend :D 

I dont know if it is good or bad tho in general 😅

cihangir: @sinan @gokmen wdtyt?

sinan: they're just `constants` i don't see why not.

sinan: only problem would be that if we split the codebase in the future we'd need to duplicate them, which is also fine imo.

gokmen: agree with @sinan 👍

PR #9097, opened 2016-09-22
Fix redux react problems

usirin:
This should fix all the errors in the console, i will take a closer look to the ongoing branches.

Once, this and #9096 are merged, we can do a new deploy.

/cc @diclee


gokmen:
@usirin rebased with `master` restarted tests, fyi.

gokmen (approved):

PR #9101, opened 2016-09-22
Workers: Disable gowebserver and use node web server instead

gokmen:
Currently gowebserver is not using updated model data and this causing some unexpected issues for client side application. To prevent that until we fix those issues or decide on something else this PR disables gowebserver in config and make the node webserver as default handler for all requests.

cc/ @szkl @usirin @hakankaradis


szkl (approved):
Looks good to me, overall.

cihangir:
i will clean up nginx.coffee later

cihangir:
i thought you also removed the gowebserver's code :D

gokmen:
@cihangir ah I didn't I left it for you :)

PR #9103, opened 2016-09-23
client: use global.currentGroup instead of bongo module

usirin:
In sandbox, group is not revived and hasn't been put to the `bongo` reducer's state.

This pr makes it so that it only reads from the `globals.currentGroup` so; no `ReferenceError`


gokmen (approved):

usirin:
It's merge ready, we will test this on sandbox.

Issue #9106, opened 2016-09-23
Email with confused empty team list is sent when user doesn't participate any team


alex-ionochkin:
## Expected Behavior

There should be a notification that user doesn't participate any team when they request a list of their teams
## Current Behavior

User gets email with content on the screenshot
https://uploads.intercomcdn.com/i/o/10765279/5f0efe779536acb7c92c7c88/IMG_0132.PNG
## Steps to Reproduce (for bugs)

Go to `/Teams/FindTeam` and request a team list for user who doesn't participate any team
##

diclee:
Cannot verify this issue because all of my test users participate in at least a team. And currently registration of a new user is only performed while creating a team. Closing this issue. If it's reported again we can open and investigate further.

PR #9110, opened 2016-09-23
Find team: show a notification if team list is empty

alex-ionochkin:
## Description

This PR avoids sending email to user if their team list is empty. Instead, it shows a notification saying that user doesn't participate in any team
## Motivation and Context

https://github.com/koding/koding/issues/9106
## Types of changes

### DIFF
--- a/client/landing/site.landing/coffee/teams/findteamview.coffee
+++ b/client/landing/site.landing/coffee/teams/findteamview.coffee
@@ -6,6 +6,9 @@ MainHeaderView = require './../core/mainheaderview'
 FindTeamForm   = require './findteamform'
 
 EMPTY_TEAM_LIST_ERROR = 'Empty team list'
+SOLO_USER_ERROR = 'Solo user detected'
+
+FAREWELL_SOLO_URL = 'https://www.koding.com/farewell-solo'
 
 track = (action) ->
 
@@ -57,7 +60,7 @@ module.exports = class FindTeamView extends kd.TabPaneView
     utils.findTeam email,
       error       : (xhr) =>
         { responseText } = xhr
-        @showNotification @formatServerError responseText
+        @handleServerError responseText
         @form.button.hideLoader()
       success     : =>
         @form.button.hideLoader()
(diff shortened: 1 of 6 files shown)


client/landing/site.landing/coffee/teams/findteamview.coffee
```diff
@@ -55,21 +57,32 @@ module.exports = class FindTeamView extends kd.TabPaneView
...
-          content  : 'We\'ve sent you a list of your teams.'
-          duration : 4500
+        @showNotification 'Check your email', 'We\'ve sent you a list of your teams.'
 
         kd.singletons.router.handleRoute '/'
 
 
+  formatServerError: (err) ->
+
+    return 'You don\'t participate in any team'  if err is EMPTY_TEAM_LIST_ERROR
```
sinan: `"We couldn't find any teams that you have joined or was invited!"`

sinan (changes requested):

client/landing/site.landing/coffee/teams/findteamview.coffee
sinan: also please make sure that the last line of the email is only shown to people who participate more than one teams.

img

alex-ionochkin: The screenshot you mentioned is not available but I guess you meant this - http://take.ms/fuFaP. We can also change `You can sign in or join to the following teams:` to `You can sign in or join to the following team:` if user has the only team. Does it make sense?

servers/lib/server/handlers/findteam.coffee
```diff
@@ -24,24 +28,32 @@ module.exports = (req, res) ->
...
-        next err, account, groups
+        next err, user, account, groups
 
-    (account, groups, next) ->
+    (user, account, groups, next) ->
       groups = groups.filter (group) -> group.slug isnt 'koding'
 
+      if not groups.length
+        return next(
+          if user.lastLoginDate < FAREWELL_SOLO_DATE
```
alex-ionochkin: @sinan, here I detect old solo user checking if they don't have any group and haven't logged in since the day when solo login was closed (it's a date of solo login blocking changes deployment - 07/19/16). If it's a solo user, they are redirected to farewell-solo page.
However, I'm not sure this check is enough. Maybe additionally we need to check that user is not blocked in any group?

client/landing/site.landing/coffee/teams/findteamview.coffee
sinan: yeah definitely, we should also fix the pluralization.

servers/lib/server/handlers/findteam.coffee
sinan: yes that makes sense, we shouldn't show the disabled teams in the team list.

alex-ionochkin:
PR has been updated. I've added a check for solo users to exclude users who are blocked in any group. Also, I made a change team modal overlay darker

alex-ionochkin:
change-team-modal

sinan (approved):
👍 looks good, thx.

PR #9121, opened 2016-09-25
payment: fix for missing wg.Done() function call

ppknap:
This PR fixes potential deadlock issue which may happen after an early return statement [here](https://github.com/koding/koding/blob/master/go/src/socialapi/workers/payment/customer.go#L133). 

Depends on: #9119

/cc @cihangir @rjeczalik 
## Description

If an error occurs, we set `infoErr` variable and return early from `withCheck` function. In such case, `wg.Done` function will not be called. `sync/errgroup` package was used in order to simplify synchronization logic.   
## Motivation and Context

Potential deadlock.
## How Has This Been Tested?

Manualy, running `payment` package unit tests.

none
## Types of changes
## Checklist:

### DIFF
--- a/go/src/glide.lock
+++ b/go/src/glide.lock
@@ -921,6 +921,10 @@ imports:
   - internal/timeseries
   - trace
   - websocket
+- name: golang.org/x/sync
+  version: 316e794f7b5e3df4e95175a45a5fb8b12f85cb4f
+  subpackages:
+  - errgroup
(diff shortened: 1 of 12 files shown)


rjeczalik:
@ppknap lgtm 👍

cihangir (approved):

cihangir:
Thanks @ppknap 

I found this package last week and this is an obvious use case for these cases!

PR #9123, opened 2016-09-26
go/payment: allow non-admins to fetch sub info

cihangir:
Allow users to see subscription info


go/src/socialapi/workers/payment/api/subscription.go
```diff
@@ -23,10 +23,6 @@ func DeleteSubscription(u *url.URL, h http.Header, _ interface{}, context *model
 
 // GetSubscription gets the subscription of group
 func GetSubscription(u *url.URL, h http.Header, _ interface{}, context *models.Context) (int, http.Header, interface{}, error) {
-	if err := checkContext(context); err != nil {
```
rjeczalik: Shouldn't we test for `context.IsLogged()` here instead?

cihangir: you are absolutely right

cihangir: fixed.

rjeczalik (approved):

Issue #9021, opened 2016-09-14
cmd/tunnelserver: add path routing


rjeczalik:
Currently each new tunnel is routed based on a hostname. New client is assigned a random session ID, so the tunnel is reachable from `<sessionID>.tunnel.example.com`. In order to have host-based routing working, the tunnelserver needs to manage DNS entries, that's why it requires Route53 credentials to do so.

Extend tunnelserver to do path-based routing, so each new tunnel is accessible under `tunnel.example.com/<sessionID>`. This mode would allow for dropping dependency on Route53.

The https://github.com/koding/tunnel needs to be extended as well, as currently the proto allows for only sending virtualHost to the user, and it'd need to send full URL instead.

Related: https://github.com/koding/koding/issues/8999 https://github.com/koding/koding/issues/8920

/cc @twk3

rjeczalik:
/cc @gokmen This would also require change in `proxifyTransport` as the tunnel urls would change - e.g. the path no longer would be `/klient/kite` etc.

The proxifyTransport would need to be configurable, to take tunnelserver domain name or IP address in order to detect tunnelled kites so it can use discover API for local routing feature in vagrant.

Issue #8920, opened 2016-08-31
koding: make tunnelserver a part of open-source environment


rjeczalik:
Currently installing kd fails in open-source environments due to mismatched keys - klient tries to connect to t.koding.com with kite.key generated by user's key pair, which is not going to work.

## Expected Behavior

KD should install successfully in open-source environment.

## Current Behavior

KD does not install successfully in open-source environment.

## Possible Solution

To make kd working in open-source environment we should:

- start tunnelserver with docker-compose
- pass tunnel-kite-url to kloud so it uses correct one for vagrant stacks
- adapt klient/install.sh to allow for overwriting -tunnel-kite-url

## Steps to Reproduce (for bugs)

Related https://github.com/koding/koding/issues/8919.

rjeczalik:
Related https://koding-dev.slack.com/archives/general/p1472460267000030.

CRACKBOY:
thank you @rjeczalik for you help.
Klient is working and bound just on IPv6...
tcp6       0      0 :::56790                :::*                    LISTEN      11856/klient

rjeczalik:
@CRACKBOY The `::56790` is another HTTP server of klient, the one that does TLS termination for web IDE communication. It is started regardless klient is able to get up or not.

CRACKBOY:
ok @rjeczalik 
here what happened: 
               after getting kite and kd installed, Kd trying to consulte 127.0.0.1:56789/kite which expected to be exposed by klient. but klient bound on 56789 on IPv6 as I said in the previews comment.

here, what happened exactly: http://pastebin.com/mS9Ugjim

rjeczalik:
@CRACKBOY Klient starts two servers:

- TLS one on :56790, dual stack [#1](https://github.com/koding/koding/blob/master/go/src/koding/klient/tunnel/tlsproxy/tlsproxy.go#L123)
- kite one on :56789 (the api server I mentioned above) , IPv4 only [#1](https://github.com/koding/koding/blob/master/go/src/koding/klient/app/klient.go#L629) [#2](https://github.com/koding/kite/blob/c81a54521915fd3f4a8b9ee20d10bf2c158dc500/server.go#L72)

Because of the problem described in this issue, the api server does not start and that's why it does not respond to `127.0.01:56789/kite` requests, which are used by KD as a healthcheck.

This is roughly how it's supposed to work. :-)

CRACKBOY:
Very clear, Thank you @rjeczalik

[...]

PR #9140, opened 2016-09-28
tunnelserver: drop Route53 dependency on tunnelserver

rjeczalik:
This PR drops Route53 dependency requirement on tunnelserver.

The tunnelserver requires a domain to be set up for it, in order to make it possible to Host-route tunnels. Tunnelserver requires two records
- e.g. A record, pointing to a single tunnelserver instance, e.g. `tunnel.intra.company.com`
- a CNAME wildcard record, that points to the above record: `*.tunnel.intra.company.com` -> `tunnel.intra.company.com`

If Route53 credentials and hosted zone argument are passed for the tunnelserver, it is going to take care of upserting proper records on it's own.

If no Route53 credentials are passed, tunnelserver assumes the domain was set up elsewhere and passed with `-basevirtualhost` parameter. 

The domain requirement is going to be dropped by https://github.com/koding/koding/issues/9021.

Related to https://github.com/koding/koding/issues/8920.

### DIFF
--- a/go/src/koding/kites/tunnelproxy/server.go
+++ b/go/src/koding/kites/tunnelproxy/server.go
@@ -34,9 +34,9 @@ import (
 type ServerOptions struct {
 	// Server config.
 	BaseVirtualHost string `json:"baseVirtualHost"`
-	HostedZone      string `json:"hostedZone" required:"true"`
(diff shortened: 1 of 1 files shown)


cihangir (approved):
LGTM

PR #9133, opened 2016-09-27
Start charging after first month free & Cancelling should charge immediately.

cihangir:
## Description

We should start charging our users after their ~~trial~~ free period ends. So their presence info will be collected after first moth ends. During free month, set presence info's is processed as true.

On cancelling subscription, we should immediately charge and cancel the sub.
## Motivation and Context

See pricing SRS
## How Has This Been Tested?

Has unit tests
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #9146, opened 2016-09-29
Members can read sub

usirin:
waiting for #9133


client/app/lib/redux/modules/payment/info.coffee
```diff
@@ -0,0 +1,61 @@
...
+  (billingDate) -> dateDiffInDays(new Date(billingDate), new Date)
+)
+
+endsAt = createSelector(
+  nextBillingDate
+  (billingDate) -> (new Date billingDate).getTime()
+)
+
+
+module.exports = _.assign reducer, {
```
sinan: pls just export a vanilla js obj.

sinan (changes requested):
minor comment

Issue #9111, opened 2016-09-23
FindTeam: add a captcha on the page


alex-ionochkin:
User should fill in a captcha before sending a team list request

alex-ionochkin:
Captcha should be shown only if user submits the page 3 or more times per day

diclee:
Verify passed on latest.

PR #9150, opened 2016-09-29
Find team: captcha

alex-ionochkin:
## Description

It adds google recaptcha widget to `/Teams/FindTeam` page. Recaptcha appears only if user has submitted the form 3 or more times per day
## Motivation and Context

https://github.com/koding/koding/issues/9111
## Screenshots:

http://recordit.co/nw99GiyXNb
## Types of changes

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -2,6 +2,8 @@ $       = require 'jquery'
 kd      = require 'kd'
 kookies = require 'kookies'
 
+RECATCHA_JS = 'https://www.google.com/recaptcha/api.js?onload=onRecaptchaLoaded&render=explicit'
+
 createFormData = (teamData) ->
 
   teamData ?= utils.getTeamData()
@@ -17,6 +19,12 @@ createFormData = (teamData) ->
         formData[field] = value
 
   return formData
+
+recaptchaLoadCallbacks = []
+window.onRecaptchaLoaded = ->
+
+  callback() for callback in recaptchaLoadCallbacks
+  recaptchaLoadCallbacks = []
 
 
 module.exports = utils = {
@@ -335,16 +343,6 @@ module.exports = utils = {
         callbacks.error.call this, xhr.responseText
 
 
-  findTeam: (email, recaptcha, callbacks = {}) ->
-
-    $.ajax
-      url         : '/findteam'
-      data        : { email, recaptcha, _csrf : Cookies.get '_csrf' }
-      type        : 'POST'
-      error       : callbacks.error
-      success     : callbacks.success
-
-
   getGravatarUrl : (size = 80, hash) ->
 
     fallback = "https://koding-cdn.s3.amazonaws.com/square-avatars/default.avatar.#{size}.png"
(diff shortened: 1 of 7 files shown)


client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -2,6 +2,8 @@ $       = require 'jquery'
 kd      = require 'kd'
 kookies = require 'kookies'
 
+RECATCHA_JS = 'https://www.google.com/recaptcha/api.js?onload=onRecaptchaLoaded&render=explicit'
```
gokmen: typo: `RECAPTCHA_JS`
and we already have it here: https://github.com/koding/koding/blob/master/client/landing/site.landing/coffee/login/AppView.coffee#L18

you can update it there to use it from one place.

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -425,4 +418,18 @@ module.exports = utils = {
...
+
+  loadRecaptchaScript: (callback) ->
+
+    return callback()  if grecaptcha?
+
+    recaptchaLoadCallbacks.push callback
+    recaptchaScript = new kd.CustomHTMLView
+      tagName    : 'script'
+      attributes :
+        src      : RECATCHA_JS
```
gokmen: typo: `RECAPTCHA_JS`

gokmen (changes requested):
let's unify those recaptcha usages site-wide then it looks good to me!

client/landing/site.landing/coffee/core/utils.coffee
alex-ionochkin: Good catch! I created `utils.loadRecaptchaScript()` to use it in all views with recaptcha but missed old signup 😞

alex-ionochkin:
PR has been fixed

sinan:
@gokmen pls approve so that I can merge.

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -425,4 +418,18 @@ module.exports = utils = {
...
+
+    recaptchaLoadCallbacks.push callback
+    recaptchaScript = new kd.CustomHTMLView
+      tagName    : 'script'
+      attributes :
+        src      : RECAPTCHA_JS
+        async    : yes
+        defer    : yes
+
+    recaptchaScript.appendToDomBody()
```
gokmen: wouldn't this create multiple `script` elements in DOM?

gokmen (changes requested):

client/landing/site.landing/coffee/core/utils.coffee
gokmen: nope, it still causes multiple `script` tags. to prevent this on line 426 needs to be changed as follow;

```coffee
recaptchaLoadCallbacks.push callback
return  if recaptchaLoadCallbacks.length > 1
```

sinan: yes, @alex-ionochkin pls make sure that the script isn't added multiple times.

alex-ionochkin: you're right, @gokmen, I didn't take into account that script can be requested during the time it's being loaded but is not loaded yet

alex-ionochkin:
PR has been fixed

gokmen (approved):

PR #9178, opened 2016-10-01
client: update for vms' details

### LINKED ISSUE #9173: Incorrect Amount of RAM and CPUs Displayed in VM Details
## Current Behavior

Incorrect amount of ram and CPUs are displayed in the VM details of the dashboard section

An M4.2xl shows 1GB of RAM and 1xCPU:
image

Go to the Dashboard and open the details of a VM.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38010724-incorrect-amount-of-ram-and-cpus-displayed-in-vm-details?utm_campaign=plugin&utm_content=tracker%2F41989991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F41989991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cansirin:
## Description

Currently, VM's CPU and RAM weren't displayed correctly. This PR updates that information.
## Motivation and Context

Fixes #9173 
## How Has This Been Tested?

Manually

image
## Types of changes


client/home/lib/virtualmachines/components/machineslist/machinedetails.coffee
```diff
@@ -225,8 +225,15 @@ generateSpecs = (machine) ->
...
+    'm4.large'  : '8GB RAM'
+    'm4.xlarge' : '16GB RAM'
+    'm4.2xlarge': '32GB RAM'
   }[type] ? '1GB RAM'
-  specs.push cpu = '1x CPU'
+  specs.push cpu = {
+    'm4.large'   : '2x CPU'
+    'm4.xlarge'  : '4x CPU'
+    'm4.2xlarge' : '8X CPU'
+  }[type] ? '1x CPU'
```
usirin: @gokmen is there a dynamic way that we can get this list?

gokmen: not currently and this existing list also needed to be provider specific, these are only valid for `aws` provider only. I suggest to create a provider specific list somewhere else and not keep all these information in the view.

usirin (approved):

hakankaradis:
@cansirin needs rebase

client/home/lib/virtualmachines/components/machineslist/machinedetails.coffee
```diff
@@ -217,16 +218,17 @@ generateSpecs = (machine) ->
 
   specs = []
 
+  providerName = jMachine.meta?.type
+  instanceType = jMachine.meta?.instance_type ? 't2.micro'
   size = jMachine.meta?.storage_size
 
+  { ram } = providerSpecs[providerName][instanceType]
+  { cpu } = providerSpecs[providerName][instanceType]
```
gokmen: this will fail if vm has different provider like `vagrant` there needs to be defaults for each spec.

client/home/lib/virtualmachines/components/machineslist/machinedetails.coffee
```diff
@@ -217,16 +218,17 @@ generateSpecs = (machine) ->
 
   specs = []
 
+  providerName = jMachine.meta?.type
+  instanceType = jMachine.meta?.instance_type ? 't2.micro'
```
gokmen: `t2.micro` shouldn't be default here, it can be `base-vm` something cc/ @sinan

client/app/lib/constants/providerSpecs.coffee
```diff
@@ -0,0 +1,20 @@
+module.exports =
```
gokmen: let's make this a method takes `provider` as a parameter and returns list of available types

gokmen (changes requested):

client/app/lib/constants/providerSpecs.coffee
usirin: 👍 So @cansirin we need to move this file under `client/app/lib/util`, make it a function that accepts a `provider` argument, and return the list depending on this `provider` argument. And rename it to `makeProviderStacks.coffee`

client/home/lib/virtualmachines/components/machineslist/machinedetails.coffee
usirin: @rjeczalik Can you please direct us to the correct place where we can find the implemented `provider`s and their specs? Especially i am interested in to see what the defaults are for each provider, so that we can have that info and show those to the users correctly.

cansirin:
updated. /cc @gokmen

gokmen (approved):

gokmen:
LGTM 👍 but let's wait for this one for deploy and/or for #9319

Issue #9118, opened 2016-09-25
client: improve error message formatting


rjeczalik:
Context: https://koding.slack.com/files/devrim/F2DV06RG8/pasted_image_at_2016_09_20_16_19.png
## Expected Behavior

The error messages should be formatted with line breaks where necessary.
## Current Behavior

They are not.
## Possible Solution

They should be.
##

PR #9201, opened 2016-10-05
Resource state modal: bug fixes

alex-ionochkin:
## Description

This PR contains error handling fixes in `ResourceStateModal`
## Motivation and Context
- https://github.com/koding/koding/issues/9118
- It fixes VM endless loading animation after team cleanup process. In this case stack status is `Initialized` but VM status is `NotInitialized`. A fix correctly handles such case and shows a proper error message

### DIFF
--- a/client/app/lib/providers/resourcestatemodal/controllers/machineflowcontroller.coffee
+++ b/client/app/lib/providers/resourcestatemodal/controllers/machineflowcontroller.coffee
@@ -34,7 +34,7 @@ module.exports = class MachineFlowController extends kd.Controller
       .then (response) =>
         @onDataLoaded()
         @updateStatus
-          status     : response.state
+          status     : response.State
           eventId    : machineId
       .catch (err) =>
         @onDataLoaded()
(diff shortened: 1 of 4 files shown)


client/app/lib/providers/resourcestatemodal/controllers/machineflowcontroller.coffee
```diff
@@ -33,8 +34,7 @@ module.exports = class MachineFlowController extends kd.Controller
       .then (response) =>
         @onDataLoaded()
         @updateStatus
-          status     : response.State
-          percentage : response.percentage
+          status     : response.state
```
gokmen: this is fixed in kloud it's still `response.State` cc/ @hakankaradis @rjeczalik

gokmen (changes requested):
this needs to be reverted before merge https://github.com/koding/koding/pull/9201/files#r82023590 rest LGTM

client/app/lib/providers/resourcestatemodal/controllers/machineflowcontroller.coffee
rjeczalik: I'd rather make it `state` in kloud instead.

hakankaradis: Kloud fixed with #9190

alex-ionochkin:
`MachineFlowController` has been fixed according to the latest kloud fix

gokmen (approved):

PR #9236, opened 2016-10-10
object: preserve case of the field name

### LINKED ISSUE #9219: IP address is displayed as '0.0.0.0' even though the VM is on.
## Current Behavior
image

rjeczalik:
This PR changes previous behavior, which created
objects by lowering the case of struct's field.

In order to keep previous behavior set FieldFunc
to a strings.ToLower while creating Builder, like:

  var b = &Builder{
    FieldFunc: strings.ToLower,
  }

Fixes #9219.


ppknap (approved):

cihangir:
Lets not send any pr without tests from now on to increase our coverage everywhere. 
ps: I am not requesting 100% coverage :) just up to reasonable level

PR #9240, opened 2016-10-10
config: escape nginx variables in .env.sh script

ppknap:
This PR escapes '$' character for nginx-like variables in .env.sh. 
## Motivation and Context

Currently, `generateShellEnv.coffee` doesn't escape '$' character in nginx-like variables which generates invalid values. Eg.

``` shell
$ export KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS=${KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS:-http://kontrol/$1$is_args$args}
$ env | grep KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS
KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS=http://kontrol/
```

This fix will escape all `$` characters that precede either a lower case character or a digit(`.env.sh` is not meant to be called with command line arguments)

Without escaping it is also not possible to `source .env.sh`
## How Has This Been Tested?

Manually.
## Types of changes

### DIFF
--- a/config/generateShellEnv.coffee
+++ b/config/generateShellEnv.coffee
@@ -4,6 +4,10 @@ module.exports.create = (KONFIG, options = {}) ->
   env = ''
 
   add = (name, value) ->
+    # Escape $ sign when it precedes lower-case character which means that this
+    # is not a bash variable but nginx one.
+    value = (value.replace /(\$)[a-z0-9]/g, (match) -> "\\#{match}") if typeof value is 'string'
+
     env += "export #{name}=${#{name}:-#{value}}\n"
 
   traverse.paths(KONFIG).forEach (path) ->
--- a/config/generateShellEnv.coffee
+++ b/config/generateShellEnv.coffee
@@ -6,7 +6,7 @@ module.exports.create = (KONFIG, options = {}) ->
   add = (name, value) ->
     # Escape $ sign when it precedes lower-case character which means that this
     # is not a bash variable but nginx one.
-    value = (value.replace /(\$)[a-z0-9]/g, (match) -> "\\#{match}") if typeof value is 'string'
+    value = (value.replace /\$[a-z0-9]/g, (match) -> "\\#{match}") if typeof value is 'string'
 
     env += "export #{name}=${#{name}:-#{value}}\n"
 


rjeczalik:
Instead of `$ -> \$` wouldn't it be enough to escape with single quotes `a -> 'a'`?

Like:

```bash
$ export KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS=${KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS:-'http://kontrol/$1$is_args$args'}
$ env | grep KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS
KONFIG_WORKERS_KONTROL_NGINX_LOCATIONS_0_PROXYPASS=http://kontrol/$1$is_args$args
```

ppknap:
@rjeczalik i noticed variables like:

```shell
export KONFIG_TERRAFORMER_LOCALSTOREPATH=${KONFIG_TERRAFORMER_LOCALSTOREPATH:-$KONFIG_PROJECTROOT/go/data/terraformer}
```

so your solution will not work.

config/generateShellEnv.coffee
```diff
@@ -4,6 +4,10 @@ module.exports.create = (KONFIG, options = {}) ->
   env = ''
 
   add = (name, value) ->
+    # Escape $ sign when it precedes lower-case character which means that this
+    # is not a bash variable but nginx one.
+    value = (value.replace /\$[a-z0-9]/g, (match) -> "\\#{match}") if typeof value is 'string'
```
cihangir: imo we can completely filter out nginx location parameters, they are not required on configuration context for the app. wdyt @szkl ?

ppknap: Yes, this would be great! I didn't find any usage for them in `.env.sh`

szkl: Agreed. This script definitely needs a way to exclude some properties.

szkl:
@ppknap what sort of issue are you having with sourcing .env.sh right now?

ppknap:
@szkl I wanted to `source .env.sh` in order to locally set an environment for go generate but got the error: `/opt/koding/.env.sh: line 466: $1: unbound variable`. Despite of this, it is always nice to not have broken/unused code in the codebase. :)

szkl:
@ppknap what's your $SHELL --version?

ppknap:
```sh
$ `echo $0` --version
GNU bash, version 4.3.46(1)-release (x86_64-unknown-linux-gnu)
```

szkl:
Weird. I'm (and deployment environments) not having any issues.

```
root@d07439a8384e:/opt/koding# source .env.sh
root@d07439a8384e:/opt/koding# echo $?
0
root@d07439a8384e:/opt/koding# $SHELL --version
GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)
```

ppknap:
Oh, I was sourcing this file in `./go/build.sh` that has `set -u` option which treats unset parameters as an error. That's the reason why this works in shell and doesn't in script. I should have mentioned about this earlier but focused only on escaping `\$` characters. Sorry @szkl

szkl:
@ppknap that's fine. I wanted to figure out your case before proceeding any further :)

rjeczalik:
Do we merge this or close as won't fix?

szkl:
I've decided to close it for now. I can't think of a way to distinguish a shell variable expansion expression from a regular expression. It's definitely a bug but not harmful at the moment.

ppknap:
Not a problem at all. I decided to not use environment variables in go's config.

PR #9246, opened 2016-10-11
Improvements to docker-compose specification

szkl:
- ./run is_ready
- client build service
- client/landing build service

### DIFF
--- a/scripts/bootstrap-container
+++ b/scripts/bootstrap-container
@@ -24,6 +24,12 @@ function run_backend() {
   tail --follow --lines +0 --quiet .logs/*.log
 }
 
+function is_ready() {
+  until [ -x ./run ] && ./run is_ready; do
+    sleep 10
+  done
+}
+
 function run() {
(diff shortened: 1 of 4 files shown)


scripts/bootstrap-container
```diff
@@ -4,37 +4,65 @@ set -o errexit
...
   nginx -c $(pwd)/nginx.conf
   ./run exec supervisord -c supervisord.conf
   sleep 2
+  tail --follow --lines +0 --quiet .logs/*.log
 }
 
-function tail_logs() {
-  tail --follow --lines +0 --quiet .logs/*.log
+function is_ready() {
+  until [ -x ./run ] && ./run is_ready; do
```
cihangir: lets have timeout here

cihangir (approved):

scripts/bootstrap-container
cihangir: 10mins? we can increase if we have false positives

PR #9275, opened 2016-10-13
Auto generated go config package.

ppknap:
This PR removes hard-coded static configuration from config package. Golang `config` package is now auto-generated from values set in ~~`config/goGenerate.coffee`~~ `config/generateKonfig.coffee` module.

`main.production.coffee` file should be updated with following code:

``` coffeescript
options.publicLogsS3BucketName or= 'koding-publiclogs'
options.proxySubdomain or= 'p'
```
## Motivation and Context

Store configuration in one place and allow to generate it based on ./configure options.
## How Has This Been Tested?

Unit tests for config package. (added to wercker)
~~Golden-file based tests made in order to validate go generate results. (added to wercker)~~


go/src/koding/config/base_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+func TestGetBucket(t *testing.T) {
+	var testTmpls = map[string]*template.Template{
+		`buckets.A`:    template.Must(template.New(`buckets.A`).Parse(`{"name":"bucket","region":"reg"}`)),
+		`buckets.Tmpl`: template.Must(template.New(`buckets.Tmpl`).Parse(`{"name":"test-{{.Environment}}","region":"reg-{{.Group}}"}`)),
+		`invalid.Tmpl`: template.Must(template.New(`invalid.Tmpl`).Parse(`{"name":"{{.Missing}}","region":"reg"}`)),
+		`invalid.Json`: template.Must(template.New(`invalid.Json`).Parse(`{"name":bucket,"region":"reg"}`)),
+	}
+
+	tests := []struct {
```
rjeczalik: just a nitpick - if you replace `[]struct{` with `map[string]struct{` and label each test with a description, you could remove `// 0 //` comments + have a nice description of a test-case (`"this one should fail because...": {` etc.)

go/src/koding/config/genconfig.go
```diff
@@ -0,0 +1,233 @@
...
+	"strings"
+	"text/template"
+)
+
+var (
+	input  = flag.String("i", "-", "")
+	output = flag.String("o", "-", "")
+)
+
+var t = template.Must(template.New("").Parse(`// This file is auto-generated. DO NOT EDIT!
```
cihangir: i am not sure why we need this template and all the file and template generation. could you please explain? You are already generating a json file  in `goGenerate`, why not just use json encode/decode? is not it simpler? we are already generating the config for each env while deploying

config/goGenerate.coffee
```diff
@@ -0,0 +1,18 @@
+module.exports = (options, credentials) ->
```
usirin: can we rename this to `generage-go.coffee` or `generateGo.coffee`?

config/main.dev.coffee
```diff
@@ -84,6 +86,9 @@ Configuration = (options = {}) ->
   KONFIG.workers = require('./workers')(KONFIG, options, credentials)
   KONFIG.client.runtimeOptions = require('./generateRuntimeConfig')(KONFIG, credentials, options)
 
+    # Generate static variables for Go.
+  KONFIG.goGenerate = require('./goGenerate')(options, credentials)
```
usirin: Same here.

config/main.sandbox.coffee
```diff
@@ -123,6 +125,9 @@ Configuration = (options = {}) ->
 
   KONFIG.client.runtimeOptions = require('./generateRuntimeConfig')(KONFIG, credentials, options)
 
+    # Generate static variables for Go.
+  KONFIG.goGenerate = require('./goGenerate')(options, credentials)
```
usirin: Same here.

config/generateShellEnv.coffee
```diff
@@ -9,7 +9,7 @@ module.exports.create = (KONFIG, options = {}) ->
   traverse.paths(KONFIG).forEach (path) ->
     node = traverse.get KONFIG, path
     return  if typeof node is 'object'
-    add "KONFIG_#{path.join('_')}".toUpperCase(), node
+    add "KONFIG_#{path.join('_').replace(/\./g, "_")}".toUpperCase(), node
```
szkl: What property had a dot in its name?

config/goGenerate.coffee
szkl: Module name should explain what it generates.

ppknap: @szkl you are right. I followed naming conversions I see files like `workers.cofffee` which also returns an object. According to this, what about `goConfig.coffee` or `generateGoConfig.coffee` (however the former includes a `generate` verb which suggests that we do some actions inside). I'd rather not use `generage-go.coffee` since it's inconsistent with other file names in `config` directory.

rjeczalik: generateGoConfig.coffee?

go/src/koding/config/genconfig.go
ppknap: Depending on my knowledge I assume that current implementation of `koding/config` uses `go generate` and `go-bindata` to provide a way of storing build time configuration directly in binary file. This allows us to have all necessary data in a single executable file.

Since the configuration from this package is used by klient we don't have to provide additional config.json files which may become inconsistent with those used by other parts of the infrastructure (like tunnelserver etc.).

So, as you said, we can still use `go generate` + `json.Unmarshal` + `json.Marshal` + `go-bindata`. And yes, this could be even slightly easier than proposed solution. So why did I choose it over JSON-like one:

##### Simplicity:

There is no difference between this PR and JSON solution. If you want to add a new variable, you only need to edit `dontKnowHowToNameIt.coffee` file.

##### Correctness:

Let say we used JSON solution and have statement like this:

```go
endpoint := config.Builtin.Endpoints.URL("tunel", env)
```

we incorrectly wrote `tunel` instead of `tunnel`. This will compile and may even pass the tests.

Auto-generated functions enforce compiler to check this at compile time, so if we made error like above:

```go
endpoint := config.MustTunelURL(env)
```

the code won't even compile unless we have endpoint.tunel variable defined in `dontKnowHowToNameIt.coffee` file.

##### Maintainability:

Another example, when we want to remove an unused variable from existing configuration. In JSON-like solution we just need to grep the code and remove all references to string we want to remove. But, what if we have code like that:

```go
name := blackBox.GetEndpointName() // this can be everything.
endpoint := config.Builtin.Endpoints.URL(name, env)
```

without going deeper in the logic we can't be sure that removing a variable is not going break something. Static-function-like code enforces us to make such logic explicit:

```go
switch blackBox.GetEndpointName() {
	case "tunnel":
		endpoint = config.TunnelURL(env)
	// ...
}
```

so it will be easy to check if our variable is still needed or not.

TL;DR

I used templates and auto-generated files to improve code maintainability and enforce correctness of configuration variables.

config/generateShellEnv.coffee
ppknap: I really like @cihangir 's idea proposed here #9240 to filter out unused variables. Currently, this script generates > 600 env variables which are mostly unused. I don't use variables generated by this file at all and had to add this logic only because it consumes `goGenerate` object.

`.replace(/\./g, "_")` dosn't change any of existing variables so I think it's safe to left it here.

config/goGenerate.coffee
szkl: @ppknap 👍  I'm okay with these names.

> Since the configuration from this package is used by klient ...

What about generateKlientConfig.coffee?

ppknap: these variables are not only used in Klient so `generateKlientConfig` may be misleading.

config/generateShellEnv.coffee
szkl: I've thought I've left a bug that didn't cause any issues yet :)

Well, most of the environment variables are not directly referenced because they are read into an object called KONFIG in node modules and respective struct types in golang packages during start phase of programs. It provides a way to set configuration in runtime without any configure or build process.

go/src/koding/config/genconfig.go
szkl: I'm not a big fan of auto-generation but if there isn't any way to achieve correctness and maintainability in golang like you've explained then it sounds good to me.

cihangir: i still cant follow the idea, sorry :/ 

What is the problem you see creating a json/toml from the coffee and using multiconfig to read that and using `required` tag?

It is simple, it is correct, it is maintainable - we are using it in all of our workers (> 25 workers)

cihangir (changes requested):
waiting for answer/clarification

go/src/koding/config/genconfig.go
cihangir: >Yes, I'd probably use the solution you provided if I was creating new workers. This is completely fine.

If you could use that for new workers, you can use for the old ones as well.

Yep you are saying we have a different use case but i dont see it as different. I see it as just complex.

The complexity you create comes from the first design decision that was made by trying to use "strings" to get the configuration you are looking for. And to remediate that complexity you are adding another level of complexity by creating templates on the fly and adding them as "generated" files. Another level of complexity you add comes with "code generation" part, not to use json serialization you have implemented 477 more lines of code ( i dont count the lines usually, just stating here because previously you mentioned about LOC )

What i wanted to see with my request was simple config generation on configure step, i am not against using bindata.

>  I see no advantages in such solution.

You will have a completely build-time-safe structs and you can refactor them easily whenever you want. We had great success with using multiconfig on many places.

For the future generations lets have simple and understandable codebase.

config/goGenerate.coffee
```diff
@@ -0,0 +1,28 @@
...
+  endpoints      :
+    ip           : "https://#{options.proxySubdomain}.koding.com/-/ip"
+    ipCheck      : "https://#{options.proxySubdomain}.koding.com/-/ipcheck"
+    kdLatest     : "https://koding-kd.s3.amazonaws.com/{{.Group}}/latest-version.txt"
+    klientLatest : "https://koding-klient.s3.amazonaws.com/{{.Environment}}/latest-version.txt"
+    kloud        : "#{options.publicHostname}/kloud/kite",
+    kontrol      : "#{options.publicHostname}/kontrol/kite",
+    tunnelServer : "#{options.tunnelUrl}/kite"
+  routes         :
+    {
```
cihangir: you can omit brackets

config/goGenerate.coffee
```diff
@@ -0,0 +1,28 @@
...
+module.exports = (options) ->
+  environment    : options.environment
+  buckets        :
+    publicLogs   :
+      name       : options.publicLogsS3BucketName
+      region     : 'us-east-1'
+  endpoints      :
+    ip           : "https://#{options.proxySubdomain}.koding.com/-/ip"
+    ipCheck      : "https://#{options.proxySubdomain}.koding.com/-/ipcheck"
+    kdLatest     : "https://koding-kd.s3.amazonaws.com/{{.Group}}/latest-version.txt"
```
cihangir: this config should be decided on configure step, not on go side.

config/goGenerate.coffee
```diff
@@ -0,0 +1,28 @@
...
+    klientLatest : "https://koding-klient.s3.amazonaws.com/{{.Environment}}/latest-version.txt"
```
cihangir: same here with previous comment

go/src/koding/config/genconfig.go
ppknap: Okay, I will try to simplify this a bit.

config/main.default.coffee
```diff
@@ -45,6 +45,8 @@ Configuration = (options = {}) ->
   options.paymentBlockDuration = 2 * 60 * 1000 # 2 minutes
   options.credentialPath or= "$KONFIG_PROJECTROOT/config/credentials.#{options.environment}.coffee"
   options.clientUploadS3BucketName or= 'kodingdev-client'
+  options.publicLogsS3BucketName or= 'kodingdev-publiclogs'
+  options.proxySubdomain or= 'dev-p2'
```
ppknap: yes, and it stores `dev-t.koding.com`. :)

config/goGenerate.coffee
ppknap: this can be hard to achieve for environments like `sandbox` which gets their version from `development` bucket. Maybe it's better to just create `sandbox` bucket on S3 so we can indeed get rid of {{.Group}}.

config/goGenerate.coffee
ppknap: the same problem as with previous one. This also includes managed/devmanaged environments which can be requested by klient.

config/goGenerate.coffee
rjeczalik: > Maybe it's better to just create sandbox bucket on S3 so we can indeed get rid of {{.Group}}

I don't think it's a good idea - currently we have `production`, `development` and `managed` and `devmanaged` channels (buckets) - we should get rid at least of two managed and do not create more buckets.

ppknap:
Much more simplified version. @cihangir @rjeczalik

go/src/koding/klient/info/publicip/publicip.go
```diff
@@ -64,15 +63,15 @@ func (d *director) EndpointPublicIP() string {
...
+	return config.Builtin.Endpoints.IP
 }
 
 func (d *director) EndpointIsReachable(port string) string {
 	if d.fallback {
 		return "http://rjk.io/test/" + port
 	}
 
-	return config.Builtin.Endpoints.URL("ipcheck", protocol.Environment) + "/" + port
+	return config.Builtin.Endpoints.IPCheck + "/" + port
```
cihangir: this string interpolation can go into config as well and we can get rid of this helper function

go/src/koding/klient/main.go
```diff
@@ -67,15 +67,15 @@ func defaultNoTunnel() bool {
 }
 
 func defaultKeygenURL() string {
```
cihangir: since we have the struct we dont need helper functions

cihangir:
could you please rebase too?

cihangir (approved):

PR #9279, opened 2016-10-13
mongo migrations: add support for migration for mongodb

cihangir:


configure
```diff
@@ -8,7 +8,7 @@ git = require 'git-rev'
 request = require 'request'
 { extend, defaults, clone } = require 'underscore'
 
-defaultEnv = argv.config or process.env.DEFAULTENV or 'default'
+defaultEnv = argv.config or process.env.DEFAULTENV or 'dev'
```
gokmen: this should be reverted.

gokmen (approved):
💯

PR #9297, opened 2016-10-15
Modularize how create team flow works in docs/landing

usirin:
Related with #9227


sinan (approved):
taking in as an interim, better than current solution, we'll work on this later. thank you!

Issue #9266, opened 2016-10-13
kloud: provider improvements


rjeczalik:
A TODO list of improvements for existing kloud providers:

~~\- Improve credential validation; for now `google` and `digitalocean` (*Credential).Valid() returns nil error for a Credential zero-value:~~

```
$ cat /Volumes/rjk.io/aws-dev.json | ./kd c create -f - -team foobar
Creating credential... 
Error communicating with Koding: stack: explicit provider name is required for this credential; matches [aws digitalocean google] (c140dedfe24ab438)
```
- improve generated schemas, adding enums for `azure` (mostly storage-related), `google` and `digitalocean`; /cc'ing @gokmen - I remember there was a feature request to extend schema to  include also `instance type -> (cpu count, memory size)` mapping per provider.

_(more coming)_

ppknap:
@rjeczalik It looks like you need to improve your current wip branch `stack-create`:

I assume that [here](https://github.com/koding/koding/blob/1c98c5aab4161100f8a84c1aa190ec19cb6f1fd3/go/src/koding/kites/kloud/stack/credential.go#L258) you wanted to add:

```go
cred := p.NewCredential()
```

and [here](https://github.com/koding/koding/blob/1c98c5aab4161100f8a84c1aa190ec19cb6f1fd3/go/src/koding/kites/kloud/stack/credential.go#L265-L267) you were going to add provider name into `matching` slice when `err == nil`:

```go
if err := v.Valid(); err == nil {
	matching = append(matching, name)
}
```

gokmen:
@ppknap got this while bootstrap on google provider; 

```
[18:02:50] An error occurred: apply failed with code: 1, output: e:Error loading config: 
Error loading /.../go/data/terraformer/kodingdev-vms-camel-131e94842ee37081a3860285485567c1/main.tf.json: 
position -: resource must be followed by exactly two strings, a type and a name
```

rjeczalik:
@ppknap It's most likely a tf bug, you can apply a workaround like:

https://github.com/koding/koding/blob/master/go/src/koding/kites/kloud/stack/provider/builder.go#L537-L543

ppknap:
@gokmen I made a workaround for this issue. However, there is another bug on hashicorp side. Their hcl parser panics on interface values that store structures. I'm gonna to send them PR with the fix. Until then you will see:

[...]

PR #9299, opened 2016-10-16
provider/google: workaround for hcl parser not supporting struct types behind raw interface

ppknap:
Replace `struct` objects with raw `map[string]interface{}` in google stack.
## Description

Terraform hcl decoder has a bug which causes panics during parsing to structures that contain interfaces pointing to struct types `var x := interface{}(T{})`. This PR replaces structures with `map[string]interface{}` type that is not affected by this bug.
## Motivation and Context

ref: https://github.com/koding/koding/issues/9266
## How Has This Been Tested?

Manually.
## Types of changes


rjeczalik (approved):

cihangir:
👍

PR #9308, opened 2016-10-17
Trollmode remove

mehmetalisavas:
Remove socialapi
## Description

Remove trollmode and dependencies from socialapi 
## Motivation and Context

Required to clean codebase


cihangir (changes requested):
could you remove from coffee files as well, eg: workers.coffee

cihangir (approved):

PR #9322, opened 2016-10-18
Rework get group name from location function

szkl:
Test cases: https://runkit.com/szkl/5805a409e9d9ed00130c9869

Signed-off-by: Sonmez Kartal sonmez@koding.com

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -60,31 +60,40 @@ module.exports = utils = {
 
   getGroupNameFromLocation: (hostname) ->
 
-    { hostname }     = location  unless hostname
-    ipV4Pattern      = '(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.xip\.io$'
-    subDomainPattern = '[A-Za-z0-9_]{1}[A-Za-z0-9_-]+'
-    kodingDomains    = '(?:dev|sandbox|latest|prod)'
+    { hostname } = location  unless hostname
+    basePattern  = (kd.config.domains.main.split ':')[0].replace /\./g, '\\.'
 
-    # Base Domain may include port information which we don't need ~GG
-    baseDomain       = (kd.config.domains.main.split ':')[0]
-    baseDomain       = "#{baseDomain}".replace /\./g, '\\.'
+    predefinedPatterns = [
+      # <team-domain>.<(dev|sandbox|latest|prod)>.koding.com
+      { predict: "(dev|sandbox|latest|prod)\.#{basePattern}" }
 
-    # e.g. [teamName.]dev|sandbox|latest|prod.koding.com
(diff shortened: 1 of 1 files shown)


gokmen (approved):
this is 💯  but we may also need to update main app which might not be handled like this; 
https://github.com/koding/koding/blob/master/servers/lib/server/setsession.coffee#L12-L17 just adding for reference.

gokmen:
@szkl I've restarted tests but it failed again in the same place, can u check?

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -60,31 +60,40 @@ module.exports = utils = {
...
+        labelFn: (hostname) ->
+          labels = hostname.split('.')
+          labels.splice -6, 6 # 'A.B.C.D.xip.io'.length
+          return labels.pop()
+      }
+    ]
+
+    for pattern in predefinedPatterns
+
+      if typeof pattern is 'string'
```
cihangir: instead of typeof checkings, we could write `predict` to each pattern :D

szkl: Much simpler.

szkl:
🏁

gokmen (approved):

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -60,31 +60,40 @@ module.exports = utils = {
...
+
+    for pattern in predefinedPatterns
+
+      { predict, labelFn } = pattern
+
+      unless labelFn
+        labelFn = do (predict) -> (hostname) ->
+          hostname.replace(new RegExp("\.?#{predict}$"), '').split('.').pop()
+
+      if hostname.match new RegExp "#{predict}$"
```
gokmen: you probably know and not a big deal but you can make string interpolation in regular expressions with 

```coffee
///#{predict}$///
```

gokmen:
@szkl should we take this in or wait until next deploy? your call it LGTM 👍

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -60,31 +60,40 @@ module.exports = utils = {
...
+      { predict: "(dev|sandbox|latest|prod)\.#{basePattern}" }
+
+      # <team-domain>.koding.com
+      # <team-domain>.<username>.koding.team
+      { predict: "#{basePattern}" }
+
+      # <team-domain>.<IPv4>.xip.io
+      # IPV4: A.B.C.D
+      {
+        predict: 'xip\.io'
```
usirin: Can we somehow generalize this to work with `xip.io` like service?

szkl:
Let's wait for the next cycle. No need to rush it.

client/landing/site.landing/coffee/core/utils.coffee
szkl: I can't think of any practical way to do that. It is much easier to add a new rule for a target service. Some expect host address to be constructed with dashes instead of dots etc.

szkl:
Thanks for the review.

PR #9324, opened 2016-10-18
provider/google: add ssh public keys per each new instance instead of one global value per project

ppknap:
This PR changes google provider behavior. Instead of creating one sshKeys metadata per project google provider will inject ssh-keys directly to created instance.
## Description

This change will prevent errors that occur when sshKeys metadata key already exists in the project.

Additionally, instances that define `block-project-ssh-keys` matadata property will receive correct ssh keys.
## Motivation and Context

Fix stack bootstrapping error:

```
* google_compute_project_metadata.default: Error, key 'sshKeys' already exists in project 'kodingdev-vms'
```
## How Has This Been Tested?

Manually.
Unit tests.
## Types of changes


rjeczalik (approved):

cihangir (approved):
<3

Issue #9251, opened 2016-10-11
KD: Stack/Vagrant installations


hakankaradis:
Once klient or kd has a method to work on all Unix operating systems (#9250), Stacks (Vagrant and Kloud VMs) should use that method to install Klient.

The main blocker preventing Stacks from working on all Unix OSs is that Klient (on Stacks) is installed with a .deb file. If we instead install klient with a method that supports all Unix OSs, koding will support all unix OS stacks.

/cc: @rjeczalik

rjeczalik:
```
2016-10-20 12:04:42 [kd] INFO     Installation created log file at "/var/log/kd.log"
2016-10-20 12:04:47 [kd] ERROR    error registering klient: exit status 1, klient stderr: remove /etc/hosts: device or resource busy
2016-10-20 12:04:48 [kd] ERROR    ExitErrAction encountered error. err:error registering klient: exit status 1, klient stderr: remove /etc/hosts: device or resource busy
```

klient upon start adds a static route for dev.kodi.ng, it does that by writing the /etc/hosts file "atomically" - creates a temporary file in /etc, writes the new content, on success deletes /etc/hosts and move the tmp one there. It's not possible to rm /etc/hosts on all distros, thus the atomic write should be dropped and replaced with direct edit of the /etc/hosts file.

Issue #8965, opened 2016-09-07
klient/kd: configuration v2


rjeczalik:
## Description

Implement configuration file support for Klient. By issue we should:
- move /etc/kite/kite.key to the configuration file (`~/.config/koding/klient.toml`)
- move variables from `/Library//LaunchDaemons/klient.plist` to the configuration file
- move variables from `/opt/kite/klient/klient.sh` to the configuration file
- remove `/opt/kite/klient/klient.sh` and use the klient binary directly
- add generating configuration file to kontrol and update `klient -register` implementation to fetch configuration file instead of only kite.key
- replace every sed-hack in kloud/userdata and install*.sh scripts with writing the configuration file instead
## Discussion

Continuation of [the discussion here](https://github.com/koding/koding/issues/8935#issuecomment-244325107).

The aim of this task is to refactor hardcoded klient configuration in order to help solving https://github.com/koding/koding/issues/8920. Refactoring means to move any compile-time configuration like endpoint URLs and alike to runtime configuration. The runtime configuration ideally should be a separate configuration file, e.g. ~/config/koding/klient.cfg. Additionally, but not essential, is to have kite endpoint with crud ops on configuration, that'd allow kd or kloudctl to configure klient remotely without a need to restart it (to preserve mounts etc.).

Currently klient contains default configuration for resources like [kontrol URL](https://github.com/koding/koding/blob/427d5e74886d8ac550327b66cfb85a63ce6f5c31/go/src/koding/klientctl/config/config.go#L83), [tunnelserver URL](https://github.com/koding/koding/blob/427d5e74886d8ac550327b66cfb85a63ce6f5c31/go/src/koding/klientctl/config/config.go#L88) or [klient URL](https://github.com/koding/koding/blob/427d5e74886d8ac550327b66cfb85a63ce6f5c31/go/src/koding/klientctl/config/config.go#L24).

The hardcoded URLs are then [replaced during compile-time via linker flags](https://github.com/koding/koding/blob/427d5e74886d8ac550327b66cfb85a63ce6f5c31/go/src/koding/klientctl/build.sh#L37).

For environments other than production, the defaults are overwritten by exporting environmental variables like:

```
export KONTROLURL=http://54.82.228.61:8090/kontrol/kite; curl -sL https://kodi.ng/c/p/kd | bash -s ead1daa4
```

[...]

PR #9328, opened 2016-10-19
klient: add -metadata and -metadata-file flags

rjeczalik:
Currently starting klient requires kite.key written in /etc/kite, script file in /opt/kite/klient and os-specific service installed (more on this [here](https://github.com/koding/koding/issues/8965) and [here](https://github.com/koding/koding/issues/9251)).

In order to be able to start klient within a container, it needs to be possible to configure it and run in a preferably single command, without above-mentioned dependencies. As it might be not possible to write to /etc/kite or /opt/kite/klient in a container's filesystem.

This PR integrates klient with `kd config` and adds support for configuring klient / kd by importing "Koding metadata" via -metadata or -metadata-file flags.

The Koding metadata is a simple json that contains configuration (marshalled `(koding/kites/config).Konfig` struct) that is applied to klient / kd upon start.

This feature will be also used for installing klient by [this](https://github.com/koding/koding/issues/9251) issue.


go/src/koding/kites/config/konfig.go
```diff
@@ -32,24 +23,66 @@ var KonfigCache = &CacheOptions{
...
 	// Klient / KD auto-update endpoints.
 	KlientLatestURL string `json:"klientLatestURL,omitempty"`
 	KDLatestURL     string `json:"kdLatestURL,omitempty"`
 
-	Debug bool `json:"debug,omitempty"`
+	// Public S3 bucket for writing logs.
+	PublicBucketName   string `json:"publicBucketName,omitempty"`
+	PublicBucketRegion string `json:"publicBucketRegion,omitempty"`
+
+	Debug int `json:"debug,omitempty"`
```
cihangir: we use bool for debug

go/src/koding/klient/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
cihangir: 👍

go/src/koding/klientctl/main.go
```diff
@@ -59,6 +59,8 @@ func main() {
 	// always set to a number of available cores.
 	runtime.GOMAXPROCS(runtime.NumCPU())
 
+	debug = debug || config.Konfig.IsDebug()
+
 	// The writer used for the logging output. Either a file, or /dev/null
 	var logWriter io.Writer
```
cihangir: @rjeczalik do you know why we handle file writing manually instead of piping them on klient command with `> file 2>&1` ?

go/src/koding/kites/config/konfig.go
cihangir: also i realized that there are other occurrences of using integer for debug statement, could you fix them too?

go/src/koding/kites/config/cache.go
```diff
@@ -34,19 +35,39 @@ type Cache struct {
...
-	return nil
+	return &Cache{
+		EncodingStorage: &storage.EncodingStorage{
+			Interface: bolt,
+		},
+	}, nil
+}
+
+func newBoltDB(o *CacheOptions) (*bolt.DB, error) {
+	os.MkdirAll(filepath.Dir(o.File), 0755)
```
rjeczalik: I think there's no need to duplicate `os` package file handling. `os.MkdirAll` is a best-effort attempt to create base directory, if it's missing. We ignore the error, since `bolt.Open` is going to report the very same error that os.MkdirAll would:

- if there are no permissions to create dir with `os.MkdirAll`, `bolt.Open` will fail with `2016/10/20 15:33:01 open abc/abc.bolt: permission denied`
- if os.MkdirAll failed for whatever other reason, `bolt.Open` will just fail with `2016/10/20 15:32:38 open abc/abc/abc.bolt: no such file or directory`, which is pretty self-describing upon inspection.

If that's not enough, how would you handle it differently?

cihangir: stating "minor" again, i just mentioned the error message, `no such file or directory` message does not say anything about why that file/folder is required or anything

cihangir (approved):

PR #9331, opened 2016-10-19
Add support for image families in google provider.

ppknap:
This PR enables support for image families in google provider.
## Description

When google provider recognize image family disk, it will use Google API to find the latest image that is part of an image family and is not deprecated. Provided solution should work for all known image families.
## Motivation and Context

Koding's version of terraform doesn't support image families. However, the newest one does. So, these changes should be reverted as soon as Koding's terraform package is updated.
## How Has This Been Tested?

Manually.
Unit tests.
## Types of changes

### DIFF
--- a/go/src/koding/kites/kloud/provider/google/imgfamily_test.go
+++ b/go/src/koding/kites/kloud/provider/google/imgfamily_test.go
@@ -1,0 +1,158 @@
+package google
+
+import (
+	"reflect"
+	"testing"
+)
+
+var getFromFamilyFixture = map[string]string{
+	"ubuntu-os-cloud/ubuntu-1404-lts":            "ubuntu-1404-trusty-v20161010",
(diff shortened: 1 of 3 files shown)


go/src/koding/kites/kloud/provider/google/imgfamily_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"rhel-cloud/rhel-7":                          "rhel-7-v20160921",
+	"coreos-cloud/coreos-stable":                 "coreos-stable-1122-2-0-v20160906",
+}
+
+var getFromFamilyTestFunc = func(project, family string) string {
+	return getFromFamilyFixture[project+"/"+family]
+}
+
+func TestFamily2Image(t *testing.T) {
+	tests := []struct {
```
rjeczalik: nitpick: use `map[string]struct{ ...` instead of `[]struct{ Name string; ...`, you get test-case randomisation for free

gokmen (approved):

PR #9363, opened 2016-10-21
kloud: explicitely log authenticate failures

rjeczalik:
Fix for https://github.com/koding/koding/pull/9357, authenticate errors needs to be logged explicitely, since the API does not return errors in the handler (probably we may want to change it in future, for now we always return AuthResponse which contains eventual error strings).


go/src/koding/kites/kloud/stack/provider/authenticate.go
```diff
@@ -1,12 +1,16 @@
...
 import (
+	"strings"
+
 	"koding/db/mongodb/modelhelper"
 	"koding/kites/kloud/stack"
 
 	"golang.org/x/net/context"
 )
 
+var removeNL = strings.NewReplacer("\n", " ", "\t", "")
```
rjeczalik: Done.

szkl (approved):

PR #9365, opened 2016-10-21
socialapi: remove admin requirement

cihangir:


cihangir:
cc/ @usirin 

First create the customer (call create customer endpoint)
Then create the subscription (call subscribe endpoint)

cihangir:
pingen sie

rjeczalik (approved):

PR #9376, opened 2016-10-24
kloud/userdata: write Koding metadata

rjeczalik:
This PR improves klient metadata support introduced by #9328, brings back custom verifyAudience func for klient (as long as we have managed channels we need to keep it around unfortunately) and makes kloud dump metadata during cloud-init provisioning.

### DIFF
--- a/go/src/koding/kites/keygen/s3.go
+++ b/go/src/koding/kites/keygen/s3.go
@@ -121,22 +121,11 @@ func (ub *UserBucket) userPut(path string, rs io.ReadSeeker) (*url.URL, error) {
 		return nil, err
 	}
 
-	u := &url.URL{
+	return &url.URL{
 		Scheme: "https",
 		Path:   "/" + path,
-	}
-
-	// S3 bucket endpoints, for more details see:
-	//
-	//   http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html
-	//
-	if ub.cfg.Region == "us-east-1" {
-		u.Host = ub.cfg.Bucket + ".s3.amazonaws.com"
-	} else {
-		u.Host = ub.cfg.Bucket + ".s3-" + ub.cfg.Region + ".amazonaws.com"
-	}
-
-	return u, nil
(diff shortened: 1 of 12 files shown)


cihangir (approved):

ppknap (approved):

rjeczalik:
ping

go/src/koding/kites/keygen/s3.go
```diff
@@ -121,22 +121,11 @@ func (ub *UserBucket) userPut(path string, rs io.ReadSeeker) (*url.URL, error) {
...
-	//   http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html
-	//
-	if ub.cfg.Region == "us-east-1" {
-		u.Host = ub.cfg.Bucket + ".s3.amazonaws.com"
-	} else {
-		u.Host = ub.cfg.Bucket + ".s3-" + ub.cfg.Region + ".amazonaws.com"
-	}
-
-	return u, nil
+		Host:   ub.cfg.Bucket + ".s3.amazonaws.com",
```
ppknap: as far as I remember URLs like this (where they defaults to us-east-1 region) are not recommended. They also will not work for other regions (301 without Location header)

go/src/koding/klient/app/klient.go
```diff
@@ -847,3 +874,81 @@ func userIn(user string, users ...string) bool {
...
+
+	if kite.Username != aud.Username {
+		return fmt.Errorf("audience: username %q not allowed (%s)", aud.Username, audience)
+	}
+
+	// Verify environment - managed environment means production klient
+	// running on a user's laptop; devmanaged is for development/sandobx
+	// environments.
+	//
+	// TODO(rjeczalik): klient should always have development/production
```
ppknap: :+1:

ppknap (approved):

go/src/koding/kites/keygen/s3.go
rjeczalik: `s3.amazonaws.com/bucket` won't work if the region is different than the default one; however `bucket.s3.amazoaws.com` will work.

PR #9380, opened 2016-10-24
provider/google: create separate virtual network for each credential identifier.

ppknap:
Since this PR, a new virtual network will be created for each new credential identifier.
## Description

Currently, google provider uses `default` network which may be already used by other
koding accounts. This PR forces bootstrap step to produce separate network for each credential identifier.
## Motivation and Context
- Consistency with AWS and Azure providers.
- Allow to use one project for different koding accounts.
- Easy to clean-up (user just needs to delete created network - all other resources will be cleaned by google).
- Bootstrap will not pollute `default` network which may already have custom configuration. 
## How Has This Been Tested?

Manually
## Types of changes
## Checklist:

/cc @gokmen


rjeczalik:
> all other resources will be cleaned by google

Is this somehow guaranteed by gcp? Last time I tried it left a number of forwarding rules and I needed to clean them up manually.

go/src/koding/kites/kloud/provider/google/google_test.go
```diff
@@ -140,6 +140,13 @@ func TestValidators(t *testing.T) {
 			},
 			IsValid: false,
 		},
+		{
+			// 12 //
+			Validator: &Bootstrap{
+				KodingNetworkID: ``,
```
rjeczalik: Why not just `""`, for consistency's sake?

rjeczalik (approved):
👍

ppknap:
> Is this somehow guaranteed by gcp? Last time I tried it left a number of forwarding rules and I needed to clean them up manually.

image

cihangir (approved):

PR #9402, opened 2016-10-26
provider/google: machine cannot stop when its disk size metadata is set to zero

ppknap:
This PR fixes the problem with google instance not stopping after requested to do so from Koding UI.
## Description

We check if metadata is valid before stopping the machine and we treated 0 disk size as an error. Thus, it was not possible to stop machine when disc size was not explicitly set in stack template.
## Motivation and Context

Fix reported bug.
## How Has This Been Tested?

Manually.
## Types of changes

/cc @gokmen @rjeczalik 

@gokmen In described case, kloud kite returns an error to kite's caller (not valid metadata). Probably, you should check if you receive this error on your end.

### DIFF
--- a/go/src/koding/kites/kloud/provider/google/google.go
+++ b/go/src/koding/kites/kloud/provider/google/google.go
@@ -160,9 +160,6 @@ func (m *Meta) Valid() error {
 	if m.Image == "" {
 		return errors.New(`metadata value for "image" is empty`)
 	}
-	if m.StorageSize == 0 {
-		return errors.New(`metadata value for "storage_size" is 0`)
-	}
 	if m.MachineType == "" {
 		return errors.New(`metadata value for "machie_type" is empty`)
 	}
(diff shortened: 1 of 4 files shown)


rjeczalik (approved):
👍

rjeczalik:
@gokmen we may want to include this one into the next deployment

ppknap:
@rjeczalik I removed `waiting for tests` label since this logic is covered by unit tests and the actual bug would be only reproducible during integration tests.

rjeczalik:
@ppknap It's more about procedure than actual coverage - yes, ideally every kloud bug fix should be covered with a test-case, but as long as I am still wip on adding integration infra for kloud - the build (and tests) should be green before we merge anything to master, regardless the actual coverage.

gokmen:
@ppknap we can merge this in but once I got this error I manually set the disk size to `10` in db and it started to `stop` process even though it couldn't finish it in time and I got `Timeout Error` as expected.

ppknap:
@gokmen yes, if you manually set disc size metadata, the machine is called to stop and it will stop. However, google api reports machine state as `terminated` instead of `stopped`. And, since we are waiting for stopped state, this was also going to time out. This PR also address this: https://github.com/koding/koding/pull/9402/files#diff-7118b476d49e68771c8b32d3bad98510R137

PR #9401, opened 2016-10-26
klient: self-installation as a system service

rjeczalik:
This PR enables klient to self-install itself as a system service.

Required for https://github.com/koding/koding/issues/9071.
~~Depends on https://github.com/koding/koding/pull/9400~~.

PR #9414, opened 2016-10-28
kloud: add Marathon provider

rjeczalik:
This PR adds support for Marathon kloud provider. It uses Terraform representation of applications provided by [terraform-provider-marathon](https://github.com/Banno/terraform-provider-marathon).

There's a number of limitations introduced by transparently injecting klient into each container created by stack:

- multiple-instance application is converted to application group, due to [lack of access to INSTANCE_ID](https://github.com/mesosphere/marathon/issues/1242), so it's impossible to insert distinct kite.key (it's a temporary "limitation", if the Marathon issue gets resolved, the provider will be reworked to support multiple instances)

- the `args` argument is not supported due to the entrypoint-wrapping kloud does, `cmd` argument should be use instead

- it's not possible to use custom entrypoint for containers, as kloud already uses one

This PR contains complete support for creating and augmenting Marathon stacks. The `start` / `stop` methods are a nop always giving success. In another PR I'm going to send a bit smarter implementation (it's not possible to stop a single container, I plan to start/stop whole stack when a single stop/start  is called - /cc @gokmen).

Deps: ~~[#9409](https://github.com/koding/koding/pull/9409)~~ ~~[#9401](https://github.com/koding/koding/pull/9401)~~


go/src/koding/kites/kloud/provider/marathon/machine.go
```diff
@@ -0,0 +1,52 @@
...
+//
+// It does not support stopping a single container within an app,
+// so stopping whichever container from the app stops the whole app.
+func (m *Machine) Stop(ctx context.Context) (interface{}, error) {
+	return nil, nil
+}
+
+// Info gives state of the app.
+func (m *Machine) Info(context.Context) (machinestate.State, interface{}, error) {
+	return machinestate.Running, nil, nil
```
cihangir: so are we sure if there is an instance of machine, it is running?

go/src/koding/kites/kloud/provider/marathon/schema.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// Credential represents credential information
+// that are required to deploy a Marathon app.
+type Credential struct {
+	URL                string `json:"url" bson:"url" hcl:"url"`
+	BasicAuthUser      string `json:"basic_auth_user" bson:"basic_auth_user" hcl:"basic_auth_user"`
+	BasicAuthPassowrd  string `json:"basic_auth_password" bson:"basic_auth_password" hcl:"basic_auth_password"`
+	RequestTimeout     int    `json:"request_timeout" bson:"request_timeout" hcl:"request_timeout"`
+	DeploymentTimepout int    `json:"deployment_timeout" bson:"deployment_timeout" hcl:"deployment_timeout"`
```
cihangir: for json and bson we follow camelCase notation

go/src/koding/kites/kloud/provider/marathon/stack.go
```diff
@@ -0,0 +1,406 @@
...
+	content, err := t.JsonOutput()
+	if err != nil {
+		return nil, err
+	}
+
+	return &stack.Template{
+		Content: content,
+	}, nil
+
+	return nil, nil
```
cihangir: unreachable code

go/src/koding/kites/kloud/provider/marathon/stack.go
```diff
@@ -0,0 +1,406 @@
...
+
+	t.Resource["marathon_app"] = resource.MarathonApp
+
+	err := t.ShadowVariables("FORBIDDEN", "marathon_basic_auth_user", "marathon_basic_auth_password")
+	if err != nil {
+		return nil, errors.New("marathon: error shadowing: " + err.Error())
+	}
+
+	if err := t.Flush(); err != nil {
+		return nil, errors.New("marathom: error flushing template: " + err.Error())
```
cihangir: s/marathom/marathon

go/src/koding/kites/kloud/provider/marathon/stack.go
```diff
@@ -0,0 +1,406 @@
...
+//
+// Since Marathon does not support instance indexing it's not possible
+// to assign unique metadata for each of the instace, thus making such
+// stack unusable for Koding. Relevant issue:
+//
+//   https://github.com/mesosphere/marathon/issues/1242
+//
+// What we do instead is we convert multiple instances of an application to
+// an application group as a workaround.
+func (s *Stack) convertInstancesToGroup(app map[string]interface{}) (originalAppID string) {
```
cihangir: test please

go/src/koding/kites/kloud/provider/marathon/stack.go
```diff
@@ -0,0 +1,406 @@
...
+// klient before running container's command.
+//
+// The entrypoint is injected in a twofold manner:
+//
+//   - if "cmd" argument is used, it's prefixed with an entrypoint.N.sh script
+//   - if container's default command (the one from Dockerfile) is used,
+//     the container's entrypoint (by default /bin/sh) is replaced
+//     with the Koding one
+//
+func (s *Stack) injectEntrypoint(app map[string]interface{}, originalAppID string) (labels []string, err error) {
```
cihangir: test here too

go/src/koding/kites/kloud/provider/marathon/schema.go
rjeczalik: We can't, here and in any other provider - the names come from Terraform attributes.

go/src/koding/kites/kloud/provider/marathon/stack.go
rjeczalik: There is already a test for this - https://github.com/koding/koding/blob/779aed8d20ba9c1a0f55116e24b0a8d90595d147/go/src/koding/kites/kloud/provider/marathon/stack_test.go#L43.

go/src/koding/kites/kloud/provider/marathon/stack.go
rjeczalik: Also there's test for this - every test case in TestApplyTemplate tests injecting entrypoints.

go/src/koding/kites/kloud/provider/marathon/stack.go
```diff
@@ -0,0 +1,406 @@
...
+			"uri":        fmt.Sprintf("%s/entrypoint.%d.sh", s.EntrypointBaseURL, i+1),
+			"executable": true,
+		})
+	}
+
+	app["fetch"] = fetch
+}
+
+func (s *Stack) injectHealthChecks(app map[string]interface{}) {
+	healthCheckGroup, ok := app["health_checks"].(map[string]interface{})
```
cihangir: test here as well

go/src/koding/kites/kloud/provider/marathon/stack.go
rjeczalik: E.g.

- command entrypoint: [before](https://github.com/koding/koding/blob/779aed8d20ba9c1a0f55116e24b0a8d90595d147/go/src/koding/kites/kloud/provider/marathon/testdata/single-app.json#L6) and [after](https://github.com/koding/koding/blob/779aed8d20ba9c1a0f55116e24b0a8d90595d147/go/src/koding/kites/kloud/provider/marathon/testdata/single-app.json.golden#L6)

- container entrypoint: [before](https://github.com/koding/koding/blob/779aed8d20ba9c1a0f55116e24b0a8d90595d147/go/src/koding/kites/kloud/provider/marathon/testdata/multi-container.json#L7) and [after](https://github.com/koding/koding/blob/779aed8d20ba9c1a0f55116e24b0a8d90595d147/go/src/koding/kites/kloud/provider/marathon/testdata/multi-container.json.golden#L12)

go/src/koding/kites/kloud/provider/marathon/stack.go
rjeczalik: See above, also tested.

go/src/koding/kites/kloud/provider/marathon/machine.go
rjeczalik: From PR description:

```
The start / stop methods are a nop always giving success. In another PR I'm going to send a bit smarter implementation (it's not possible to stop a single container, I plan to start/stop whole stack when a single stop/start is called 
```

There's no concept of stopping / starting vms in marathon, since all of the containers are part of a single app.

I'm going to improve this in next PR, so `stop` stops whole application (stack), and `start` starts it.

go/src/koding/klient/deploy-entrypoint.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
cihangir: where are we going to execute this script? in our ec2 instances?

rjeczalik: You dev instance - if you decide we need less/more entrypoints, you'd run it from your dev instance provided you have creds for prod S3 configured.

I've created it to not manually upload script files, as I'm lazy. Most probably no-one will never run it again. Adding just in case.

go/src/koding/kites/kloud/provider/marathon/machine.go
cihangir: Ok, i see it also in the comments of start and stop, but my question is, in the context of the app, did you encapsulate the instance creation in a way that if somehow i have an instance of a machine, does that mean it is running? if not, this function returns invalid info.

go/src/koding/kites/kloud/provider/marathon/schema.go
cihangir: ok got it for json since we decode/encode them multiple times, but it should not be the case for bson, it is only valid for mongodb. Anyhow for the sake of consistency with other providers, we can leave it as is.

rjeczalik: > Anyhow for the sake of consistency with other providers, we can leave it as is.

Yes, the `meta`  object differ in notation starting from solo times. `jMachine.meta` always had fields that come from AWS (`storage_size`, `instance_type` etc.), the jCredentialDatas.meta always has the fields with underscores (`access_key`, `secret_key` etc.). So adding new providers I just followed what we had. Also it would be possible in future to get rid of 3 tags each struct has currently (`json`, `bson` and `hcl`) without additional mapping (snake case -> camel case).

go/src/koding/kites/kloud/provider/marathon/machine.go
rjeczalik: If you created successfully an application, you have guarantee all the containers with klient are always running (there's a healthcheck injected for klient). If klient is not running (unhealthy), then mesos will reschedule the app. My thinking was we should not duplicate the logic here.

I was thinking whether to add a call to mesos to check for the app health, but if the app was unhealthy, then mesos would be already deploying it, but this Info method would return to client the unhealthy status - the user would then want to start the app (deploy/resome on mesos), which would fail since mesos is already deploying the app.

go/src/koding/kites/kloud/provider/marathon/stack.go
cihangir: i see them as integration tests that is why i commented, if you feel they are enough to cover all the cases, if am ok too

go/src/koding/kites/kloud/provider/marathon/machine.go
cihangir: your app definition and mines are different 😄  i meant as app is `kloud/provider/marathon/machine.go` the code itself

go/src/koding/klient/deploy-entrypoint.sh
cihangir: 👍

go/src/koding/kites/kloud/provider/marathon/stack.go
rjeczalik: Functional tests, yes. The integration tests would be added with kd. Yes, these tests have pretty much good coverage.

There are a lots of TODOs in this provider - e.g. if Mesos add INSTANCE_ID we will get rid of converting instances to groups. I chose not to add unittests that test against non-exported API in order to not have to maintain tests each time this private API changes (e.g. convert* is removed). And sometimes it's impossible to refactor existing code with having tests passing, since a lot of unittests in our codebase call unexported methods directly.

go/src/koding/kites/kloud/provider/marathon/machine.go
rjeczalik: Oh, got it. Yes, it's always running.

cihangir (approved):

ppknap (approved):

PR #9416, opened 2016-10-28
provider/google: autodiscovery for image disks sizes

ppknap:
This PR makes possible to obtain disk size during `terraform plan` phase. With this information we are able to correctly set created instance's disk size metadata.
## Description

During machine creation we look up for image size using GCP API and if we find one, we will add that information to terraform template. Then, terraform plan/apply will correctly report disk size of the instance which is going to be created.
## Motivation and Context

Fix missing disk size metadata in Google Provider.
## How Has This Been Tested?

Manually,
Unit tests.

image
## Types of changes

### DIFF
--- a/go/src/koding/kites/kloud/provider/google/disksize.go
+++ b/go/src/koding/kites/kloud/provider/google/disksize.go
@@ -1,0 +1,118 @@
+package google
+
+import "strings"
+
+// Image2Size is used to find and map image disk sizes when they are not
+// provided. This structure was created in order to discover and attach image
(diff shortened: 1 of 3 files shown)


rjeczalik (approved):
Looks good to me, but to be honest if that amount of trickery is needed to get a disk size (>300LOC) maybe we could do something better at the core?

ppknap:
Half of these lines are tests. But I'm not going to count LOC anymore

gokmen (approved):

PR #9423, opened 2016-10-30
provider/google: always report terminated state as stopped

ppknap:
Google reports stopped machines as terminated. This PR adds full support for that.
## Motivation and Context

UI calls machine.Info directly so its response needs to report proper machine status.
## Types of changes


gokmen (approved):
👍

rjeczalik (approved):

PR #9424, opened 2016-10-30
provider/azure: improvements

rjeczalik:
Fixes https://github.com/koding/koding/issues/9266#issuecomment-257156675.


go/src/koding/kites/kloud/provider/azure/stack.go
```diff
@@ -161,6 +173,10 @@ func (s *Stack) ApplyTemplate(c *stack.Credential) (*stack.Template, error) {
 	}
 
 	for name, vm := range res.AzureInstance {
+		if n, ok := vm["name"].(string); !ok || n == "" {
+			vm["name"] = name + "-" + s.id()
```
cihangir: if id() returns empty string vm[name] will look odd with an ending dash eg: vn["name"] = "koding-"

go/src/koding/kites/kloud/stack/provider/bootstrap.go
```diff
@@ -85,7 +85,8 @@ func (bs *BaseStack) bootstrap(arg *stack.BootstrapRequest) (interface{}, error)
 				}
 			} else {
 				bs.Log.Info("Creating bootstrap resources belonging to identifier '%s'", cred.Identifier)
-				bs.Log.Debug("Bootstrap template: %s", tmpl.Content)
+				bs.Log.Debug("Bootstrap template:")
```
cihangir: not an issue since these are debug statements, just to note:
it is usually better to log statements in one line than multiple ones, when the logging volume is so high it will be harder to keep track of them because others will interfere between these two

cihangir (approved):

go/src/koding/kites/kloud/stack/provider/bootstrap.go
rjeczalik: Done.

go/src/koding/kites/kloud/provider/azure/stack.go
```diff
@@ -99,6 +101,14 @@ func (s *Stack) VerifyCredential(c *stack.Credential) error {
 	return nil
 }
 
+func trim(s string, n int) string {
```
ppknap: No need for function here:
```go
string(bytes.NewBufferString(s).Next(n))
```

go/src/koding/kites/kloud/provider/azure/stack.go
```diff
@@ -316,6 +333,17 @@ func (s *Stack) injectCloudInit(vm map[string]interface{}, name, kiteKeyName str
 	return countKeys, nil
 }
 
+func (s *Stack) id() string {
```
ppknap: Could you make this function part of BaseStack? I would be useful in other providers.

rjeczalik: Do you need it now? I think it would be better to just copy it.

It's good enough for Azure, but I don't find it a good API for generic purpose - the `(*BaseStack).ID()` does not say a thing, I'd rather add something like `func GenerateUniqueName(scope Scope) string` where Scope can be `StackScope`, `CredentialScope` etc. so one does not need to wonder if strings returned by `ID()` would conflict on reinit or not.

ppknap (approved):

go/src/koding/kites/kloud/provider/azure/stack.go
rjeczalik: Creating a *bytes.Buffer just to get a slice of a string? I think it's not needed.

ppknap: Creating a separate function just to get a slice of a string? Indeed, there is an extra allocation here. But seriously, this is nothing comparing to other operations/allocations made during stack provisioning.

rjeczalik: It's not a single allocation, it's three allocations - it allocates `*bytes.Buffer`, `[]byte` from string and string from `[]byte`, unnecessarily.

We'd need to live with the function, I can eventually inline it. We need to make the code better allocation-wise, not worse, even though it may be not perfect now.

rjeczalik: Done, changed to `substringN` - `trimN` could confuse it trims N last characters, not all the characters above N.

rjeczalik:
ping

gokmen:
@rjeczalik will be merged after deploy

PR #9427, opened 2016-10-31
Set version option to `master`

szkl:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,13 @@ MAINTAINER Sonmez Kartal <sonmez@koding.com>
 ADD . /opt/koding
 WORKDIR /opt/koding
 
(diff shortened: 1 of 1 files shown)


cihangir:
would you mind explaining why we need this?

szkl:
configure script fallbacks to default 1.0 as version which is misrepresented (as 1) in some places. git revision is not available because .git directory is not available in the context. I've thought setting a predefined version would prevent mishandling of version option.

cihangir:
got it. this wont break the build command right? https://github.com/koding/koding-gitlab/blob/master/koding/deployment.yaml#L23-L24

szkl:
Nope. It will overwrite version as it is right now 👍

cihangir (approved):

PR #9431, opened 2016-10-31
Set environment variable only if given name is unset

szkl:
This will improve handling of variables set to an empty value.

### DIFF
--- a/config/generateShellEnv.coffee
+++ b/config/generateShellEnv.coffee
@@ -4,7 +4,7 @@ module.exports.create = (KONFIG, options = {}) ->
   env = ''
 
   add = (name, value) ->
-    env += "export #{name}=${#{name}:-#{value}}\n"
+    env += "declare -p #{name} &> /dev/null || export #{name}=\"#{value}\"\n"
 
   traverse.paths(KONFIG).forEach (path) ->
     node = traverse.get KONFIG, path


config/generateShellEnv.coffee
```diff
@@ -4,7 +4,7 @@ module.exports.create = (KONFIG, options = {}) ->
   env = ''
 
   add = (name, value) ->
-    env += "export #{name}=${#{name}:-#{value}}\n"
+    env += "declare -p #{name} &> /dev/null || export #{name}=\"#{value}\"\n"
```
rjeczalik: ✨ ✨ ✨

cihangir (commented):
where do we need to know if a value is not set specifically?

probably we dont need -p if we are piping to dev/null

szkl:
Empty value is distinctive for some configurations (e.g. sneakerS3). `master` version sets default value if a variable is empty which prevents setting an empty value as an override. This led me to set environment variables to default values only if names are unset.

`-p` option makes `declare` command print variable declaration instead of declaring a new name. This change only interested in exit code of `declare -p NAME` to determine whether it is set or unset. Redirection is to hide error messages like `bash: declare: KONFIG_PROJECTROOT: not found` to avoid any confusion during environment is initialized.

cihangir (approved):

PR #9439, opened 2016-10-31
stack/provider: log identifier for a failed authentication

rjeczalik:


gokmen:
@rjeczalik we'll make deploy from latest and it's better to make the current deploy first, then we can make another one with these changes.

gokmen (approved):

cihangir (approved):
you could utilize context feature of logging package

Issue #8164, opened 2016-06-21
Azure Support


tevirgen:
**Goals**

Add support for azure and azurerm terraform providers to terraformer and kloud.

**Why**

Required support for Azure Cloud in stack templates.

**Acceptance Criteria**
- it is possible to build an Azure stack.

gokmen:
@rjeczalik getting following error while building with azure provider;

```
o:azure_instance.azure-instance: Still creating... (10s elapsed)
e:Error applying plan:

1 error(s) occurred:

* azure_instance.azure-instance: You must supply a 'password' and/or a 'ssh_key_thumbprint' when using a Linux image

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
 (c7f9aafc71a79149)
```

here is my stack template;

```
provider:
  azure:
    publish_settings: '${var.azure_publish_settings}'
    subscription_id: '${var.azure_subscription_id}'

resource:
  azure_instance:
    # this is the name of your VM
    azure-instance:
      # select your instance size here: eg. Basic_A1
      size: Basic_A1
      # base image for your instance
      image: 'Ubuntu Server 14.04 LTS'
      # on custom_data section we will write bash and configure our VM
      custom_data: |-
        # let's create a file on your root folder:
        echo "hello world!" >> /helloworld.txt
```

you can try with #9319

rjeczalik:
@gokmen The template contains not enough arguments for Azure - it requires either a password or ssh keypair. I can add in kloud a dummy key pair if no is specified, but it won't be usable.

gokmen:
@rjeczalik can u provide an example template then? I used the one you were using in azure provider tests.

rjeczalik:
Done.

PR #9453, opened 2016-11-02
provider/azure: add ssh thumbprint / password fields to credentials

rjeczalik:
Addresses https://github.com/koding/koding/issues/8164#issuecomment-257718400.

### DIFF
--- a/go/src/koding/kites/kloud/contexthelper/publickeys/publickeys.go
+++ b/go/src/koding/kites/kloud/contexthelper/publickeys/publickeys.go
@@ -1,6 +1,12 @@
 package publickeys
 
-import "golang.org/x/net/context"
+import (
+	"crypto/sha256"
+	"encoding/base64"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/net/context"
+)
 
 type key int
 
@@ -15,6 +21,17 @@ type Keys struct {
 	PublicKey  string
 	PrivateKey string
 	KeyName    string
+}
+
+func (k *Keys) Thumbprint() (string, error) {
+	key, err := ssh.ParsePrivateKey([]byte(k.PrivateKey))
+	if err != nil {
+		return "", err
+	}
+
+	hash := sha256.Sum256(key.PublicKey().Marshal())
+
+	return base64.RawStdEncoding.EncodeToString(hash[:]), nil
 }
 
 func FromContext(ctx context.Context) (*Keys, bool) {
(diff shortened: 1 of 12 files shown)


gokmen (approved):

gokmen:
@rjeczalik tests were failing https://app.wercker.com/buildstep/581a1a7597b3e2010095fac5 I've restarted a new one, dunno if it's related with this PR.

go/src/koding/kites/kloud/contexthelper/publickeys/publickeys.go
```diff
@@ -17,6 +23,17 @@ type Keys struct {
 	KeyName    string
 }
 
+func (k *Keys) Thumbprint() (string, error) {
```
cihangir: fingerprint?

gokmen: I guess this is because http://security.stackexchange.com/a/71203

cihangir: Thanks for the link @gokmen TIL `thumbprint algorithm`

I was going to add another comment if this is fingerprinting hashing algo needs to be changed , because SSH Public Key File Format [RFC]( http://www.ietf.org/rfc/rfc4716.txt) _page 5 section 4_ states hashing algorithm to be MD5 not sha1 

Also one of the comments ( same guy with the author of the response ) states thumbprinting is used for root CAs where we dont have a notion of validating http://security.stackexchange.com/questions/71199/what-is-the-difference-between-a-thumbprint-algorithm-signature-algorithm-an/71203#comment208102_71203

ppknap (approved):

go/src/koding/kites/kloud/contexthelper/publickeys/publickeys.go
rjeczalik: I removed this code, it turned out it's both not valid and not usefull here for now.

> thumbprinting is used for root CAs 

And this is exactly the case for Azure, it turned out you don't simply upload a ssh public key - you need to create a cert out of your private key and upload it to Azure [0]. And then you can use the thumbprint of it as an ID. So for now we do not inject kloud's SSH key, we inject instead a random password.

Injecting random password / kloud's ssh key was only a workaround here. It's not about ssh key pair management, although it'd be good if we did it for all the providers but it's outside of the scope of this PR.

[0] https://support.microsoft.com/en-us/kb/2977336

cihangir (approved):

ppknap (approved):

PR #9461, opened 2016-11-03
kites/config: prevent rebuilding all config-dependent packages after go generate.

ppknap:
After `go generate`, golang build tool sees new modification time in `config.json.go` and rebuilds all dependent packages. This change forces `go generate` to create new `config.json.go` either when file doesn't exist or it was modified.

## Motivation and Context
Speed up go compilation.

## How Has This Been Tested?
Manually.

### DIFF
--- a/go/src/koding/kites/config/genconfig.go
+++ b/go/src/koding/kites/config/genconfig.go
@@ -1,0 +1,72 @@
+// +build ignore
+
+package main
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"flag"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/jteeuwen/go-bindata"
+)
+
+var (
+	pkg    = flag.String("pkg", "-", "")
+	input  = flag.String("i", "-", "")
+	output = flag.String("o", "-", "")
+)
+
+func main() {
+	flag.Parse()
+
+	outputTmp := "tmp_" + *output
+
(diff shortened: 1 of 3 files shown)


ppknap:
I was looking for such option in go-bindata, unfortunatelly it doesn't have it (yet). As you said, I'd rader we left this as is and update go-bindata someday or find similar tool which already has this feature.

go/src/koding/kites/config/genconfig.go
```diff
@@ -0,0 +1,72 @@
...
+		log.Fatalf("cannot create MD5 sum for %s: %v", outputTmp, err)
+	}
+
+	outMD5, err := fileMD5(*output)
+	if err != nil && !os.IsNotExist(err) {
+		log.Fatalf("cannot create MD5 sum for %s: %v", *output, err)
+	}
+
+	if tmpMD5 != outMD5 {
+		if err := os.Rename(outputTmp, *output); err != nil {
```
rjeczalik: Before you os.Rename, you need to os.Remove(*output) (yes, for Windows, sorry).

rjeczalik (approved):

go/src/koding/kites/config/genconfig.go
ppknap: > Rename renames (moves) oldpath to newpath. If newpath already exists, Rename replaces it.

https://golang.org/pkg/os/#Rename

cihangir (approved):

PR #9465, opened 2016-11-03
cleanup | models: Remove unnecessary schema properties

cihangir:


workers/social/lib/social/models/account.coffee
```diff
@@ -178,29 +178,7 @@ module.exports = class JAccount extends jraphical.Module
...
-      shareLocation         : Boolean
       socialApiId           : String
-      skillTags             : [String]
-      locationTags          : [String]
-      systemInfo            :
-        defaultToLastUsedEnvironment :
-          type              : Boolean
-          default           : yes
-      # counts                : Followable.schema.counts
-      counts                :
```
gokmen: we had some place where we were updating counts on each login/load afair.

gokmen: we may need to remove that part as well.

gokmen (approved):

workers/social/lib/social/models/account.coffee
cihangir: JGroup has them, it seems we've removed for users previously

cihangir:
PTAL

gokmen (approved):
👍

PR #9466, opened 2016-11-03
Fix Ide file and terminal icons

hakankaradis:
Fix styling in ide for editor and terminal icons.
Renaming file from editor fix
Add blue tag to active tab in all splits.

### DIFF
--- a/client/ide/lib/index.coffee
+++ b/client/ide/lib/index.coffee
@@ -258,6 +258,14 @@ class IDEAppController extends AppController
 
   setActiveTabView: (tabView) ->
 
+    activePane = tabView.getActivePane()
+
+    for view in @ideViews
+      for tab in view.holderView.tabs.subViews
+        if tab is activePane?.tabHandle
+        then activePane.tabHandle.setClass 'focus'
+        else tab.unsetClass 'focus'
+
     return  if tabView is @activeTabView
     return  if tabView.isDestroyed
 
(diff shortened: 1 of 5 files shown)


ftasdelen:
@hakankaradis Can we make the second pane active tab icon faded instead of white?

image

Can you also make the first pane selected tab file name white?

image

client/ide/lib/index.coffee
```diff
@@ -258,6 +258,14 @@ class IDEAppController extends AppController
 
   setActiveTabView: (tabView) ->
 
+    activePane = tabView.getActivePane()
+
+    for view in @ideViews
+      for tab in view.holderView.tabs.subViews
+        if tab is activePane?.tabHandle
+        then activePane.tabHandle.setClass 'focus'
+        else tab.unsetClass 'focus'
```
gokmen: we shouldn't need something like that, currently there is a style change on focus in somewhere else. I might be wrong but this looks weird to me.

gokmen: cc/ @sinan

sinan: yeah, which is `active` class.

https://github.com/koding/kd/blob/master/lib/components/tabs/tabview.coffee#L240

https://github.com/koding/kd/blob/master/lib/components/tabs/tabhandleview.coffee#L58

sinan:
@hakankaradis can you post a little gif showing you switch tabs please?

client/ide/lib/index.coffee
hakankaradis: We will need one global class name across all panes across all split views.

ftasdelen:
Looks good to me.

client/ide/lib/index.coffee
gokmen: merging this one but still looks suspicious to me :)

gokmen (approved):

PR #9477, opened 2016-11-04
provider: improve credential validation

rjeczalik:
@ppknap - is it posible to have one for google provider?

### DIFF
--- a/go/src/koding/kites/kloud/provider/aws/aws.go
+++ b/go/src/koding/kites/kloud/provider/aws/aws.go
@@ -193,19 +193,19 @@ var (
 // Valid implements the kloud.Validator interface.
 func (meta *Cred) Valid() error {
 	if meta.Region == "" {
-		return errors.New("aws meta: region is empty")
+		return errors.New("aws: region is empty")
 	}
 	if meta.AccessKey == "" {
-		return errors.New("aws meta: access key is empty")
+		return errors.New("aws: access key is empty")
 	}
 	if meta.SecretKey == "" {
-		return errors.New("aws meta: secret key is empty")
+		return errors.New("aws: secret key is empty")
 	}
 	if !reAccessKey.MatchString(meta.AccessKey) {
-		return errors.New("aws meta: access key is illformed")
+		return errors.New("aws: access key is ill-formed")
 	}
 	if !reSecretKey.MatchString(meta.SecretKey) {
-		return errors.New("aws meta: secret key is illformed")
+		return errors.New("aws: secret key is ill-formed")
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


go/src/koding/kites/kloud/provider/google/google.go
```diff
@@ -84,6 +85,12 @@ func (c *Cred) Valid() error {
 		return errors.New(`cred value for "project" is empty`)
 	}
 
+	var raw json.RawMessage
+
+	if err := json.Unmarshal([]byte(c.Credentials), &raw); err != nil {
```
rjeczalik: /cc @ppknap 

Best effort method, ensuring the Credentials are a valid JSON, should filter out attempts of inputing random text. If you want more (e.g. validating the actual schema), please commit it to the branch of this PR.

ppknap:
@rjeczalik credentials are in JSON:

```json
{
  "type": "service_account",
  "project_id": "infra-treat-103712",
  "private_key_id": "2ffd68fa2e2a840",
  "private_key": "-----BEGIN PRIVATE KEY-----private_key_here",
  "client_email": "639003701225-compute@developer.gserviceaccount.com",
  "client_id": "726069070949",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://accounts.google.com/o/oauth2/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/6303701225-compute%40developer.gserviceaccount.com"
}
```

So, you can unmarshal them to similar go struct and check if all fields are correct. Personally, I would create something like:
```go
var acckey = struct {
  Typ string `json:"type"`
  ProjectID string `json:"project_id"`
  PrivateKey string `json:"private_key"`
}{}
```

and then check if:
```go
acckey.Typ == "service_account",
acckey.ProjectID != "",
strings.HasPrefix(acckey.ProjectID, "-----BEGIN PRIVATE KEY-----"),
```

which should be sufficient for validation.

rjeczalik:
@ppknap Nice one. You can send it as separate PR, for now I want to get this PR in and then we can improve it.

go/src/koding/kites/kloud/provider/aws/aws.go
```diff
@@ -195,6 +201,12 @@ func (meta *Cred) Valid() error {
 	if meta.SecretKey == "" {
 		return errors.New("aws meta: secret key is empty")
 	}
+	if !reAccessKey.MatchString(meta.AccessKey) {
+		return errors.New("aws meta: access key is illformed")
```
ppknap: I think that this should be `ill-formed`

ppknap (approved):

ppknap:
I will send either PR or commit to this branch, depending on whether this PR is merged or not.

cihangir (approved):

PR #9479, opened 2016-11-06
Terraform GitHub stack

mehmetalisavas:
use koding's github provider instead of terraform's


go/build.sh
```diff
@@ -67,6 +67,9 @@ export COMMANDS=(
 export TERRAFORM_COMMANDS=(
 	vendor/github.com/hashicorp/terraform
 	vendor/github.com/hashicorp/terraform/builtin/bins/...
+)
+
+export TERRAFORM_CUSTOM_COMMANDS=(
```
mehmetalisavas: terraform's own github provider overrides koding's github provider. we disable terraform's github provider.  we will change koding github provider and enable back terraform one. But for a while we will use koding's github provider

go/src/vendor/github.com/koding/terraform-provider-github/resource_github_adduser_record.go
```diff
@@ -127,8 +127,10 @@ func resourceGithubAddUserCreate(d *schema.ResourceData, meta interface{}) error
 		return errors.New("team name is not defined")
 	}
 
-	member, _, err := client.Organizations.GetOrgMembership("", org)
-	if err != nil {
+	member, resp, err := client.Organizations.GetOrgMembership("", org)
```
cihangir: this change is not in original repo, please add there too.

cihangir (approved):

PR #9488, opened 2016-11-07
kloud/kd: make it possible to create a stack with kd

rjeczalik:
~deps: #10229~

This PR adds `import` kite method, which is used by kd to send kd.yml and credentials, and basing on those to:

- create a template
- create a stack
- build a stack

= import a stack.

A todo is to make it possible to use existing template instead.

### DIFF
--- a/go/src/koding/api/apitest/apitest.go
+++ b/go/src/koding/api/apitest/apitest.go
@@ -379,6 +379,10 @@ func (sh *StubHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var v interface{}
 	v, *sh = (*sh)[0], (*sh)[1:]
 
+	if v == nil {
+		v = make(map[string]interface{}) // return empty object instead
+	}
+
 	w.Header().Set("Content-Type", "application/json")
 
 	if err := json.NewEncoder(w).Encode(v); err != nil {
--- a/go/src/koding/api/apitest/apitest.go
+++ b/go/src/koding/api/apitest/apitest.go
@@ -1,3 +1,6 @@
+// Package apitest provides a number of mocks and other testing
+// utilities, handy when writing tests for clients that
+// communicate with socialworker / remote.api.
 package apitest
 
 import (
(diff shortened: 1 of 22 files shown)


rjeczalik:
deps: #10190

rjeczalik:
deps: #10219

rjeczalik:
deps: #10223

rjeczalik:
deps: #10229

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -210,6 +210,11 @@ module.exports = class JStackTemplate extends Module
       { group }    = client.r # we have revived JGroup and JUser here ~ GG
       { delegate } = client.connection
 
+      # TODO(rjeczalik): move app/util/getPokemonName.coffee here
+      # and refactor JStackTemplate.create to not require title;
+      # if title is missing or empty, generate pokemon name as
+      # it's done currently on client side; this way we do not
+      # need to reimplement pokemon naming in kloud as well.
```
rjeczalik: /cc @gokmen

cihangir (changes requested):
could you write tests?

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
gokmen: will be fixed with #10275

rjeczalik:
@cihangir Done. Previously wanted to do that in a separate PR, since it's kind of big now. And was blocked by https://github.com/koding/koding/pull/10267. Since the cov for the import.go file is 73% I'm going to add a couple of test-cases for error paths.

go/src/koding/api/apitest/apitest.go
```diff
@@ -31,20 +36,33 @@ type HTTPIdleConnectionsCloser interface {
...
+// used with http server.
 type FakeAuth struct {
 	Sessions map[string]*api.Session
 
 	mu sync.RWMutex
 }
 
-var _ http.Handler = http.HandlerFunc((&FakeAuth{}).GetSession)
+var (
+	_ http.Handler = http.HandlerFunc((&FakeAuth{}).GetSession)
```
ppknap: This one is probably not needed. If `GetSession` is not of `http.HandlerFunc` signature, transport tests won't compile.

ppknap (approved):

go/src/koding/api/apitest/apitest.go
rjeczalik: The same applies to any other type - you could argue that blank interface variables are not needed, because a test or package elsewhere won't build.

Those blanks are also for documentation purpose, so you see in one variable group what kind of functionality does the type serve.

cihangir (approved):

PR #9491, opened 2016-11-07
resource state modal: remove copy to clipboard feature

hakankaradis:


hakankaradis:
When admin create an team stack, this modal showed up after sharing with team.
Styl updated

usirin (approved):

PR #9506, opened 2016-11-08
Fix Trial Periods to known values

cihangir:
## Description
Due to time issues sometime users might get slightly less than 30 day trial periods, fix them to pre defined time.


cihangir:
cc/ @sinan

rjeczalik (approved):

PR #9514, opened 2016-11-09
Stacktemplateshare

hakankaradis:
User1 -> share stack template
- All users can see
- All users can clone
- All admins can edit
Once shared stack template is edited by its owner or any admin, All other stacks that are generated by cloned stack template which is originally shared stack template will get notification.

User can either update the stack with new one or close the connection between this stack and original stack template so that this stack will not be notified anymore.

Here is the final screencast http://recordit.co/jS7l965kSH

Member share the stack template and admin edit it,
All stacks that are generated from clone of the shared stack template got notification,
`Don't Warn Me again` delete the `clonedFrom` and `needUpdate` attributes from stack and its stack template
`Click to Update` switch the content of stack template with the original one's content.


client/stack-editor/lib/index.coffee
```diff
@@ -64,6 +62,15 @@ module.exports = class StackEditorAppController extends AppController
           @removeEditor stackTemplate._id
           kd.singletons.router.handleRoute '/IDE'
 
+        isClonedTemplate stackTemplate, (originalTemplate) =>
+          if originalTemplate
+            editor.addClonedFrom originalTemplate
+            originalTemplate.on 'update', (update) ->
+              if stackTemplate.template.sum isnt originalTemplate.template.sum
+                return editor.addCloneUpdateView originalTemplate
+            if stackTemplate.template.sum isnt originalTemplate.template.sum
```
hakankaradis: It looks weird but when stack-editor view is open for member, and when admin update the original stack template, I need to add the update event here.

client/stack-editor/lib/index.coffee
```diff
@@ -167,33 +176,33 @@ module.exports = class StackEditorAppController extends AppController
...
-            callback: =>
-              view.destroy()
-              delete @editors[stackTemplate._id]
-              @openEditor stackTemplate._id
-        content: '''
-          <div class='modalformline'
-            <p>This stack template is updated by another admin. Do you want to reload?</p>
-          </div>
-          '''
+    # stackTemplate.on? 'update', =>
```
hakankaradis: I commented out those lines because of #9527.
I think we need to get rid of those lines? @gokmen

client/stack-editor/lib/styl/stackeditor.styl
```diff
@@ -190,6 +191,84 @@ $borderColor                = #E3E3E3
...
+      display               inline-block
+      cursor                pointer
+      float                 right
+      top                   -16px
+      right                 24px
+      r-sprite              stack-editor dontwarnmeagain
+      bg                    color transparent
+      box-sizing            border-box
+
+    div.tooltip span.tooltiptext
```
hakankaradis: I'll update tooltip style soon.

client/stack-editor/lib/index.coffee
gokmen: nope we need to fix that issue, not sure how.

workers/social/lib/social/traits/notifiable.coffee
```diff
@@ -33,6 +33,25 @@ module.exports = class Notifiable
           account.sendNotification 'InstanceChanged', opts
 
 
+  updateTemplateAndNotify: (options, change, callback) ->
```
gokmen: traits shouldn't include something specific to a model, they are generic. we can move this into `JStackTemplate` model instead.

gokmen (changes requested):
LGTM in general just small changes and we can have another session before taking this in.

client/stack-editor/lib/index.coffee
gokmen: @hakankaradis ignore this comment, let's remove the codeblock instead of commenting out

hakankaradis:
@gokmen http://recordit.co/xbkeG2RR5N

gokmen:
@hakankaradis can u assign back to me once it's ready, thanks!

client/app/lib/components/sidebar/index.coffee
```diff
@@ -106,6 +108,10 @@ module.exports = class Sidebar extends React.Component
         computeController.makeTeamDefault draft.toJS(), yes
       when 'Delete'
         computeController.deleteStackTemplate draft.toJS(), yes
+      when 'Share With Team'
+        computeController.sharingStackTemplate draft.get('_id'), 'group'
```
gokmen: can we rename this function to `shareStackTemplate`

client/app/lib/components/sidebar/index.coffee
```diff
@@ -124,10 +130,21 @@ module.exports = class Sidebar extends React.Component
 
     if draft.getIn ['config', 'remoteDetails', 'originalUrl']
       menuItems['Open on GitLab'] = { callback }
+    if whoami()._id is draft.get 'originId'
+      menuItems['Edit'] = { callback }
```
gokmen: won't this change prevent admins to edit team stacks from sidebar?

client/app/lib/components/sidebar/index.coffee
```diff
@@ -124,10 +130,21 @@ module.exports = class Sidebar extends React.Component
...
+    else
+      menuItems['View Stack'] = { callback }
```
gokmen: hmm this one covers that case, ok.

client/app/lib/components/sidebar/index.coffee
```diff
@@ -124,10 +130,21 @@ module.exports = class Sidebar extends React.Component
...
+    menuItems['Initialize'] = { callback }
     menuItems['Clone'] = { callback }  if canCreateStacks() or isAdmin()
-    menuItems['Make Team Default'] = { callback } if isAdmin() and draft.get('machines').size
+
+    if draft.get('machines').size
+      menuItems['Make Team Default'] = { callback } if isAdmin()
+      if whoami()._id is draft.get('originId')
+        if draft.get('accessLevel') is 'private'
+        then menuItems['Share With Team'] = { callback }
+        else  menuItems['Make Private'] = { callback }
```
gokmen: 👍

client/app/lib/providers/computecontroller.coffee
```diff
@@ -997,6 +999,47 @@ module.exports = class ComputeController extends KDController
           @emit 'StacksInconsistent', stack
 
 
+  sharingStackTemplate: (_id, type) ->
+    remote.api.JStackTemplate.one { _id }, (err, template) ->
+      template.setAccess type
```
gokmen: let's remove this helper from here and use where needed as is and rename it as `setTemplateAccessLevel`

client/app/lib/providers/computecontroller.coffee
```diff
@@ -997,6 +999,47 @@ module.exports = class ComputeController extends KDController
...
+
+
+  setStackTemplateAccessLevel: (params) ->
```
gokmen: maybe we can move this out of computecontroller since it's not doing anything related with computecontroller but stores it seems.

client/app/lib/providers/computecontroller.coffee
```diff
@@ -997,6 +999,47 @@ module.exports = class ComputeController extends KDController
...
+    stacks.forEach (stack) ->
+      config  = stack.config ?= {}
+      config.needUpdate = group
+      unless group
+        delete config.needUpdate
+        delete config.clonedFrom
+      stack.modify { config }, (err) ->
+        stack.config = config
+        reactor.dispatch 'STACK_UPDATED', stack
+
```
gokmen: 👍

client/app/lib/providers/computecontroller.coffee
```diff
@@ -1258,6 +1301,33 @@ module.exports = class ComputeController extends KDController
     else showError 'Failed to share credential'
 
 
+  removeClonedFromAttr: (stackTemplate, callback = kd.noop) ->
```
gokmen: I think we can move this functionality to `JStackTemplate` itself.

gokmen (changes requested):
LGTM but there are some changes required for code structure let's have another session to finalise those issues. Then we can merge this in, thanks!

client/app/lib/providers/computecontroller.coffee
hakankaradis: What helper?

client/app/lib/providers/computecontroller.coffee
hakankaradis: This function clean the JComputeStack and JStackTemplate instance's config information.
And update instance and stores.
Do you want me to put this functionality to `JStackTemplate` just for clean `clonedFrom` and `needUpdate` attributes?

client/app/lib/providers/computecontroller.coffee
hakankaradis: Actually it does, It is using `checkRevisionFromOrifinalStackTemplate` and update the JComputeStack instance.

client/app/lib/providers/computecontroller.coffee
hakankaradis: do you want me to update this function name?
https://github.com/koding/koding/blob/master/workers/social/lib/social/models/computeproviders/stacktemplate.coffee#L56

client/app/lib/providers/computecontroller.coffee
gokmen: yes, it does. but it's updating stacks individually which needs to be handled by stack itself.

client/app/lib/providers/computecontroller.coffee
gokmen: yes, we may need to move some functionality into backend like these. we'll need such functionality over api support and in that case we'll end-up with duplicate codes.

client/app/lib/providers/computecontroller.coffee
gokmen: yes, `sharingStackTemplate` tells something ongoing but we need to put the action into method name not the state.

gokmen (approved):

PR #9524, opened 2016-11-09
cleanup | workers: remove JMailNotification

cihangir:


gokmen (approved):
it looks good to me, maybe at some point we also need to revise `notify` trait usages. 👍

Issue #9136, opened 2016-09-28
terraformer: convert state file storage to tree structure


rjeczalik:
Currently we store tf state files for team stacks using flat directory structure:

```
tf-bucket/
├── ${GroupName}-${StackID}                            <-- apply state
├── ${Provider}-${GroupName}-${Identifier}             <-- default key state
└── ${awsAccountID}-${GroupName}-${CredIdentifier}     <-- bootstrap state (AWS)
```

Instead it should have tree structure to enable easier and provider-agnostic way of doing cleaning for bootstrap resources that were not destroyed. I propose the following structure:

```
tf-bucket/
└── ${Provider}
    └── ${CredIdentifier}
        ├── ${GroupName}
        │   └── ${StackTemplateID}     <--- apply state 
        └── bootstrap                  <--- boostrap state
```

It would enable us for creating a cleaner, that'd recursively read state files under `${Provider}/{CredIdentifier}/bootstrap` keys and just destroy them. Currently it's not possible.

In order to get there we need to add backward compatibility mode first, so any bootstrap data or stacks created under old key is either migrated to new one or continues to work.

Also make Kloud to save bootstrap and apply keys (content IDs) to jCredentials and/or jComputeStacks. So the content IDs are read/written to database, so any future changes are easier to perform.

rjeczalik:
Prefix bootstrap with region/location, if we decided to allow for multi-region bootstraps, e.g.:

```
tf-bucket/
└── ${Provider}
    └── ${CredIdentifier}
        ├── ${GroupID}
        │   └── ${StackTemplateID}     <--- apply state 
        └── bootstrap
            ├── us-east-1              <--- boostrap state
            └── eu-west-1              <--- boostrap state
```

rjeczalik:
Drop any timestamp / random value from key name and use jCredential.identifier if necessary. (/cc @ppknap)

rjeczalik:
A number of more changes would be required here:

- ensure each document like jComputeStack, jCredential etc. has terraformer key set
- using the above add a migration from old to new key format
- ensure stacks use jStackTemplate._id in the key instead of jComputeStack._id - the difference here's the latter changes each reinitialization, thus each time new tf state file is created; instead the state file should be preserved across multiple of builds for a single stack

(/cc @ppknap @gokmen)

cihangir:
Are you planning to keep each tf state in one folder if they belong to same provider?

rjeczalik:
[...]

PR #9542, opened 2016-11-11
provider/google: add var.koding_stack_id variable to google templates

ppknap:
This PR introduces `koding_stack_id` variable that changes its value every time when stack is modified.

## Description
There are two facts that implies this issue:
- Google cloud platform requires unique name for its resources.
- Kloud deletes old resources and creates new ones asynchronously.

Because of this, stack re-initialization fails when resource name doesn't change:

To reproduce:
- create and initialize google stack.
- change any field(apart from name) in stack template.
- you will be asked to re-initialize it.
- during re-initialization, you will receive following error:

image

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:
- [x] My change requires a change to the documentation.

@gokmen in order to fix this bug, you will need to replace: [this](https://github.com/koding/koding/blob/master/client/app/lib/providers/templates/google.coffee#L17) and [this](https://github.com/koding/koding/blob/master/client/app/lib/providers/templates/google.coffee#L49) lines with:

`name: '${var.koding_user_username}-${var.koding_group_slug}-${var.koding_stack_id}'`

this change will be also consistent with AWS resource [naming](https://github.com/koding/koding/blob/master/client/app/lib/providers/templates/aws.coffee#L17).

### DIFF
--- a/go/src/koding/kites/kloud/provider/google/stack.go
+++ b/go/src/koding/kites/kloud/provider/google/stack.go
@@ -258,7 +258,7 @@ func (s *Stack) ApplyTemplate(c *stack.Credential) (*stack.Template, error) {
 	t.Resource["google_compute_instance"] = resource.GCInstance
 
 	t.Variable["koding_stack_id"] = map[string]interface{}{
-		"default": shorten(s.id(), 8),
+		"default": shortN(s.id(), 8),
 	}
 
 	err = t.ShadowVariables("FORBIDDEN", "google_credentials")
@@ -377,7 +377,7 @@ func (s *Stack) id() string {
 	}
 }
 
-func shorten(str string, size int) string {
+func shortN(str string, size int) string {
 	sum := sha1.Sum([]byte(str))
 	hash := hex.EncodeToString(sum[:])
 
--- a/go/src/koding/kites/kloud/provider/google/stack.go
+++ b/go/src/koding/kites/kloud/provider/google/stack.go
@@ -2,10 +2,13 @@ package google
 
 import (
 	"bytes"
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"strconv"
 	"strings"
 	"text/template"
+	"time"
 
 	"koding/kites/kloud/stack"
 	"koding/kites/kloud/stack/provider"
@@ -254,6 +257,10 @@ func (s *Stack) ApplyTemplate(c *stack.Credential) (*stack.Template, error) {
 
 	t.Resource["google_compute_instance"] = resource.GCInstance
 
+	t.Variable["koding_stack_id"] = map[string]interface{}{
+		"default": shorten(s.id(), 8),
+	}
+
 	err = t.ShadowVariables("FORBIDDEN", "google_credentials")
 	if err != nil {
 		return nil, err
@@ -357,6 +364,28 @@ func (s *Stack) getDiskSize(currentProject string, computeService *compute.Servi
 
 		return int(computeImg.DiskSizeGb)
 	}
+}
+
+func (s *Stack) id() string {
+	switch arg := s.Arg.(type) {
+	case *stack.ApplyRequest:
+		return arg.StackID
+	case *stack.PlanRequest:
+		return arg.StackTemplateID
+	default:
+		return strconv.FormatInt(time.Now().UTC().UnixNano(), 10)
+	}
+}
+
+func shorten(str string, size int) string {
+	sum := sha1.Sum([]byte(str))
+	hash := hex.EncodeToString(sum[:])
+
+	if len(hash) < size {
+		return hash
+	}
+
+	return hash[:size]
 }
 
 func mustAsset(s string) string {


cihangir:
I dont think this PR solves the real problem, i am not sure we can enforce the naming schema to the users. In AWS case it is just a placeholder for the machine name, they can change it whenever they want.

cc/ @gokmen

gokmen:
@ppknap I agree with @cihangir here we shouldn't rely on something like that to make it work.

ppknap:
@cihangir @gokemn we will not enforcing anything, user is still able to change machine name to whatever string he wants. The problem with GCP is that resource name must be unique so the user will have to change the name each time he/she change the stack template.

The other solution is to change current stack re-initialization logic i.e: make destroying old stack and creating new one synchronous. But, this may drastically increase stack re-initialization times.

I'm open to other ideas.

rjeczalik:
FWIW we do the same for marathon and azure. The difference between them and aws is that aws does not require from user to input unique instance ID - it takes care of generating one. Azure requires user to provider unique name for each instance - [so if user does not provide one, kloud generates one](https://github.com/koding/koding/blob/master/go/src/koding/kites/kloud/provider/azure/stack.go#L179-L181). The same for Marathon.

go/src/koding/kites/kloud/provider/google/stack.go
```diff
@@ -254,6 +257,10 @@ func (s *Stack) ApplyTemplate(c *stack.Credential) (*stack.Template, error) {
 
 	t.Resource["google_compute_instance"] = resource.GCInstance
 
+	t.Variable["koding_stack_id"] = map[string]interface{}{
```
rjeczalik: I don't at this point we should go into exporting `koding_stack_id` to the user of the template, instead we should just generate unique name for the instance if not set explicitly.

When we're done with improving kloud, then we can improve it for more advanced users by exporting all the needed IDs and adding support for external cloud-init. For now there's no much gain in my opinion in exporting just one ID. I'm fan of full solution, where we design an API first and look back at what we have right now.

ppknap: How then user will be able to ensure unique naming for other resources than `google_compute_instance`?

Lets say user defines `google_compute_address` resource in order to create EIP for his instances. Each resource in GCP requires unique name. How we handle this without `koding_stack_id`?

Also yes, we should expose some IDs in more generic way as well as we should allow user to provide his own cloud-init. But this is much bigger task and should not change `koding_stack_id` variable in the future.

rjeczalik: This is the very same problem which is not solved for AWS and other providers. We just did not hit the problem so far, so my take is to not half-solve-it. Just wait when it becomes a problem, and then design full solution.

I know it's unrelated, but I'd want to solve it by adding concept of koding stacks (generic across providers, right now each stack handles things like `debug`, `alwaysOn` on each own) and solve [this one at once](https://github.com/koding/koding/issues/9136). We're just not there yet, and I would want to restrain from extending the templates until needed.

You add here `koding_stack_id` and next thing what will happen is user asking how he could use it in AWS.

gokmen: why don't we simply disable the `name` field and set it by kloud on it's own?

ppknap: @gokmen do you want to dissable `name` field for every of ~40 resource types GCP provider can define?

gokmen: @ppknap instead of disabling what about adding the `stack_id` at the end of it automatically?

ppknap: Eg.
```yaml
# Here is your stack preview
# You can make advanced changes like modifying your VM,
# installing packages, and running shell commands.
provider:
  google:
    credentials: '${var.google_credentials}'
    project: '${var.google_project}'
    region: '${var.google_region}'
resource:
  google_compute_address:
    google_ca:
      name: 'user-defined-name'
  google_compute_instance:
    # this is the name of your VM
    google-instance:
      # and this is its identifier (required)
      name: google-instance
      # select your instance_type here: eg. n1-standard-1
      machine_type: f1-micro
      # base image for your instance
      disk:
        image: ubuntu-1404-lts
      # select your instance zone which must be in provided region: eg. us-central1-a
      zone: us-central1-a
```
We need to replace "google_ca.name" as well. Because it will fail as already existing resource during re-initialization.

ppknap: > @ppknap instead of disabling what about adding the stack_id at the end of it automatically?

We need to do this for all possible resource types in GCP. I want to make this explicit in order to let user decide.

ppknap:
@cihangir @gokmen @rjeczalik I want to write a quick summary why this PR exists and why this is a real problem.

Some of new providers like GCP and Azure require unique identifiers for each resource they mange. Currently, our logic during stack re-initialization tries to create new resources before destroying the old ones. For example, if we have stack template like this:

```yaml
resource:
  google_compute_address:
    google_ca:
      name: resource-name
```  

After stack initialization CGP will create google compute address with unique name `resource-name`.

If we change our template, we will need to reinitialize it. At this step Kloud will try to create another google compute address with already taken unique name `resource-name`. This will fail due to resource name collision.

There are few possible solutions:

- introduce a variable which changes its value each time stack tamplate is changed. Thus, template name will have `name: resource-name-${var.koding_xxx}` syntax. This explicitly shows that we will add some kind of index to each resource name.

- silently attach unique ID during stack apply step. This implicitly enforces naming schema. So, if the user specifies "name: resource-name" it won't end up with `resource-name` but some kind of `resource-name-XXXXXX`. We will need to change name for each possible resources that can be created with given provider.

- make stack re-initialization synchronous. This is probably the right and best solution. However, after digging into the code and asking Rafal, we both agreed that this will require a huge amount of work and effort to make stack re-initialization behave in such way.

I proposed the most flexible solution for now. If you don't like it. We can close this PR and left it as is. I'm open for other ideas and solutions.

PS. @diclee could you please add to QA provider tests following use case:
- Create a base stack (with default template)
- Initialize it.
- Go back to stack template and change eq. user data string from "hello world" to "hello koding".
- You will be asked to reinitialize the stack.
- Check if re-initialization succeeds.

ppknap:
@cihangir @gokmen @rjeczalik what are we going to do with this PR?

gokmen:
@ppknap I still voting for this one;
> silently attach unique ID during stack apply step. This implicitly enforces naming schema. So, if the user specifies "name: resource-name" it won't end up with resource-name but some kind of resource-name-XXXXXX. We will need to change name for each possible resources that can be created with given provider.

as I said here before https://github.com/koding/koding/pull/9542#discussion_r87588700

ppknap:
So there is 1-1 :). Note that I'm not forcing this PR in any way. Just pointed out the problem and trying to find the solution on which we all agree.

rjeczalik:
Whatever we do it would be good to have a generic solution, as other providers also suffer from this problem. I'm ok for both explicit solution (using Koding variables as we do for AWS) or implicit one - e.g. extending providers to mark all unique fields, so they can be augmented by `*stack.Builder`.

gokmen:
@ppknap we need this some how otherwise the shared template won't work for members, since all the team will use the same stack template and they will use the same resource name as it's named in stack template. this is same and valid for all other providers and it's a must in koding case, in other words this needs to be fixed to mark gce provider as done.

ppknap:
@gokmen I completely agree. I can do the implicit solution attaching ID to instance name just as is done in Azure. But, in GCP and Azure cases this fix will work only for instances (and not for other resources). So, we need to find a generic solution for all resources which require unique name.

cihangir:
> silently attach unique ID during stack apply step. 

After some thougth process, i think i would not choose this step. There could be many problems than we expect currently. Eg: users could cross referance the the names as "strings", not by tf output vars. So if we append suffixes they will fail big time.  


> make stack re-initialization synchronous. ... we both agreed that this will require a huge amount of work and effort ...


What about not destroying the stack and just reapplying into previous tf.state? TF already handles re-creating or terminating the resources

rjeczalik:
> What about not destroying the stack and just reapplying into previous tf.state? TF already handles re-creating or terminating the resources

It will fix reinit, but won't fix stack creation - if you have a template with e.g. unique instance name, it won't be possible to share such stack with a team, if the team has more than one member - building such stack would fail due to name conflict.

cihangir:
https://github.com/koding/koding/pull/9542#issuecomment-260627595 right

I still dont support the idea of adding suffixes implicitly, we could merge this PR and advise users to either use the provided id explicitly or another way of generating unique resource names (ie : using profile_nickname in the resource name)

On the other hand we should definitely change how the reinit works, it will both improve the flow and UX from the speed perspective.

rjeczalik:
> On the other hand we should definitely change how the reinit works, it will both improve the flow and UX from the speed perspective.

I second.

ppknap:
What about merging this, but not changing stack template. I will send another PR that adds unique ID to instance name (implicity) in order to make this consistent with Azure logic. If someone adds another resource we will be able to tell him/her that it's possible to use `${var.koding_stack_id}`. This implies two issues which we will need to do (but not now) in order to make things clean and generic:

- move `${var.koding_stack_id}` (and other vars proposed by @rjeczalik) to generic metadata. (non breaking change)
- improve reinit. (should also be a non breaking change)

ppknap:
closing this as #9590 would fix it in more generic way.

PR #9561, opened 2016-11-14
klient: bug fix for goroutine leaks on kite disconnect

ppknap:
This PR fixes klient's go-routine leak caused by invalid assumption that `timer.Stop()` closes its `C` [channel](https://github.com/koding/koding/blob/master/go/src/koding/klient/app/klient.go#L487). On the contrary, it *prevents* channel from [firing](https://golang.org/pkg/time/#Timer.Stop). This means that if the time gap between two arbitrary klient's kite function calls was less that 1 minute, go-routine created on OnDisconnect callback would be leaked.

## Motivation and Context
Fixing performance issues with KD.

## How Has This Been Tested?
Manually,
Unit tests.

subset of klient goroutines after few calls to `kd list`, they are not GC at all.
```sh
koding/kites/tunnelproxy.(*Client).eventloop(0xc42026a2a0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
koding/klient/app.(*Klient).RegisterMethods.func3.1(0xc42007ccf0)
```

## Types of changes

### DIFF
--- a/go/src/koding/klient/app/defertime.go
+++ b/go/src/koding/klient/app/defertime.go
@@ -1,0 +1,82 @@
+package app
+
+import "time"
+
+// DeferTime allows to invoke registered function after provided duration. It
+// can cancel deferred function call even if internal timer was already started.
+type DeferTime struct {
+	d time.Duration
+	f func()
+
+	startC chan struct{}
+	stopC  chan struct{}
+}
+
+// NewDeferTime creates a new DeferTime instance and runs internal worker
+// go-routine. It is not possible to stop internal loop since it is meant to
+// run during the entire application's life cycle.
+func NewDeferTime(d time.Duration, f func()) *DeferTime {
+	dt := &DeferTime{
+		d:      d,
+		f:      f,
+		startC: make(chan struct{}, 1),
+		stopC:  make(chan struct{}, 1),
+	}
+
+	go dt.loop()
+
+	return dt
+}
+
+// Start runs registered function after `d` duration. If this method is called
+// multiple times before `d` duration pass, each call will reset the timer.
+// Calling Start method after registered function is invoked, will repeat the
+// process. This method is thread safe.
+func (dt *DeferTime) Start() {
+	select {
+	case dt.startC <- struct{}{}:
+	default:
+	}
+}
+
+// Stop prevents DeferTime from calling registered function even if it was
+// already scheduled by Start method. It is no-op when Start method was not
+// called. It is safe to call Stop method from multiple go-routines.
+func (dt *DeferTime) Stop() {
+	select {
+	case dt.stopC <- struct{}{}:
+	default:
+	}
+}
+
+func (dt *DeferTime) loop() {
+	var (
+		timer    *time.Timer      = nil
+		c        <-chan time.Time = nil
+		sawTimer                  = false
+	)
+
+	for {
+		select {
+		case <-c:
+			sawTimer = true
+			dt.f()
+			timer, c = nil, nil
+		case <-dt.stopC:
+			if timer != nil && !timer.Stop() && !sawTimer {
+				<-timer.C // drain previous timer since it was not triggered.
+			}
+			timer, c = nil, nil
+		case <-dt.startC:
+			if timer != nil {
+				if !timer.Stop() {
+					<-timer.C
+				}
+				timer.Reset(dt.d)
+			} else {
+				timer = time.NewTimer(dt.d)
+			}
+			c = timer.C
+		}
+	}
+}
--- a/go/src/koding/klient/app/defertime.go
+++ b/go/src/koding/klient/app/defertime.go
@@ -10,6 +10,7 @@ type DeferTime struct {
 
 	startC chan struct{}
 	stopC  chan struct{}
+	closeC chan struct{}
 }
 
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/app/defertime.go
```diff
@@ -0,0 +1,82 @@
+package app
+
+import "time"
+
+// DeferTime allows to invoke registered function after provided duration. It
+// can cancel deferred function call even if internal timer was already started.
+type DeferTime struct {
```
cihangir: Why this is superior to Golang's `time.AfterFunc` ?

ppknap: `time.AfterFunc` starts the timer immediately after called, the returned timer will still need logic for its cancellation and re-initialization.

cihangir: I still dont see the value, yours require the same logic as well?

ppknap: Ok, here is the use case:
All the collaborator's terminals must be disconnected when kite owner disconnects the klient. So we use `OnDisconnect` kite callback. If user goes back to klient we need to stop the timer and run it again when he disconnects.

I could use `time.AfterFunc` with extra cancellation/re-initialization logic added to `OnFirstRequest` and `OnDisconnect` kite callbacks. However, I decided to move the code to `DeferTime` structure in order to make it testable(it would be hard to write tests if I have to run the entire klient) and to remove `disconnectTimer` timer from Klient structure which is used only in callbacks. I think this simplifies Klient a bit.

rjeczalik (approved):

go/src/koding/klient/app/defertime.go
cihangir: Sorry but i see it as another level of indirection. You moved code to an another callback and implemented the same functionality from time.Timer + better Stop functionality while still leaking go routines. 
Please check the code here https://github.com/koding/koding/pull/9561/files#diff-3b0ed4f4068baf2a0714d960aece64bdR59 there is no code block as return statement

go/src/koding/klient/app/defertime_test.go
```diff
@@ -0,0 +1,104 @@
+package app
```
cihangir: Thanks for writing tests

go/src/koding/klient/app/defertime_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"testing"
+	"time"
+)
+
+func TestDeferTimeStart(t *testing.T) {
+	const timeToWait = 500 * time.Millisecond
```
cihangir: we could decrease the timeToWait to ~10 ms instead of 500

go/src/koding/klient/app/defertime.go
```diff
@@ -0,0 +1,82 @@
...
+	)
+
+	for {
+		select {
+		case <-c:
+			sawTimer = true
+			dt.f()
+			timer, c = nil, nil
+		case <-dt.stopC:
+			if timer != nil && !timer.Stop() && !sawTimer {
```
cihangir: 👍  for implementing the best practice for timer stop

cihangir (changes requested):
lets fix the go routine leak

go/src/koding/klient/app/defertime.go
ppknap: Yes, you are right, I'm leaking 1 go-routine which is meant to live during the entire klient's life-cycle. I wanted to add `Stop()` method but as we leak other goroutines that also have to be open during klient life, i decided to not complicate this struct. I see no point in it.

@cihangir I can inline this logic in klient, but I won't be able to test it easily. If you let me to leave it, I will add `Exit()` method which will stop *leaked* go-routine.

rjeczalik: Shouldn't the `(*kite.Kite).Close()` stop any goroutine started by *kite.Kite?

ppknap: @rjeczalik I'm thinking about long running go-routines like [this](https://github.com/koding/koding/blob/ed922385cd322370f31e2fe8c97a9da89dce1146/go/src/koding/klient/app/klient.go#L577) they are also "leaked" but this is fine, since klient object is meant to act like a singleton.

ppknap: oh you are right, I thought that `logFiles` is a channel:d. But you see my point .

ppknap:
@cihangir I added close method and increased tests speed

cihangir (approved):

cihangir:
Thanks @ppknap

rjeczalik (approved):

Issue #9528, opened 2016-11-10
Member shouldn't be able to clone a stack template when stack creation is disabled.


diclee:
## Current Behavior
![gif](http://g.recordit.co/gH5ftwgu14.gif)
http://recordit.co/gH5ftwgu14

diclee:
Verify passed on latest.

Issue #9539, opened 2016-11-10
File cannot be renamed from IDE header.


diclee:
## Current Behavior
![gif](http://g.recordit.co/l9uQe9VWwX.gif)
http://recordit.co/l9uQe9VWwX

Editable mode on double click should be fixed or removed for files.

hakankaradis:
@diclee I think the file that you are trying to rename has not been saved before.
So you cannot change its name. Can you please try again with saved file?

diclee:
@hakankaradis I've tried with a saved file and it worked. But for the sake of user experience is it possible to disable renaming for an unsaved file?

hakankaradis:
@diclee I removed the Rename menu item from dummy files. fyi

diclee:
Verify passed on latest.

Issue #9538, opened 2016-11-10
Turn off VM Sharing is not working.


diclee:
## Current Behavior
![gif](http://g.recordit.co/AUfINEKNPk.gif)
http://recordit.co/AUfINEKNPk

Even though it's displayed as turned off, the participants are still there when you turn it on. And participant can accept invitations and access that vm when it's displayed as turned off.

error log:
![image](https://cloud.githubusercontent.com/assets/8261892/20183999/dd3659a2-a76f-11e6-8ed4-308a03ea17f4.png)
![image](https://cloud.githubusercontent.com/assets/8261892/20184028/f7bba6ec-a76f-11e6-9d7c-9e3d9c8f2aa4.png)

diclee:
Verify passed on latest.

Issue #9537, opened 2016-11-10
IDE becomes unstable when there are lots of tabs.


diclee:
## Current Behavior
![gif](http://g.recordit.co/GTSgr4y61J.gif)
http://recordit.co/GTSgr4y61J
Tabs cannot be closed after having too many tabs and when file menu is not visible.

![gif](http://g.recordit.co/1rYj4XoSZh.gif)
http://recordit.co/1rYj4XoSZh

hakankaradis:
@diclee I think it is totally style issue, I think the buttons which are `menu` and `x` buttons next to file name overlaps. I will take a look

diclee:
Verify passed on latest.

PR #9562, opened 2016-11-14
Release Bugs

hakankaradis:
Fixes: #9528, #9529, #9534, #9535, #9537, #9538, #9539  

for #9535, http://recordit.co/I6uVFkUof9
for #9538, http://recordit.co/kK3PXIxsoj
for #9534, I temporarily add first machine's id to url so that at least first machine will come expanded.
I think we need to remove stack title information from the row and divide it regarding stack title
Something like this cc: @ftasdelen, wdyt?
https://monosnap.com/file/XuNbCKJdxVx5ICOd3AmbGom7KtAkuo

Also virtual machines tab section style updated, added tooltip

for #9537, http://recordit.co/7JGZTMr9li

### DIFF
--- a/client/app/lib/notificationcontroller.coffee
+++ b/client/app/lib/notificationcontroller.coffee
@@ -8,6 +8,7 @@ KDModalView        = kd.ModalView
 KDNotificationView = kd.NotificationView
 KDObject           = kd.Object
 ContentModal = require 'app/components/contentModal'
+EnvironmentFlux = require 'app/flux/environment'
 
 remote_extensions  = require 'app/remote-extensions'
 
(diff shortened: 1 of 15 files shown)


gokmen (approved):
LGTM 👍

client/app/lib/notificationcontroller.coffee
```diff
@@ -103,7 +103,6 @@ module.exports = class NotificationController extends KDObject
 
     @on 'MachineShareListUpdated', (data = {}) ->
 
-      console.log 'dattaaa ', data
```
szkl: Why don't you amend the commit introduced this?

Issue #9545, opened 2016-11-11
'Success' modal for onboarding items keeps appearing after login.


diclee:
## Expected Behavior
It should appear once only after the user completes onboarding items.

## Current Behavior
![gif](http://g.recordit.co/S4L9zccllV.gif)
http://recordit.co/S4L9zccllV

diclee:
Verify passed on latest.

PR #9587, opened 2016-11-16
'Success' modal for onboarding items keeps appearing after login

hakankaradis:
Aim of this pr is preventing to show the 'Success' modal for onboarding items if it is shown before.

Instead of fetching `OnboardingSuccessModalShown` from appStorage, extracted from appStorage object.

Fixes: #9545 

http://recordit.co/pzvfoLlI0t
And able to write it to database

### DIFF
--- a/client/home/lib/welcome/welcomemodal.coffee
+++ b/client/home/lib/welcome/welcomemodal.coffee
@@ -31,32 +31,33 @@ module.exports = class WelcomeModal extends kd.ModalView
   showCongratulationModal: ->
 
     { appStorageController } = kd.singletons
-
     appStorage = appStorageController.storage "WelcomeSteps-#{globals.currentGroup.slug}"
-    appStorage.fetchValue 'OnboardingSuccessModalShown', (result) ->
+    buckets = appStorage.storage.bucket
+    welcomeStepsStore = buckets["WelcomeSteps-#{globals.currentGroup.slug}"]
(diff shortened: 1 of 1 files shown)


client/home/lib/welcome/welcomemodal.coffee
```diff
@@ -31,32 +31,33 @@ module.exports = class WelcomeModal extends kd.ModalView
   showCongratulationModal: ->
 
     { appStorageController } = kd.singletons
-
     appStorage = appStorageController.storage "WelcomeSteps-#{globals.currentGroup.slug}"
-    appStorage.fetchValue 'OnboardingSuccessModalShown', (result) ->
```
hakankaradis: fetchValue is not working at will. so decided to get `OnboardingSuccessModalShown` attribute from appStorage??

gokmen: weird, they are coming from same place.

hakankaradis: @gokmen not sure but when I debug I had some experience
https://github.com/koding/koding/blob/master/client/app/lib/appstorage.coffee#L72
In fetchValue function, It is called again and again before it returns the callback,
also had same experience with setValue when I pass callback.

gokmen (approved):
Approving this one for now to make this fix in as is but not convinced yet about the solution, there must be something else in somewhere else.

PR #9590, opened 2016-11-16
stack/provider: add new koding vars / add SSH key injecting during boostrap

rjeczalik:
This PR adds two following koding variables for use with stack templates:

- ${var.koding_stack_id}
- ${var.koding_template_id}

Supersedes #9542.


cihangir (approved):

client/app/lib/providers/templates/google.coffee
```diff
@@ -46,7 +46,7 @@ resource:
     # this is the name of your VM
     google-instance:
       # and this is its identifier (required)
-      name: google-instance
+      name: 'koding-${var.koding_group_slug}-${var.koding_stack_id}-${count.index+1}'
```
ppknap: Could you remove https://github.com/koding/koding/blob/c8c894efcb80d89cd4d49f43ef737a1bc148c289/go/src/koding/kites/kloud/provider/google/stack.go#L201 from google as well?

ppknap (approved):

client/app/lib/providers/templates/google.coffee
rjeczalik: Done.

gokmen (approved):

rjeczalik:
I've added also a fix for Digital Ocean. Please do not merge it yet, as Digital Ocean most likely also suffers from the unique name problem, but I need to test it first to confirm.

rjeczalik:
ptal

rjeczalik:
ptal

ppknap:
@gokmen I wanted to test this change but I cannot initialize google provider 
image

It works for aws, I assume this is something with credentials.

ppknap:
Credentials were added to mongo.
```
> db.jCredentials.find().pretty()                                                                                                                                            
{
        "meta" : {
                "createdAt" : ISODate("2016-11-16T13:56:13.854Z"),
                "modifiedAt" : ISODate("2016-11-16T13:56:13.854Z")
        },
        "verified" : true,
        "accessLevel" : "private",
        "provider" : "google",
        "title" : "Pawel google",
        "identifier" : "227c458b6969906418ad470d1ef4f411",
        "originId" : ObjectId("582c6546603100fc525e9d0e"),
        "_id" : ObjectId("582c657dadbc91165369221f")
}
```

rjeczalik:
> Credentials were added to mongo.

This is only jCredential, what about the actual credential?

ppknap:
@rjeczalik they are in social api and look correct:
image

rjeczalik:
@gokmen Thanks, works for me.

@cihangir @ppknap Ptal, so we can merge and deploy to sandbox.

go/src/koding/tools/util/pretty.go
```diff
@@ -0,0 +1,27 @@
...
+func (p PrettyJSON) String() string {
+	var v interface{}
+
+	if err := json.Unmarshal([]byte(p), &v); err != nil {
+		return fmt.Sprintf("!JSON(%s)", []byte(p))
+	}
+
+	p, err := json.MarshalIndent(v, "", "\t")
+	if err != nil {
+		return fmt.Sprintf("!JSON(%s)", []byte(p))
```
ppknap: In case of `err != nil`, `p` will always be nil. Could you please not replace `p` during `MarshalIndent`?

PR #9595, opened 2016-11-16
ide/drawing: fix drawing pane error

### LINKED ISSUE #9523: 'Drawing Board' feature is broken.
## Current Behavior
gif
http://recordit.co/VW66sy7qmT

hakankaradis:
fixed #9523 
Somehow @$canvas.getDomElement() wasn't returning the correct dom element.
Add cssClass attribute and used jquery to find dom element.

http://recordit.co/nNKIkuVHFE

### DIFF
--- a/client/ide/lib/workspace/panes/idedrawingpane.coffee
+++ b/client/ide/lib/workspace/panes/idedrawingpane.coffee
@@ -30,6 +30,7 @@ module.exports = class IDEDrawingPane extends IDEPane
   createCanvas: ->
 
     @canvas      = new KDCustomHTMLView
+      cssClass   : 'drawing'
       tagName    : 'canvas'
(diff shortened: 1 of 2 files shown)


client/ide/lib/workspace/panes/idedrawingpane.coffee
```diff
@@ -110,12 +111,12 @@ module.exports = class IDEDrawingPane extends IDEPane
 
   init: ->
 
-    @$canvas = @canvas.getDomElement()
+    @$canvas = $('canvas.drawing')
```
gokmen: this is not a correct solution, but let's merge this one for now. we need to revert this later. this selector looks dangerous to me.

gokmen (approved):

PR #9637, opened 2016-11-19
Signup fixes

sinan:
## Description
 - [x] fixed signup flow fonts
 - [x] removed typekit from the app
 - [x] changed meta titles and descriptions

before:
before

after:
after


sinan:
cc/ @gokmen @cihangir

usirin (approved):

Issue #9641, opened 2016-11-21
kd: add stack subcommand


rjeczalik:
Adds support for kd to create and list stacks.

TODO

- [x] kloud: add backend for stack.import
- [x] kd: add frontend for stack.import
- [x] remoteapi: add client
- [x] kloud: integrate remoteapi with stack.import

Implements:

`kd stack create -f stack.hcl -c my-cred`

rjeczalik:
Done by #10229.

Issue #9640, opened 2016-11-21
kd: add credential subcommand


rjeczalik:
Adds support for kd to list and create credentials.

TODO

- [x] kloud: backend for credential.list (https://github.com/koding/koding/pull/9182)
- [x] kloud: backend for credential.create (https://github.com/koding/koding/pull/9182)
- [x] kloud: backend for credential.describe (https://github.com/koding/koding/pull/9182)
- [x] kd: frontend for credential.list (https://github.com/koding/koding/pull/9182)
- [x] kd: frontend for credential.create (https://github.com/koding/koding/pull/9182)
- [x] kd: frontend for credential.describe

The describe method for credentials is used to implement interactive mode for credential creation.

Implements:

`kd credential create -f credential.json -title <title>`
`kd credential create`
`kd credential list`

PR #9666, opened 2016-11-22
kd: rework credential / stack commands

rjeczalik:
Prerequisite for https://github.com/koding/koding/issues/9640 and https://github.com/koding/koding/issues/9641.


go/src/koding/kites/kloud/utils/object/builders.go
```diff
@@ -16,3 +18,17 @@ var MetaBuilder = &Builder{
 	Prefix:    "meta",
 	Recursive: true,
 }
+
+// TabPrinter is used to print any slice of values
+// encoded with a tabwriter.
+var TabPrinter = &Printer{
```
rjeczalik: /cc @ppknap

go/src/koding/kites/config/metadata.go
```diff
@@ -161,6 +162,39 @@ func chown(file string, u *user.User) error {
...
+func FixOwner(home string, owner *user.User) error {
+	if home == "" {
+		home = KodingHome()
+	}
+
+	if owner == nil {
+		owner = CurrentUser
+	}
+
+	// Don't change owner
```
cihangir: 👍

go/src/koding/kites/kloud/utils/object/printer.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return enc.Encode(v)
+}
+
+var (
+	fprintf  = reflect.ValueOf(fmt.Fprintf)
+	stringer = reflect.TypeOf((*fmt.Stringer)(nil)).Elem()
+)
+
+func (p *Printer) printTab(v interface{}) error {
```
cihangir: I believe we can implement the same functionality with

* encode to json 
* decode the result to map[string]interface{}
* get the keys and output them.

we can also add embedded field support easily with a recursive function.

Lets not use reflection whenever possible in our app logic.

ie: getting recursive field names
```
func getKeys(prefix string, data map[string]interface{}) {
	for key, val := range data {
		fieldKey := prefix + key
                // do your thingy with fieldKey
		next, ok := val.(map[string]interface{})
		if ok {
			getKeys(fieldKey+".", next)
		}
	}
}
```

rjeczalik: Will take a try at reworking this.

rjeczalik: > this may not work for empty arrays since they will be decoded to [ ]

Yes, this plus the columns would always be ordered alphabetically, so it won't be possible order them per-struct. I can make some hack, if column == "id" then display it first etc. However it would not work the same.

go/src/koding/klientctl/kloud/stack/stack.go
```diff
@@ -0,0 +1,223 @@
...
+// hcl.Unmarshal encodes each object as []map[string]interface{},
+// and kloud expects JSON objects to not be wrapped in a 1-element
+// slice.
+//
+// This function converts []map[string]interface{} to map[string]interface{}
+// if length of the slice is 1.
+//
+// BUG(rjeczalik): This is going to break templates, which have legit
+// 1-element []map[string]interface{} values.
+func fixHCL(v interface{}) {
```
cihangir: 👍

go/src/koding/klientctl/kloud/stack/stack.go
```diff
@@ -0,0 +1,223 @@
...
+var DefaultClient Client
+
+// Client
+type Client struct {
+	Kloud      *kloud.Client
+	Credential *credential.Client
+}
+
+// Create
+func (c *Client) Create(opts *CreateOptions) (*stack.ImportResponse, error) {
```
cihangir: could you write test for this function?

go/src/koding/kites/kloud/utils/object/printer.go
cihangir: @ppknap could you please show me an example where would you need to show the Empty values of an array?

@rjeczalik you will have the `fieldKey` you can order them however you like.

cihangir (changes requested):

cihangir:
@rjeczalik this is for you  https://koding.slack.com/archives/engineering/p1479810522000885 ;)

go/src/koding/kites/kloud/utils/object/printer.go
ppknap: @cihangir I think about showing headers(keys) when th array is empty. Example from docker:
```
(archpk) {machine_mount} ~/koding/koding/go/src/koding$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

The same behavior is in `kd (?:machine)? list` (I'm not able to show it because my klient is currently down -.-)

cihangir: You have the key there i am asking on the values of a key where that is another array and that array has empty values.

With the map[string]interface, you have the keys, same with a struct.

go/src/koding/klientctl/kloud/stack/stack.go
rjeczalik: What kind of test are you thinking of?

go/src/koding/kites/kloud/utils/object/printer.go
cihangir: On the other hand, probably we wouldnt want to show each and every field of a response in the cli output, we will require some kind of filtering and that fieldKey solves that problem as well

rjeczalik: I'm going to just remove this generic tab / json printer and use tabwriter package directly in cli.

cihangir: Completely diff issue: if we also use json there, we would allow users to pass their own template - ie: http://kubernetes.io/docs/user-guide/jsonpath/ ( not saying we should implement this right now )

rjeczalik: @cihangir Agreed, JSONPath and text/template support (`kubectl -t '{{.Something}}'`) are really cool features for scripting with tool. For now they can be achieved with `jq` since our backend API does not support patching / server-side filtering - so there's no difference whether we do:

```
kd stack list | jq .Name
```

or

```
kd stack list -jsonpath .Name
````

ppknap: > With the map[string]interface, you have the keys, same with a struct.

yes, but objects passed to this function are meant to be arrays so we will not have `map[string]interface{}` but rather `[]map[string]interface{}/[]interface{}` and when `len(passed_argument) == 0` we will have no keys.

go/src/koding/klientctl/kloud/stack/stack.go
rjeczalik: ping

ppknap (approved):

go/src/koding/klientctl/kloud/stack/stack.go
cihangir: unit test?

rjeczalik: Done.

cihangir (approved):

PR #9704, opened 2016-11-23
Rename kodingdev_master key variable

szkl:


gokmen (approved):

cihangir:
@szkl could you please revert the name change? It awfully sucks while trying to recycle keys not knowing which key is which

szkl:
I've removed the timestamp in default config because it doesn't actually have a key. I didn't see any reason to rename the variable.

cihangir:
the change you did in vault is fine, ignore my comment here

PR #9760, opened 2016-11-25
Credential improvements

gokmen:
With the current implementation we're shadowing sensitive keys per provider when `::fetchData` is requested. But, with new providers we have some information needs to remain visible to owners but also needs to be shadowed for others. This PR fixes that issue and introduces a new field type called `secretKeys` for `JCredential` per provider. Provider implementations defines that as we do for `sensitiveKeys`. This new `secretKeys` will be shadowed to everyone like `sensitiveKeys` before, but with this PR `sensitiveKeys` will be visible to owners but not the others.

ref: http://recordit.co/ZC5KTwcEcq

### DIFF
--- a/workers/social/lib/social/models/computeproviders/softlayer.coffee
+++ b/workers/social/lib/social/models/computeproviders/softlayer.coffee
@@ -9,11 +9,11 @@ KodingError       = require '../../error'
 
 module.exports = class Softlayer extends ProviderInterface
 
-  @providerSlug = 'softlayer'
+  @providerSlug  = 'softlayer'
 
   @bootstrapKeys = ['key_id']
 
-  @sensitiveKeys = ['api_key']
+  @secretKeys    = ['api_key']
 
 
   @ping = (client, options, callback) ->
--- a/workers/social/lib/social/models/computeproviders/softlayer.coffee
+++ b/workers/social/lib/social/models/computeproviders/softlayer.coffee
@@ -15,6 +15,8 @@ module.exports = class Softlayer extends ProviderInterface
 
   @secretKeys    = ['api_key']
 
+  @sensitiveKeys = ['ssh_private_key', 'ssh_public_key']
+
 
   @ping = (client, options, callback) ->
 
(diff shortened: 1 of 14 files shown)


workers/social/lib/social/models/computeproviders/softlayer.coffee
```diff
@@ -9,11 +9,13 @@ KodingError       = require '../../error'
...
 
-  @providerSlug = 'softlayer'
+  @providerSlug  = 'softlayer'
 
   @bootstrapKeys = ['key_id']
 
-  @sensitiveKeys = ['api_key']
+  @secretKeys    = ['api_key']
+
+  @sensitiveKeys = ['ssh_private_key', 'ssh_public_key']
```
rjeczalik: We could make this global - `ssh_private_key` and `ssh_public_key` are going to be effectively a part of credentials for every provider.

gokmen: sure, it's easy 👍

workers/social/lib/social/models/group/permissionset.coffee
```diff
@@ -202,10 +208,32 @@ module.exports = class JPermissionSet extends Module
...
+            #       { permission   : 'modify object', superadmin: yes }
+            #     ]
+            #
+            #     success: (client, callback) ->
+            #       ...
+            #
+            # and this permission is granted because you were a superadmin then
+            # client._allowedPermissionIndex will become 2. But if you are owner of
+            # this instance this time it will become 0 because the first rule granted.
+            args[0]._allowedPermissionIndex = permissionIndex
```
cihangir: this may lead to unintended problems. generally we dont care about the order of permission but this adds importance now. 
If someone unknowingly passes the client around ( client is populated in bongo, but we can pass it from one function to another in models ) and if that client has `_allowedPermissionIndex` could cause issues. 
generally positional indexes are evil for me

cihangir (commented):
i dont disapprove or approve :) i need to think about how we can do it better (from my side)

workers/social/lib/social/models/group/permissionset.coffee
gokmen: you're right with your concerns, but if this `client` will be passing around this `_allowedPermissionIndex` will be overwritten on the place last it's used. so, it will be the issue on that last method it's used. if one is using this feature inside of an insecure method it's an issue with the implementation. similar cases valid for whole permit flow btw :) 

I'm open for any alternative ways to achieve this ;)

rjeczalik (approved):

PR #9784, opened 2016-11-28
vendor: add go-openapi/runtime

rjeczalik:
```
koding/remoteapi $ go generate
...
2016/11/28 14:43:29 creating "koding_client.go" in "client" as facade
Generation completed!

For this generation to compile you need to have some packages in your GOPATH:

  * github.com/go-openapi/runtime
  * golang.org/x/net/context
  * golang.org/x/net/context/ctxhttp

You can get these now with: go get -u -f ./...
```

PR #9786, opened 2016-11-28
koding/remoteapi: add client for remote.api

rjeczalik:
~~deps: https://github.com/koding/koding/pull/9784~~

### DIFF
--- a/go/src/koding/kites/config/konfig.go
+++ b/go/src/koding/kites/config/konfig.go
@@ -34,6 +34,7 @@ type Konfig struct {
 	KlientURL  string `json:"klientURL,omitempty"`
 	KloudURL   string `json:"kloudURL,omitempty"`
 	TunnelURL  string `json:"tunnelURL,omitempty"`
+	RemoteURL  string `json:"remoteURL,omitempty"`
 	IPURL      string `json:"ipURL,omitempty"`
 	IPCheckURL string `json:"ipCheckURL,omitempty"`
 
@@ -135,6 +136,7 @@ func NewKonfig(e *Environments) *Konfig {
 		KontrolURL:         Builtin.Endpoints.Kontrol,
 		KloudURL:           Builtin.Endpoints.Kloud,
 		TunnelURL:          Builtin.Endpoints.TunnelServer,
+		RemoteURL:          Builtin.Endpoints.RemoteAPI,
 		IPURL:              Builtin.Endpoints.IP,
 		IPCheckURL:         Builtin.Endpoints.IPCheck,
 		KlientLatestURL:    ReplaceEnv(Builtin.Endpoints.KlientLatest, e.klientEnv()),
--- a/go/src/koding/remoteapi/remoteapi.go
+++ b/go/src/koding/remoteapi/remoteapi.go
@@ -63,10 +63,6 @@ func (c *Client) New(clientID string) *client.Koding {
 	} else {
 		// NOTE(rjeczalik): optimization - since kloud and remote.api are
 		// accessible on local network, use 127.0.0.1 host instead.
-		//
-		// Requests will still have "Host: <original host>" header set,
-		// just in case the e.g. the c.Endpoint is behind nginx that
-		// is configured with host routing.
 		endpoint = copyURL(c.Endpoint)
 
 		if _, port, err := net.SplitHostPort(endpoint.Host); err == nil {
@@ -74,6 +70,11 @@ func (c *Client) New(clientID string) *client.Koding {
 		} else {
 			endpoint.Host = "127.0.0.1"
 		}
+
+		// Requests will still have "Host: <original host>" header set,
+		// just in case the e.g. the c.Endpoint is behind nginx that
+		// is configured with host routing.
+		httpClient.Transport.(*transport).Host = c.Endpoint.Host
 	}
 
 	return client.New(newRuntime(endpoint, httpClient), strfmt.Default)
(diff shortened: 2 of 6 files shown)


cihangir:
```go/src/koding/remoteapi/remoteapi.go:3: running "swagger": exec: "swagger": executable file not found in $PATH```

**Comment by rjeczalik** [comment]
@cihangir Still wip, need to add a transport for mongo authentication and actually test the client. Imo the best would be to not run go generate for koding/remoteapi in build.sh, since it will rebuild all packages that import it each time it's run.

**Review by gokmen — COMMENTED** [review]
do we need to define it as a different endpoint? I mean wouldn't be using `#{publicHostname}/remote.api` enough?

**Comment by rjeczalik** [comment]
> do we need to define it as a different endpoint? I mean wouldn't be using #{publicHostname}/remote.api enough?

What do you mean by defining it as a different endpoint?

**Comment by gokmen** [comment]
I mean do we need another config field for this one?

**Comment by rjeczalik** [comment]
I find it easier to manage. Currently we keep in coffee static list of endpoint URLs that Go codebase uses, during `./configure` we generate go/src/koding/kites/config/config.json file and we embed it in every Go binary build with ./go/build.sh.

A single endpoint on this static list is always a URL, even though part of the endpoint root at #{publicHostname}. E.g. for production we have:

- https://koding.com/kontrol/kite
- https://koding.com/kloud/kite
- https://p.koding.com/-/ip
- https://p.koding.com/-/ipCheck
- https://t.koding.com/kite

This PR adds another one:

- https://koding.com/remote.api

What you propose is to create some group of https://koding.com/-originated endpoints, and just keep Path part of the URLs in the configuration, instead of whole URLs. So the above list would become:

- https://koding.com/
  - /kontrol/kite
  - /kloud/kite
  - /remote.api
- https://p.koding.com/-/ip
- https://p.koding.com/-/ipCheck
- https://t.koding.com/kite

Do I get it right?

**Comment by gokmen** [comment]
yes, exactly this is what I meant. but in the meanwhile I also thought that it might be "somehow" better to make it separate if any case it's needed to serve from somewhere else. not sure if it's a valid concern but let's keep it as you did for now.

**Comment by ppknap** [comment]
@rjeczalik could we merge #9784 first? It's hard to see what actually changed|was_added.

**Comment by rjeczalik** [comment]
I've sent a separate PR with generated openapi client - https://github.com/koding/koding/pull/9824. Once it's merged, I'm going to rebase this PR again.

**Comment by rjeczalik** [comment]
Rebased - ptal.

File: go/src/koding/remoteapi/remoteapi.go [hunk 1]
```diff
@@ -1,3 +1,153 @@
...
+		uCopy.User = &userCopy
+	}
+
+	return &uCopy
+}
+
+func copyRequest(req *http.Request) *http.Request {
+	reqCopy := new(http.Request)
+	*reqCopy = *req
+	reqCopy.URL = new(url.URL)
```
ppknap: `reqCopy.URL = copyURL(req.URL)`

ppknap (approved):

go/src/koding/remoteapi/remoteapi.go
rjeczalik: done

rjeczalik:
@cihangir gentle ping

go/src/koding/kites/config/konfig.go
```diff
@@ -135,6 +136,7 @@ func NewKonfig(e *Environments) *Konfig {
 		KontrolURL:         Builtin.Endpoints.Kontrol,
 		KloudURL:           Builtin.Endpoints.Kloud,
 		TunnelURL:          Builtin.Endpoints.TunnelServer,
+		RemoteURL:          Builtin.Endpoints.RemoteAPI,
```
cihangir: not for this PR:

instead of passing all the urls in full format, could we just pass a koding url than append the paths?

cihangir (approved):

go/src/koding/kites/config/konfig.go
rjeczalik: I agree, but I'd want to fix more than this:

- currently we mix public / private endpoint configuration - some of the urls are https://koding.com/remote.api and some https://127.0.0.1/api/credential
- I think we should keep the URLs separate in case we'd want in future to have each service in separate container / POD
- moreover we should add two URLs per serivce - one public (for koding users) and one private (for inter-communication)

So actually I wanted to even more separate the URLs than to group them. Do we care?

rjeczalik: This is to fix hacks like this one:

https://github.com/koding/koding/pull/9786/files#diff-58db80dbc542b0ed622120b6235087c3R64

cihangir: > moreover we should add two URLs per serivce - one public (for koding users) and one private (for inter-communication)

if we have a "base" domain structure you can keep the paths separate

> I think we should keep the URLs separate in case we'd want in future to have each service in separate container / POD

in any case we will need a proxy in between services and that proxy will have its own url ( which translates to a base domain )

> So actually I wanted to even more separate the URLs than to group them. Do we care?

Indeed i care, this is creating problems while trying to remove the hardcoded & url dependent parts from koding ( ie: the things that i fixed while trying to make koding open source )

> This is to fix hacks like this one:

Infact they should not exist in the code base, you could use `customDomain.local` https://github.com/koding/koding/blob/master/config/main.dev.coffee#L65   Could be helpful for future if we have PODs/services, because localhost you know

PR #9791, opened 2016-11-28
klient: add machine address books

ppknap:
This PR adds machine address book type that stores machine addresses.

Depends on: #9789 

## Description
This is reworked architecture that is going to handle all `kd machine *` CLI commands. It *partially* deprecates `klient/remote` packages. There is a number of current issues which these changes are intended to solve:

- Support for multiple mounts per machine.
- Correct handling of destroyed machines.
- Correct handling of machines that changed their IP address.
- Correct handling of machines that changed their Kite.key.
- Correct handling of rsync mount processes.
- Clean ups after errors.
- Prevent machine duplication during `kd machine list` 
- Businnes logic and Kite transort separation.
- etc.

## Architecture
image

- **Address Book** - a set of machine addresses - including old ones.

## How Has This Been Tested?
Unit tests.

## Screenshots (if appropriate):
none

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #9793, opened 2016-11-28
klient: add machine group addresses

ppknap:
This PR adds `addresses` package that manages multiple machines' addresses.

Depends on: #9791

## Description

- Clean ups after errors.
- etc.

## Architecture
image

- **Addresses** - a set of remote machines address books.

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinegroup/addresses/addresses.go
+++ b/go/src/koding/klient/machine/machinegroup/addresses/addresses.go
@@ -1,0 +1,102 @@
+package addresses
+
+import (
+	"sync"
+
+	"koding/klient/machine"
+)
+
+// Addresses store addresses of all machines in the group.
+type Addresses struct {
+	mu sync.RWMutex
+	m  map[machine.ID]*machine.AddrBook
+}
+
+// New creates an empty Addresses object.
+func New() *Addresses {
+	return &Addresses{
+		m: make(map[machine.ID]*machine.AddrBook),
+	}
+}
+
+// Add adds provided address to a given machine.
+func (a *Addresses) Add(id machine.ID, addr machine.Addr) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	ab, ok := a.m[id]
+	if !ok {
+		ab = &machine.AddrBook{}
+	}
+
+	ab.Add(addr)
+	a.m[id] = ab
+
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


ppknap:
Note: `color_fruit` syntax instead of `fruit` is just my personal preference based on the fact that k8s and docker use two words to make aliases. We can change that easily and I won't complain.

rjeczalik:
> color_fruit syntax instead of fruit

@ppknap I still believe the fruit alias should be part of jMachine doc - e.g. `jMachine.meta.fruitAlias`, so it's possible to use the same alias on two different latops - right now the same machine will get assigned different alias. Thus creating an alias would be a responsibility of remote.api and not kd (/cc @gokmen).

ppknap:
@rjeczalik lets move this to #9794 because this is not a right place for aliases discussion.

cihangir:
could you please rebase this one too?

ppknap:
@cihangir done

cihangir (approved):

go/src/koding/klient/machine/machinegroup/addresses/addresses.go
```diff
@@ -0,0 +1,102 @@
...
+
+	ab.Add(addr)
+	a.m[id] = ab
+
+	return nil
+}
+
+// Drop removes addresses which are binded to provided machine ID.
+func (a *Addresses) Drop(id machine.ID) error {
+	a.mu.Lock()
```
rjeczalik: No need for defer:

```go
    a.mu.Lock()
    delete(a.m, id)
    a.mu.Unlock()
```

go/src/koding/klient/machine/machinegroup/addresses/addresses.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return nil
+}
+
+// Latests returns the latest known address for a given machine and network.
+func (a *Addresses) Latest(id machine.ID, network string) (machine.Addr, error) {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+
+	ab, ok := a.m[id]
```
rjeczalik: ditto

```go
a.mu.RLock()
ab, ok := a.m[id]
a.mu.RUnlock()
```

rjeczalik (approved):

PR #9794, opened 2016-11-28
klient: add machine group aliases

ppknap:
This PR adds `aliases` package that maps machine ID to user friendly name.

Depends on: #9789 

## Description

- Clean ups after errors.
- etc.

## Architecture
image

- **Aliases** - a set of machines nicknames.

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinegroup/aliases/aliases.go
+++ b/go/src/koding/klient/machine/machinegroup/aliases/aliases.go
@@ -1,0 +1,157 @@
+package aliases
+
+import (
+	"math/rand"
+	"strconv"
+	"sync"
+
+	"koding/klient/machine"
+)
+
+// colors defines the available colors for aliases. The first letter of each
+// color name must be unique across the slice.
+var colors = [...]string{
+	"aqua",
+	"black",
+	"fuchsia",
+	"green",
+	"lime",
+	"maroon",
+	"navy",
+	"olive",
+	"purple",
+	"red",
+	"silver",
+	"teal",
+	"white",
+	"yellow",
+}
+
+// fruits defines the available fruits for aliases. The first letter of each
+// fruit name must be unique across the slice.
+var fruits = [...]string{
+	"apple",
+	"banana",
+	"coconut",
+	"date",
+	"fig",
+	"grape",
+	"jackfruit",
+	"kiwi",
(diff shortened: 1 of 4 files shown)


ppknap:
this is a continuation of #9793 PR comments.

@rjeczalik even if we have aliases logic in kloud, I still need to manage them locally eg. for mounts `kd machine mount <alias>`. I took into account that aliases may be managed by kloud and I added [Add](https://github.com/koding/koding/pull/9794/files#diff-9e2c392ffc161773e7c19f6b32afe029R8) method which can substitute `Create` method when jMachines collection has aliases.

cihangir:
rebase please

ppknap:
@cihangir done

go/src/koding/klient/machine/machinegroup/aliases/aliases.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// Get current aliases.
+	current := make(map[string]struct{}, len(a.m))
+	for _, alias := range a.m {
+		current[alias] = struct{}{}
+	}
+
+	// Generate new alias.
+	for i := 0; ; i++ {
```
cihangir: lets have a limit here

cihangir (approved):

cihangir:
just sayin,  there has been a lot of discussion on labeling, and you are waking up the devils again :)

rjeczalik (approved):

go/src/koding/klient/machine/machinegroup/aliases/aliases.go
ppknap: no need for limit here, this is quarantined to return unless we have 2^63 − 1 machines. Every limit number will decrease the number of available unique aliases to that limit.

If we have a truly uniform random number generator, this loop will break when `i ≃ (number of machines)`

cihangir: :)

I did not comment just to comment. I have written the similar code 3 years ago, thinking i dont need a limit because it will return, right? then someone added two lines in between my code and we had infinite loop. At least you have test.


> If we have a truly uniform random number generator

Do we have? :)

PR #9792, opened 2016-11-28
klient: add dynamic client

ppknap:
This PR adds dynamic client to `machine` package.

Depends on: #9790 #9791

## Description
This is reworked architecture that is going to handle all `kd machine *` CLI commands. It *partially* deprecates `klient/remote` packages. There is a number of current issues which these changes are intended to solve:

- Support for multiple mounts per machine.
- Correct handling of destroyed machines.
- Correct handling of machines that changed their IP address.
- Correct handling of machines that changed their Kite.key.
- Correct handling of rsync mount processes.
- Clean ups after errors.
- Prevent machine duplication during `kd machine list` 
- Businnes logic and Kite transort separation.
- etc.

## Architecture
image

- **Dynamic Client** - client that can adapt its endpoint address using machine's address book.
- **Disconnected Client** - abstract client factory that describes non existing connection.
- **Kite Client** - Kite client that connects to remote machine Klient.

## How Has This Been Tested?
:warning: WIP - unit tests.

## Screenshots (if appropriate):
none

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #9795, opened 2016-11-28
klient: add dynamic clients group

ppknap:
This PR adds `clients` package that manages machines' dynamic clients.

Depends on: #9792 

## Description

- Clean ups after errors.
- etc.

## Architecture
image

- **Clients** - a set of machine dynamic clients.

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinegroup/clients/clients.go
+++ b/go/src/koding/klient/machine/machinegroup/clients/clients.go
@@ -1,0 +1,181 @@
+package clients
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"koding/klient/machine"
+
+	"github.com/koding/logging"
+)
+
+// ClientsOpts are the options used to configure set of dynamic clients.
+type ClientsOpts struct {
+	// Builder is a factory used to build clients.
+	Builder machine.ClientBuilder
+
+	// DynAddrInterval indicates how often dynamic client should pull address
+	// function looking for new addresses.
+	DynAddrInterval time.Duration
+
+	// PingInterval indicates how often dynamic client should ping external
+	// machine.
+	PingInterval time.Duration
+
+	// Log is used for logging. If nil, default logger will be created.
+	Log logging.Logger
+}
+
+// Valid checks if provided options are correct.
+func (opts *ClientsOpts) Valid() error {
+	if opts.Builder == nil {
+		return errors.New("nil client builder")
+	}
+	if opts.DynAddrInterval == 0 {
+		return errors.New("dynamic address check interval is not set")
+	}
+	if opts.PingInterval == 0 {
+		return errors.New("ping interval is not set")
+	}
+
+	return nil
+}
+
+// Clients is a set of dynamic clients binded to unique machine ID.
+type Clients struct {
+	builder         machine.ClientBuilder
+	dynAddrInterval time.Duration
+	pingInterval    time.Duration
+
+	log logging.Logger
+
+	mu sync.RWMutex
+	m  map[machine.ID]*machine.DynamicClient
+}
+
+// New creates a new Clients object.
+func New(opts *ClientsOpts) (*Clients, error) {
+	if err := opts.Valid(); err != nil {
+		return nil, err
+	}
+
+	c := &Clients{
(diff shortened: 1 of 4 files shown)


go/src/koding/klient/machine/machinegroup/clients/clients.go
```diff
@@ -0,0 +1,181 @@
...
+		c.mu.Unlock()
+		return nil
+	}
+	delete(c.m, id)
+	c.mu.Unlock()
+
+	if dc != nil {
+		dc.Close()
+	} else {
+		c.log.Critical("Non existing client registered under %s ID", id)
```
cihangir: if this is critical you could return ErrMachineNotFound and handle it where required. logging in package functions leads to spamming _sometimes_

cihangir:
is this still wip?

ppknap:
@cihangir yes, when I'm done with payments I will add tests which are missing now

go/src/koding/klient/machine/machinegroup/clients/clients.go
ppknap: this message indicates a programming error which not break anything. The logic ensures that there is always non-nil client. Whether it will be `kite` klient or `disconnected` klient is not important here. If we have nil client, this means that something is wrong with the code itself. So, that Critical log is only a sanity check for us(developers) and is better than `panic` because even if dropped machine doesn't have client, we don't care - it's meant to be dropped anyway.

Dropping of not existing machine is a no-op here so I'm not returning `ErrMachineNotFound`.

cihangir: if it is for developers you should see the error while developing, so you can use panic, but if you say we can see this in the prod only, please add alerting to our logging system and if you get this message, fix the issue.

ppknap:
@cihangir @rjeczalik ping

go/src/koding/klient/machine/machinegroup/clients/clients.go
```diff
@@ -0,0 +1,202 @@
...
+	c := &Clients{
+		builder:         opts.Builder,
+		dynAddrInterval: opts.DynAddrInterval,
+		pingInterval:    opts.PingInterval,
+		log:             opts.Log,
+		m:               make(map[machine.ID]*machine.DynamicClient),
+	}
+
+	if c.log == nil {
+		c.log = logging.NewLogger("")
```
rjeczalik: What about moving this to top-level `var defaultLog = logging.NewLogger("clients")`?

go/src/koding/klient/machine/machinegroup/clients/clients.go
```diff
@@ -0,0 +1,202 @@
...
+	dc, ok := c.m[id]
+	if !ok {
+		c.mu.Unlock()
+		return nil
+	}
+	delete(c.m, id)
+	c.mu.Unlock()
+
+	if dc != nil {
+		dc.Close()
```
rjeczalik: Shouldn't this be called `Stop()` (like `timer.Stop()`), since it does not implement io.Closer interface?

rjeczalik (approved):

go/src/koding/klient/machine/machinegroup/clients/clients.go
ppknap: I followed `httptest.Server` convention where they use `Close()`(which also not implement io.Closer) to stop the server. I think `Stop()` means something recoverable like `time.Timer` and `Close()` something that cannot be reused after the call.

go/src/koding/klient/machine/machinegroup/clients/clients_test.go
```diff
@@ -0,0 +1,71 @@
+package clients
```
ppknap: we don't need to access internal implementation in any of `machine` package-related tests.

rjeczalik (approved):

cihangir (changes requested):
please see my comments in https://github.com/koding/koding/pull/9796

ppknap:
@cihangir done in #9796

cihangir (approved):

PR #9803, opened 2016-11-28
kd: add auth command for registration

mehmetalisavas:
Klientctl auth command and subcommand


go/src/koding/klientctl/auth/register.go
```diff
@@ -0,0 +1,187 @@
...
+		return 1
+	}
+	host := c.GlobalString("host")
+
+	// TODO(mehmetali): make a generalized client to be used in klientctl
+	client := &http.Client{}
+
+	token, err := doRegisterRequest(c, client, host)
+	if err != nil {
+		log.Error(err.Error())
```
rjeczalik: Please use `log.Error("%s", err)` - if err.Error() string contained `%` character, the whole log line will be broken when used like that: `log.Error(err.Error())`.

Also we should print some info to the user, like: `fmt.Fprintln(os.Stderr, "Register failed with error:", err)`. The `log` logs to a file (think of it like diagnostics) - e.g. under macOS it's /Library/Logs/kd.log.

Yes, the logging handling is really verbose in kd currently - we plan to rework it in future by changing cli framework.

Currently we're handling

go/src/koding/klientctl/auth/register.go
```diff
@@ -0,0 +1,206 @@
...
+
+// NewRegisterSubCommand provides the subcommand for registering
+func NewRegisterSubCommand(log logging.Logger) cli.Command {
+	return cli.Command{
+		Name:   "register",
+		Usage:  "Registers the user",
+		Action: ctlcli.ExitAction(RegisterCommand, log, "register"),
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "username",
```
rjeczalik: Probably we'd need an interactive mode for this as well.

rjeczalik:
Would the registration add a user to `kd-io` team and requested team if provided?

/cc @usirin

go/src/koding/klientctl/auth/register.go
mehmetalisavas: @rjeczalik what do you mean with interactive mode? registering with default without `--username name` ?

rjeczalik: >  registering with default without --username name ?

Instead of taking values from flags, ask user on stdin, like `kd credential create`: http://recordit.co/UwA6e8JjS8

mehmetalisavas:
ping

go/src/koding/klientctl/main.go
```diff
@@ -155,6 +156,24 @@ func run(args []string) {
...
+					Name:  "json",
+					Usage: "Output in JSON format",
+				},
+				cli.BoolFlag{
+					Name:  "all",
+					Usage: "Include machines that have been offline for more than 24h.",
+				},
+			},
+			Subcommands: []cli.Command{
+				auth.NewRegisterSubCommand(log),
```
ppknap: could you move this to experimental commands (line 433).

go/src/koding/klientctl/main.go
```diff
@@ -155,6 +156,24 @@ func run(args []string) {
 			},
 		},
 		{
+			Name:  "auth",
```
ppknap: this is already defined in experimental commands (line 431)

go/src/koding/klientctl/auth/register.go
```diff
@@ -0,0 +1,326 @@
...
+		}
+	}
+	if r.Slug == "" {
+		r.Slug, err = helper.Ask("Team Name []: ")
+		if err != nil {
+			return nil, err
+		}
+	}
+	if r.CompanyName == "" {
+		r.CompanyName, err = helper.Ask("Company Name []: ")
```
ppknap: shouldn't this default to r.Slug ?

go/src/koding/klientctl/auth/register.go
```diff
@@ -0,0 +1,326 @@
...
+
+	// team cannot be empty (because of required while registering)
+	// otherwise it return error while registering user
+	// store groupName or slug as "team" inside the cache
+	if err = kloud.Cache().Set("team", c.String("team")); err != nil {
+		log.Error("error while caching team")
+		return 1
+	}
+
+	fmt.Println("This is your session ID, keep it safe.", clientID)
```
ppknap: what about:

```go
fmt.Printf("This is your session ID: %s, keep it safe.\n", clientID)
```

go/src/koding/klientctl/auth/register.go
```diff
@@ -0,0 +1,326 @@
...
+			},
+			cli.StringFlag{
+				Name:  "company",
+				Usage: "Team Company Name to register (default: team name)",
+			},
+			cli.BoolFlag{
+				Name:  "newsletter",
+				Usage: "Do you want to get occasional newsletters from Koding?",
+			},
+			cli.BoolFlag{
```
ppknap: if you use strings to carry these vars, shouldn't this be a string too? because if user specifies --alreadyMember=false it will be asked of this anyway (` r.AlreadyMember == falseValue ` in  `checkAndAskRequiredFields` method)

go/src/koding/kites/config/konfig.go
```diff
@@ -55,6 +55,17 @@ type Konfig struct {
...
 
+// KodingBaseURL specifies the base url for koding
+func (k *Konfig) KodingBaseURL() string {
+	u, err := url.Parse(k.KloudURL)
+	if err != nil {
+		return ""
+	}
+
+	u.Path = ""
+	return u.String()
```
ppknap: could we `return u.Scheme + "://" + u.Host` 

if `k.KloudURL` had query/user/fragment parts, this wouldn't work.

go/src/koding/klientctl/auth/register.go
mehmetalisavas: yeah u are right i know its weird , but while registering, this value is taking this boolean as string. But  in any case, im checking this situation.

go/src/koding/klientctl/auth/register.go
mehmetalisavas: user might not take companyname's slug if its taken. it should be unique. its also optional case. if user doesn't enter company name it would be slug name as default.

go/src/koding/klientctl/auth/register.go
mehmetalisavas: i didn't want to put it between 2 sentences.

rjeczalik: We should not print the session ID at all, it should be used internally only. There's no place user can input it. Instead we cache it per team.

See: https://github.com/koding/koding/blob/master/go/src/koding/klientctl/endpoint/auth/auth.go#L70

go/src/koding/klientctl/auth/register.go
ppknap: oh, I just thought that there should be default value between `[` and `]` characters and should be set to `r.Slug`. If this is not a case, please ignore this comment.

go/src/koding/kites/config/konfig.go
mehmetalisavas: added URL field for this into Konfig

mehmetalisavas:
PTAL @rjeczalik @ppknap

ppknap (approved):

rjeczalik (approved):

Issue #9756, opened 2016-11-25
Filetree is not working if expected home directory doesn't exist.


diclee:
## Expected Behavior
It should at least fallback to '/' directory

## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/20622269/b1ea3542-b30a-11e6-9211-59b405bbdd29.png)

It works properly when top folder is changed:
![image](https://cloud.githubusercontent.com/assets/8261892/20622312/f297ae26-b30a-11e6-8315-52212f8e464c.png)

cc @rjeczalik This is the situation for Marathon provider currently.

rjeczalik:
> This is the situation for Marathon provider currently.

The home directory is assumed to be `/home/user`, however in a container klient always runs as root. Either use `/` or call klient kite with methods `os.currentUsername` + `os.home` to find home for a user that klient is started with.

gokmen:
yup, IDE was designed like that and my old request is coming along; what about providing this info from klient for the current user? like `klient.info` can return;

```json
{
  username: 'foobar'
  home: '/home/foobar'
  groups: ['users', 'sudo', ...]
}
```

wdyt?

rjeczalik:
@gokmen I've extended klient.info according to your suggestion and deployed klient to development channel.

gokmen:
@rjeczalik I'll update IDE to use that information 👍

diclee:
Verify passed on sandbox.

diclee:
Verify failed on latest probably because of klient updates haven't deployed to production channel yet.

diclee:
Verify passed on latest.

PR #9804, opened 2016-11-28
IDE, AppStorage fixes - load improvements

gokmen:
This PR fixes most of the resurrection issues on IDE and it also introduces faster loads with AppStorage improvements. This is a set of required changes to make #9756 happen, I'll continue over this one.

ref: http://recordit.co/hW8hkBkYvX

Marathon provider is using `/root` as home folder based on the `klient.info` response:

### DIFF
--- a/client/ide/lib/index.coffee
+++ b/client/ide/lib/index.coffee
@@ -514,10 +514,8 @@ class IDEAppController extends AppController
     panel     = @workspace.getView()
     filesPane = panel.getPaneByName 'filesPane'
 
-    path = @workspaceData?.rootPath
-
     @workspace.ready ->
-      filesPane.emit 'MachineMountRequested', machineData, path
+      filesPane.emit 'MachineMountRequested', machineData
 
 
   unmountMachine: (machineData) ->
--- a/client/ide/lib/index.coffee
+++ b/client/ide/lib/index.coffee
@@ -509,10 +509,10 @@ class IDEAppController extends AppController
   mountMachine: (machineData) ->
 
     # interrupt if workspace was changed
-    return if machineData.uid isnt @workspaceData.machineUId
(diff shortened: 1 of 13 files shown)


gokmen:
I've reverted https://github.com/koding/koding/pull/9587/commits/fe820824e965014002c1deacf19bb327eca80bbd since the main issue in AppStorage is fixed in this PR cc/ @hakankaradis

gokmen:
I'm done with this one but please don't merge it even it's ok, there are some experimental changes that might affect some place else that I didn't tested and verified on my own. We can merge this in once it's ready after the upcoming deploy to production. cc/ @sinan @szkl

rjeczalik:
Will we have backward compatibility with old klient.info responses? I'm thinking of old managed klient which do not have autoupdates enabled.

gokmen:
@rjeczalik I covered the case of error here https://github.com/koding/koding/pull/9804/files#diff-52d8d249234cc4460894a6fb44d36fbbR172 but not checked `info` result, good catch let me update.

client/ide/lib/index.coffee
```diff
@@ -549,52 +533,46 @@ class IDEAppController extends AppController
 
   createInitialView: (withFakeViews) ->
```
sinan: Seems like this method became too big. Let's separate `if/else` section into separate methods.

gokmen: it's becoming shorter actually it was https://github.com/koding/koding/blob/master/client/ide/lib/index.coffee#L550-L597 (47 lines) but now https://github.com/koding/koding/blob/d8c0fd4522a2e8223437930943602aa8f110ab69/client/ide/lib/index.coffee#L534-L575 (41 lines) :) but I can split it out of course.

sinan (approved):

PR #9750, opened 2016-11-25
Landing stripe

usirin:
/cc @devrim @cihangir

PR #9820, opened 2016-11-29
Add credit card requirement to Signup Flow

usirin:
Just for the signup.

/cc @gokmen: right now it's not disabled for `default` or `dev` environments yet, still not sure how to implement it correctly. So i might defer it into another PR, fyi.

Continuation of #9750


client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -228,6 +273,9 @@ module.exports = utils = {
 
   createTeam: (callbacks = {}) ->
 
+    unless token = utils.getPayment()?.token
+      new kd.NotificationView { title : 'You need to enter your credit card!' }
```
sinan: shouldn't it return error here?

sinan (approved):
Looks good to me! just one minor comment, can also be major, pls check :)

PR #9851, opened 2016-12-01
api: add Client / Transport for clientID-authentication

rjeczalik:
~~Depends on: https://github.com/koding/koding/pull/9786~~


go/src/koding/kites/kloud/kloud/kloud.go
```diff
@@ -235,12 +237,46 @@ func New(conf *Config) (*Kloud, error) {
...
+		return &socialapi.Session{
+			Username: s.Username,
+			Team:     s.GroupName,
+			ClientID: s.ClientId,
+		}, nil
+	}
+
+	transport := &socialapi.Transport{
+		RoundTripper: storeOpts.Client.Transport,
+		Host:         remoteURL.Host,
```
ppknap: this will `panic` if remoteURL is nil (you're checking that case in line 267)

go/src/koding/socialapi/transport.go
```diff
@@ -0,0 +1,320 @@
...
+// requests with the given session.
+//
+// Session is expected to be have a valid clientID. In order to be
+// able to refresh the session if it gets invalidated, the session
+// must also contain Username / Team information.
+//
+// This is usefull for api clients that give no control over HTTP requests,
+// like the swagger-generated ones (remote.api).
+func (t *Transport) NewSingleClient(session *Session) http.RoundTripper {
+	return &DirectorTransport{
```
ppknap: Hmm don't know the current status but this may interest you(first 40secs): https://youtu.be/FARQMJndUn0?t=1151

rjeczalik: The point is no-one should type-assert on a concrete type, rather do interface upgrade instead.

go/src/koding/kites/kloud/kloud/kloud.go
rjeczalik: Done, thanks!

go/src/koding/socialapi/transport.go
ppknap: This is only fyi that DefaultTransport cannot be replaced. Nothing here to fix.

go/src/koding/socialapi/cache_test.go
```diff
@@ -0,0 +1,128 @@
...
+	var auth = NewFakeAuth()
+
+	users := []*socialapi.Session{
+		{Username: "user1", Team: "foobar"},
+		{Username: "user2", Team: "foobar"},
+		{Username: "user3", Team: "foobar"},
+		{Username: "user", Team: "team"},
+	}
+
+	cases := []struct {
```
ppknap: could you change it to `map[string]struct { .... }` ?

go/src/koding/socialapi/transport.go
```diff
@@ -0,0 +1,329 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+const maxBodyLen = 2 << 20 // 2 MiB
```
ppknap: :+1:

go/src/koding/socialapi/transport.go
```diff
@@ -0,0 +1,329 @@
...
+	Team string
+}
+
+// Key is used for SessionCache to cache sessions by keys.
+func (s *Session) Key() string {
+	return s.Team + "/" + s.Username
+}
+
+// Valid implements the stack.Validator interface.
+func (s *Session) Valid() error {
```
ppknap: do we allow empty teams and usernames? If not, could you add them to validator too?

ppknap (approved):

go/src/koding/socialapi/transport.go
rjeczalik: Yes, we currently do.

go/src/koding/kites/kloud/kloud/kloud.go
```diff
@@ -30,6 +31,7 @@ import (
 	"koding/kites/kloud/terraformer"
 	"koding/kites/kloud/userdata"
 	"koding/remoteapi"
+	"koding/socialapi"
```
cihangir: can we rename socialapi to api or clients? socialapi is one of the service name and we have the same package path?

go/src/koding/kites/kloud/kloud/kloud.go
```diff
@@ -235,12 +237,45 @@ func New(conf *Config) (*Kloud, error) {
...
+	if remoteURL != nil {
+		transport.Host = remoteURL.Host
+
+		localRemoteURL := *remoteURL
+
+		// TODO(rjeczalik): Add PrivateRemoteURL to koding/config.
+		if _, port, err := net.SplitHostPort(localRemoteURL.Host); err == nil {
+			localRemoteURL.Host = net.JoinHostPort("127.0.0.1", port)
+		} else {
+			localRemoteURL.Host = "127.0.0.1"
```
cihangir: lets remove the hardcoded configurations and read them from config

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+	// new session with Refresh set to true.
+	AuthFunc AuthFunc
+
+	// Storage is an external storage for storing session.
+	//
+	// If nil, in-memory map is going to be used instead.
+	Storage Storage
+
+	cacheMu sync.RWMutex
+	cache   map[string]*Session
```
cihangir: if we have Storage what is the use of cache here? lets use kodig/cache whenever possible

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+//
+//   cache := socialapi.NewSessionCache(authFn)
+//
+//   t := &socialapi.Transport{
+//       AuthFunc: cache.Auth,
+//   }
+//
+func (s *SessionCache) Auth(opts *AuthOptions) (*Session, error) {
+	if s.AuthFunc == nil {
+		panic("socialapi: AuthFunc is nil")
```
cihangir: this is not socialapi :)

go/src/koding/socialapi/cache.go
cihangir: you can implement MultiCache which fallbacks to another layer if required

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+
+	session := opts.Session
+
+	// Early return - return existing session if it's valid
+	// and we were not ask to invalidate it.
+	if err := session.Valid(); err == nil && !opts.Refresh {
+		return session, nil
+	}
+
+	if session == nil {
```
cihangir: better to check session before validating it

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, errors.New("cannot determine user session")
+	}
+
+	if !opts.Refresh {
```
cihangir: could you seperate refreshing logic into another function?

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+	//   for the next request
+	//
+	// System-wise Kloud uses in-memory cache while KD and Klient
+	// use BoltDB-backed cache. Since the latter two take exclusive
+	// lock of the database file, the only reason storage can
+	// fail are filesystem errors - there is no recovery from
+	// that, and falling back to memory makes sense only for
+	// Klient.
+	s.cacheMu.Lock()
+	if opts.Refresh {
```
cihangir: you already checked if this is a refresh operation on line 87

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if err == nil {
+		_ = s.set(session)
+	}
+	s.cacheMu.Unlock()
+
+	return session, err
+}
+
+func (s *SessionCache) get(session *Session) error {
```
cihangir: "get" is a misleading word for this function, what about something like populate/ensure?

go/src/koding/socialapi/cache.go
```diff
@@ -0,0 +1,165 @@
...
+// when requested session was not found.
+var ErrSessionNotFound = errors.New("session not found")
+
+// Storage is an interface for external session storage.
+type Storage interface {
+	// Get gets a session.
+	//
+	// It returns ErrSessionNotFound if requested
+	// session was not cached.
+	Get(*Session) error
```
cihangir: Get should return the session

go/src/koding/socialapi/helper_test.go
```diff
@@ -0,0 +1,232 @@
...
+	for _, s := range fa.Sessions {
+		if s.ClientID == req.ClientID {
+			session = s
+			break
+		}
+	}
+	fa.mu.RUnlock()
+
+	p, err := ioutil.ReadAll(r.Body)
+	log.Printf("FakeAuth.GetSession: read body: p=%q, err=%v", p, err)
```
cihangir: could you use our logger or remove this?

go/src/koding/socialapi/helper_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+	"sync"
+
+	"koding/kites/kloud/utils"
+	"koding/socialapi"
+)
+
+type FakeAuth struct {
```
cihangir: instead of mocking could we use the real code to test?

go/src/koding/socialapi/helper_test.go
```diff
@@ -0,0 +1,232 @@
...
+type contextKey struct {
+	name string
+}
+
+var (
+	ErrorContextKey    = &contextKey{"fake-error"}
+	ResponseContextKey = &contextKey{"fake-response"}
+)
+
+type FakeTransport struct {
```
cihangir: instead of the fake transport lets use the real one

go/src/koding/socialapi/presence/presence.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"errors"
+	"net/http"
+	"net/url"
+
+	"koding/socialapi"
+)
+
+type Client struct {
```
cihangir: could you use the provided client in the socialapi/workers/presence/client here https://github.com/koding/koding/pull/9863/files#diff-9da7fe549db0224bc2de4a603ce84d51R47 (after you merge :D )

go/src/koding/socialapi/transport_test.go
```diff
@@ -0,0 +1,142 @@
...
+		},
+	}}
+
+	rec := &AuthRecorder{
+		AuthFunc: cache.Auth,
+	}
+
+	client := http.Client{
+		Transport: &socialapi.Transport{
+			RoundTripper: &FakeTransport{},
```
cihangir: could you use the real transport here?

go/src/koding/socialapi/transport.go
```diff
@@ -0,0 +1,329 @@
...
+	return &DirectorTransport{
+		RoundTripper: t,
+		Director: func(req *http.Request) {
+			*req = *session.WithRequest(req)
+		},
+	}
+}
+
+// RoundTrip implements the http.RoundTripper interface.
+func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
```
cihangir: not checking this function for now since may need to refactor to be able to use in your tests

go/src/koding/socialapi/transport_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"net/http"
+	"strings"
+	"testing"
+
+	"koding/kites/tunnelproxy/discover/discovertest"
+	"koding/socialapi"
+)
+
+func TestTransport(t *testing.T) {
+	var storage TrxStorage
```
cihangir: could you use the real storage here?

cihangir (changes requested):
instead of mocking, lets use the the real functionality in the tests

go/src/koding/socialapi/transport_test.go
rjeczalik: No, I can't. The real transport is wrapped with FakeTransport. FakeTransport is used to fake transport errors.

go/src/koding/socialapi/presence/presence.go
rjeczalik: Not without refactoring it. I need to have access to *http.Request. I can refactor it thought to allow for that. For now the requests are internally created, I order to request authentication I'd need to have access to requests' context:

https://github.com/koding/koding/pull/9863/files#diff-9da7fe549db0224bc2de4a603ce84d51R63

go/src/koding/socialapi/transport_test.go
rjeczalik: No, I can't. The TrxStorage is kind of recording storage. It records all calls to Set / Get / Delete functions and then we verify in tests.

go/src/koding/socialapi/transport_test.go
cihangir: i believe we can use after a bit of refactoring. 

As we talked previously and all agreed on, instead of mocking lets use the real functionality in our tests.

go/src/koding/socialapi/transport.go
cihangir: btw i have alreday checked but not commenting

go/src/koding/socialapi/cache.go
rjeczalik: > lets use kodig/cache whenever possible

The `cache` is default memory cache if Storage is nil. I avoided koding/cache (the `Memory` one) since I did not like it implementation [0] - moreover the implementation is so trivial (map + mutex) that I thought it does not varrant pulling in external dependency. 

[0] should have sync.RWMutex instead - https://github.com/koding/cache/blob/master/memory.go

go/src/koding/socialapi/cache.go
rjeczalik: Yes, and here I check it again under critical section in order to update storage.

go/src/koding/socialapi/helper_test.go
rjeczalik: Which real code?

go/src/koding/socialapi/transport_test.go
rjeczalik: How do you want to test for transport errors using real functionality?

go/src/koding/kites/kloud/kloud/kloud.go
rjeczalik: Done in https://github.com/koding/koding/pull/9870/commits/4ff0b9494a13d1dfaf3680748d4209c044bc6646.

Renamed to`api`. Note for future - we're going to merge the following packages:

1) `koding/klientctl/endpoint`
2) `koding/remoteapi`

Under `koding/api` in future. For now we're not done with 1) API-wise so we do not see yet how would the final result look like. Just fyi.

go/src/koding/kites/kloud/kloud/kloud.go
rjeczalik: Done in https://github.com/koding/koding/pull/9870/commits/5e4f7a6c88f28b9bea1222e204ffeb719e410ebb.

go/src/koding/socialapi/helper_test.go
rjeczalik: Done in 3b9b8a9bc2dfbe425b675c07bfb75f67d42e5241.

go/src/koding/socialapi/cache.go
rjeczalik: Done - `get` is removed by a1ff5b919a8d5a2c32d7b6ceb98a666297ce7677.

go/src/koding/socialapi/cache.go
rjeczalik: Done in dd3d5b9edacdcd39fc8d5a2bc7f1209192869ce4.

go/src/koding/socialapi/cache.go
rjeczalik: Done in 6305645866dbda08d850ca8fcb5b24a07d86963d.

go/src/koding/socialapi/cache.go
rjeczalik: Done in 0d10ce61f121760fcd4ce0a5b02441b8961967a5.

rjeczalik:
@cihangir Addressed all except comments about FakeAuth and FakeTransport.

- there's no real replacement for FakeAuth - in production we use 1a) `auth.login` kite method or 1b) `modelhelper.FetchOrCreateSession` - using 1a) is not possible as there's no mean to start koding as a part of this unittest; using 1b) would not change much in my opinion, as the only thing that changes is the storage - mongo instead of memory, and `api` package responsibility is not to test mongo

- if we want to get rid of FakeTransport, we also get rid of [this](https://github.com/koding/koding/pull/9851/files#diff-c519253ffa5abcd1f4885d8a72242b96R66) and [this](https://github.com/koding/koding/pull/9851/files#diff-c519253ffa5abcd1f4885d8a72242b96R74) test case.

How to proceed?

ppknap:
@rjeczalik I wrote a simple [test case](https://gist.github.com/ppknap/6e2974b7999901be596953e51b09d901) for cache type. Is it valid? If yes, this means that the cache itself or test fixtures have [data-races](https://gist.github.com/ppknap/b708b1af302238d5132888affd630a52).

rjeczalik:
> If yes, this means that the cache itself or test fixtures have data-races

@ppknap Thanks.

I've included your test and a fix in https://github.com/koding/koding/pull/9870/commits/8631e946e5a686b401576558c6154ac6b8e84380.

go/src/koding/kites/kloud/kloud/kloud.go
cihangir: thanks

go/src/koding/socialapi/cache.go
cihangir: i believe you wrote these in frustration :) the Rafal i know does not complain instead fixes them whenever he see a problem. 

Other than that i would strongly suggest you to use LRU instead of simple memory cache, since the talk all started from leaking memory not the trivial implementation itself. In the previous-previous talk we both were OK with a simple map

rjeczalik: > Other than that i would strongly suggest you to use LRU instead of simple memory cache, since the talk all started from leaking memory not the trivial implementation itself.

Sure, but let's do this in a separate PR, can we?

go/src/koding/socialapi/cache.go
rjeczalik: The method was reworked - is there still anything do to here?

go/src/koding/socialapi/helper_test.go
cihangir: socialapi.NewCache

go/src/koding/socialapi/presence/presence.go
cihangir: why do you need to refactor? this package has the exact same functionality. Why do you need to access to http.Request?

go/src/koding/socialapi/transport_test.go
cihangir: i believe you can. it is same implementation with an inmemory storage with an GetAll(maybe?) function, and it would eliminate some of the code

go/src/koding/socialapi/presence/presence.go
rjeczalik: To add api.SessionContextKey to http.Request.Context here:

https://github.com/koding/koding/pull/9851/files#diff-64803beca13535937999dd1daec4743cR22

This way single transport knows for which user it should get the ClientID.

go/src/koding/socialapi/transport_test.go
rjeczalik: Done in https://github.com/koding/koding/commit/28fc46282c43fa1c87fbec1908bd19894d652cdc.

go/src/koding/socialapi/cache.go
cihangir: If we are going to start using this right after we merge, lets implement, otherwise we can do later

go/src/koding/socialapi/transport_test.go
cihangir: i was thinking smth similar to these https://github.com/golang/go/blob/master/src/net/http/transport.go#L1811-L1821

cihangir: I have checked the code test coverage, it is in good shape, it can stay as is. 
on the other i couldnt see the integration of these into wercker, did i miss it or not done yet?

go/src/koding/socialapi/presence/presence.go
cihangir: ok i see the problem, you dont have access to `clientId` since it is in the transport cache/auther.

on the other hand NewRequest is exported https://github.com/koding/koding/pull/9863/files#diff-9da7fe549db0224bc2de4a603ce84d51R72 so you have acccess to the http.request itself

go/src/koding/socialapi/transport_test.go
rjeczalik: Done in https://github.com/koding/koding/commit/bb29400bcb4a23c9c2b1a863372ec4208d8d636d.

go/src/koding/socialapi/cache.go
rjeczalik: Done in https://github.com/koding/koding/commit/6d0d33f7d7fab14a8bf0ee274df6e8ab78f12344.

rjeczalik:
@cihangir all addressed, ptal

go/src/koding/socialapi/transport_test.go
cihangir: > they are not faking or mocking any functionality.

which does not mean you cant, you can think them as small adapters and do what you want to do

for the hook debate (since you said it makes it worse), i find easier to reason about the cases/logic since they are in the code itself, not separated into another layer. i dont support having many hooks in the code itself too but i am also against having a big function and to be able to test adding another level of complexity with trying to fake the logical parts of it, which incidentally may lead to bugs in the tests itself(another question comes as how/should we (are going to) test the tests?). I would probably go with unit tests and smaller functions instead of having bigger components and mocks. 

in the end everyone has their own taste and there is no universal truth, at least i know of.

cihangir (approved):

cihangir:
Thanks @rjeczalik

cihangir:
for the followers of this thread, we decided not to have big PRs from now on, which causes lengthy discussions and making things harder to reason about while reviewing, following up with the comments and their resolutions.

go/src/koding/socialapi/presence/presence.go
cihangir: > I don't see a clean way to use the client.

yep i stated that as a problem in my previous comment

Issue #9689, opened 2016-11-23
kd: add team subcommand


rjeczalik:
Adds support for kd to list and login to teams. Adds support for generic kd-io team.

TODO

- [x] kloud: add backend for `team list`
- [x] kd: add frontend for `team list`
- [x] kd: add frontend for `team use`

Implements

`kd auth login --team`
`kd team list`
`kd team use`

usirin:
@ppknap What is the status update with this issue? Does the closed PR (#9852) implement all the requirements?

rjeczalik:
@usirin Yes, #9852 added `team list`, I've added `auth login --team` in a separate PR. Can be closed.

Sorry, there's a mess currently in the backlog, we're going to clean it up.

usirin:
> Sorry, there's a mess currently in the backlog, we're going to clean it up.

👍 That would be amazing.

PR #9852, opened 2016-12-01
Add KD team list subcommand

ppknap:
This PR adds `kd team list` subcommand.

Part of: #9689
Depends on: #9828

```
$ ./klientctl team list
NAME    SLUG          PRIVACY  MEMBERS  SUBSCRIPTION
google  google        private  1        trialing
koding  koding-pawel  private  1        trialing
$ ./klientctl team list --json
[
	{
		"name": "google",
		"slug": "google",
		"members": "1",
		"privacy": "private",
		"subscription": "trialing"
	},
	{
		"name": "koding",
		"slug": "koding-pawel",
		"members": "1",
		"privacy": "private",
		"subscription": "trialing"
	}
]
$ ./klientctl team list --slug=google
NAME    SLUG    PRIVACY  MEMBERS  SUBSCRIPTION
google  google  private  1        trialing
$ ./klientctl team list --slug=none
Cannot find "none" team.
```

### DIFF
--- a/go/src/koding/db/mongodb/modelhelper/roles.go
+++ b/go/src/koding/db/mongodb/modelhelper/roles.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"koding/db/models"
 
+	mgo "gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 )
 
@@ -103,6 +104,10 @@ func FetchAccountGroupNames(username string) ([]string, error) {
 		return nil, err
 	}
 
+	if len(rels) == 0 {
+		return nil, mgo.ErrNotFound
+	}
+
 	var ids []string
 	for _, rel := range rels {
 		ids = append(ids, rel.SourceId.Hex())
@@ -144,6 +149,10 @@ func FetchAccountGroups(username string) (groups []*models.Group, err error) {
 	rels, err := GetAllRelationships(selector)
 	if err != nil {
 		return nil, err
+	}
+
+	if len(rels) == 0 {
+		return nil, mgo.ErrNotFound
 	}
 
 	var ids []bson.ObjectId
--- a/go/src/koding/db/mongodb/modelhelper/roles.go
+++ b/go/src/koding/db/mongodb/modelhelper/roles.go
@@ -84,8 +84,8 @@ func HasAnyRole(username, groupName string, roles ...string) (bool, error) {
 	return count > 0, nil
 }
 
-// FetchAccountGroups lists the group memberships of a given username
-func FetchAccountGroups(username string) ([]string, error) {
+// FetchAccountGroupNames lists the group memberships of a given username
+func FetchAccountGroupNames(username string) ([]string, error) {
 	account, err := GetAccount(username)
 	if err != nil {
 		return nil, err
@@ -129,4 +129,52 @@ func FetchAccountGroups(username string) ([]string, error) {
 	}
 
 	return slugList, nil
+}
+
+// FetchAccountGroups lists the groups of a given username
+func FetchAccountGroups(username string) (groups []*models.Group, err error) {
+	account, err := GetAccount(username)
+	if err != nil {
+		return nil, err
+	}
+
+	selector := Selector{
+		"sourceName": "JGroup",
+		"targetId":   account.Id,
+		"targetName": "JAccount",
+		"as":         bson.M{"$in": []string{"owner", "admin", "member"}},
+	}
+
+	rels, err := GetAllRelationships(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(rels) == 0 {
+		return nil, nil
+	}
+
+	var ids []bson.ObjectId
+	for _, rel := range rels {
+		ids = append(ids, rel.SourceId)
+	}
+
+	all, err := GetGroupsByIds(ids...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Unify the list.
+	slugs := make(map[string]struct{})
+	for _, group := range all {
+		// Skip already added groups.
+		if _, ok := slugs[group.Slug]; ok {
+			continue
+		}
+
+		groups = append(groups, group)
+		slugs[group.Slug] = struct{}{}
+	}
+
+	return groups, nil
 }
--- a/go/src/koding/db/mongodb/modelhelper/roles.go
+++ b/go/src/koding/db/mongodb/modelhelper/roles.go
@@ -103,10 +103,6 @@ func FetchAccountGroupNames(username string) ([]string, error) {
 		return nil, err
 	}
 
-	if len(rels) == 0 {
-		return nil, nil
-	}
-
 	var ids []string
 	for _, rel := range rels {
 		ids = append(ids, rel.SourceId.Hex())
@@ -148,10 +144,6 @@ func FetchAccountGroups(username string) (groups []*models.Group, err error) {
 	rels, err := GetAllRelationships(selector)
 	if err != nil {
 		return nil, err
-	}
-
-	if len(rels) == 0 {
-		return nil, nil
 	}
 
 	var ids []bson.ObjectId
--- a/go/src/koding/kites/kloud/team/mongo.go
+++ b/go/src/koding/kites/kloud/team/mongo.go
@@ -7,6 +7,7 @@ import (
 	"koding/db/models"
 	"koding/db/mongodb/modelhelper"
 
+	mgo "gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 )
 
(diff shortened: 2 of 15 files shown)


go/src/koding/kites/kloud/team/mongo.go
```diff
@@ -0,0 +1,132 @@
...
+// fetchOne returns only specified team.
+func (m *MongoDatabase) fetchOne(accID bson.ObjectId, user, slug string) ([]*Team, error) {
+	groupDB, err := m.adapter.GetGroup(slug)
+	if err == mgo.ErrNotFound {
+		return []*Team{}, nil
+	} else if err != nil {
+		return nil, models.ResError(err, modelhelper.GroupsCollectionName)
+	}
+
+	belongs := modelhelper.Selector{
```
cihangir: newly introduced modelhelper.IsParticipant is your friend 🙋‍♂️

go/src/koding/kites/kloud/team/mongo.go
```diff
@@ -0,0 +1,132 @@
...
+
+func groups2teams(groups ...*models.Group) []*Team {
+	teams := make([]*Team, 0, len(groups))
+	for _, group := range groups {
+		if group == nil || group.Slug == "koding" {
+			continue
+		}
+
+		// Filter teams that have empty subscription.
+		if group.Payment.Subscription.ID == "" {
```
cihangir: use group.IsSubActive()

cihangir (changes requested):
modelhelper changes

go/src/koding/kites/kloud/team/mongo.go
ppknap: I'm not sure the behavior will be the same, I want to have all subscriptions even these that are not paid etc. This logic is to filter groups like `ide`. Am I missing something?

go/src/koding/kites/kloud/team/mongo.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return nil, models.ResError(err, modelhelper.RelationshipColl)
+	}
+
+	return groups2teams(groupDB), nil
+}
+
+// fetchOne returns all user's teams.
+func (m *MongoDatabase) fetchAll(accID bson.ObjectId) ([]*Team, error) {
+	belongs := modelhelper.Selector{
```
ppknap: so many friends in that package:)

go/src/koding/kites/kloud/team/mongo.go
cihangir: hm, i think i didnt share the whole logic here, i thougth you are filtering the not-paid team. because if a team does not pay, their subscription is cancelled so we remove the ID, that is why i suggested  `IsSubActive`, if you want to show all the team, we need to remove this check then

go/src/koding/kites/kloud/team/mongo.go
cihangir: most important reason i commented on this is, we dont have an index on these 3 fields. second but not that important one is, lets have these kind of logic in modelhelper

go/src/koding/kites/kloud/team/mongo.go
ppknap: We don't have to, I can check `grup.Payment.Subscription.Status != ""`. I just don't want to show `""` subscriptions so maybe checking Status instead of ID will be even better.

go/src/koding/kites/kloud/team/mongo.go
ppknap: agree, more db-related-things in modelhelper == win for us in the future.

go/src/koding/db/mongodb/modelhelper/roles.go
```diff
@@ -130,3 +130,51 @@ func FetchAccountGroups(username string) ([]string, error) {
...
+		"as":         bson.M{"$in": []string{"owner", "admin", "member"}},
+	}
+
+	rels, err := GetAllRelationships(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(rels) == 0 {
+		return nil, nil
```
rjeczalik: `return nil, mgo.ErrNotFound`?

Currently we signal lack of values found by sentinel error.

go/src/koding/kites/kloud/team/mongo.go
```diff
@@ -0,0 +1,101 @@
...
+// fetchOne returns only specified team.
+func (m *MongoDatabase) fetchOne(user, slug string) ([]*Team, error) {
+	group, err := m.adapter.GetGroup(slug)
+	if err == mgo.ErrNotFound {
+		return []*Team{}, nil
+	} else if err != nil {
+		return nil, models.ResError(err, modelhelper.GroupsCollectionName)
+	}
+
+	switch participand, err := m.adapter.IsParticipant(user, slug); {
```
rjeczalik: participant

go/src/koding/klientctl/endpoint/team/team.go
```diff
@@ -40,6 +47,23 @@ func (c *Client) Used() *Team {
...
+func (c *Client) List(opts *ListOptions) ([]*team.Team, error) {
+	c.init()
+
+	req := &stack.TeamListRequest{}
+	if opts != nil {
+		req.Slug = opts.Slug
+	}
+
+	resp := &stack.TeamListResponse{}
+	if err := c.kloud().Call("team.list", req, resp); err != nil {
```
rjeczalik: Out of curiosity, question about `var resp stack.TeamListResponse` over `resp := &stack.TeamListResponse{}` - why not:

```
var resp stack.TeamListResponse

...

if err := c.kloud().Call("team.list", req, &resp); err != nil {
...
```

This way the code is more readable imho, you see what var is being written to.

go/src/koding/klientctl/team.go
```diff
@@ -3,15 +3,63 @@ package main
...
+	opts := &epteam.ListOptions{
+		Slug: c.String("slug"),
+	}
+
+	teams, err := epteam.List(opts)
+	if err != nil {
+		return 0, err
+	}
+
+	if len(teams) == 0 {
```
rjeczalik: List should return some `NotFound` error, instead of both arguments nil. The caller then may want to choose whether ignore such error or not.

go/src/koding/klientctl/endpoint/team/team.go
ppknap: Will change this to:

```
resp := stack.TeamListResponse{}
if err := c.kloud().Call("team.list", req, &resp); err != nil {
```

`var` vs `:=` is rather a personal preference here.

go/src/koding/klientctl/team.go
ppknap: It doesn't return both arguments nil (see [here](https://github.com/koding/koding/pull/9852/files#diff-725382ef8210b3747b2451c5ce354d47R79) ). This mean that we get error only on real failures. Empty list is not an error here it's just a normal result from kloud call. It's Kloud that handles NotFound errors from db and not KD.

rjeczalik: As you wish, but this is opposite what we have now currently. I tried to return various ErrNotFound errors when things to return were not found.

ppknap: No, It's not opposite to what we have now. Eg. you handle NotFound errors in `credential.list` on [Kloud side](https://github.com/koding/koding/blob/bd496ac0a9c293791db54bbb89576bef6969fe5a/go/src/koding/kites/kloud/credential/mongo.go#L393). Thus, you return empty(non nil slice) and "ignores" NotFound error in Kloud. This is exactly what I'm doing here.

go/src/koding/db/mongodb/modelhelper/roles.go
ppknap: I'm removing this entirely since `GetAllRelationships` function above will return `mgo.ErrNotFound` in case of empty result, so this check is redundant.

rjeczalik:
```
$ scripts/wercker/run-command ./run exec go test koding/db/mongodb/modelhelper
--- FAIL: TestFetchAccountGroups (0.00s)
	roles_test.go:245: no group id specified
```

rjeczalik (changes requested):
```
$ scripts/wercker/run-command ./run exec go test koding/db/mongodb/modelhelper
--- FAIL: TestFetchAccountGroups (0.00s)
	roles_test.go:245: no group id specified
```

ppknap:
@rjeczalik fixed

rjeczalik (approved):

ppknap:
@cihangir ping

go/src/koding/db/mongodb/modelhelper/roles.go
```diff
@@ -84,8 +85,8 @@ func HasAnyRole(username, groupName string, roles ...string) (bool, error) {
 	return count > 0, nil
 }
 
-// FetchAccountGroups lists the group memberships of a given username
-func FetchAccountGroups(username string) ([]string, error) {
+// FetchAccountGroupNames lists the group memberships of a given username
+func FetchAccountGroupNames(username string) ([]string, error) {
```
cihangir: s/FetchAccountGroupNames/FetchAccountGroupSlugs

go/src/koding/db/mongodb/modelhelper/roles.go
```diff
@@ -130,3 +131,51 @@ func FetchAccountGroups(username string) ([]string, error) {
 
 	return slugList, nil
 }
+
+// FetchAccountGroups lists the groups of a given username
+func FetchAccountGroups(username string) (groups []*models.Group, err error) {
```
cihangir: you could share the part to fetch the ids of groups that we have in FetchAccountGroupNames

cihangir:
@ppknap i made some comments but they are not critical, you can fix them in some of the next PRs of you

cihangir (approved):

PR #9853, opened 2016-12-01
socialapi/payment: get rid of from headache completely

cihangir:
charge events are not used but caused many problems :/


go/src/socialapi/workers/payment/webhook_test.go
```diff
@@ -39,104 +39,6 @@ func withStubData(f func(username string, groupName string, sessionID string)) {
...
 }
 
-func TestChargeSuccededHandler(t *testing.T) {
-	testData := `
-{
-    "id": "ch_00000000000000",
-    "object": "charge",
-    "amount": 100,
-    "currency": "usd",
-    "customer": "%s",
```
cihangir: here i question the tests, even if you have tests for most of the things, things still fail. Our data has customer field, and also app expects them, but stripe does not send them for non-registered user's operations. Never trust the internet.

rjeczalik (approved):

PR #9860, opened 2016-12-01
landing: set env config var from main domain url

### LINKED ISSUE #9833: Auto-filled credit card is not working on latest and a new card cannot be entered.
## Current Behavior
Therefore team cannot be created on latest.
gif
http://recordit.co/G7YaSOCB3E

usirin:
Right now it's being read from `location.pathname`, which causes `latest.koding.com` to think it has `development` as its `kd.config.environment` variable.

This PR addresses this issue by using the `main` domain from `kd.config.domains`.

Fixes #9833 

/cc @cihangir @sinan

### DIFF
--- a/client/landing/site.landing/coffee/main.coffee
+++ b/client/landing/site.landing/coffee/main.coffee
@@ -37,7 +37,7 @@ do ->
     new MainController group
 
   kd.config             or= {}
-  kd.config.environment   = if location.hostname is 'koding.com' then 'production' else 'development'
+  kd.config.environment   = window._runtimeOptions.environment
   kd.config.recaptcha     = window._runtimeOptions.recaptcha
   kd.config.google        = window._runtimeOptions.google
   kd.config.domains       = window._runtimeOptions.domains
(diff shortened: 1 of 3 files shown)


szkl:
Let's put in KONFIG.config and KONFIG.environment into client runtime options passed to landing to avoid location hostname conditional expressions.

usirin:
> Let's put in KONFIG.config and KONFIG.environment into client runtime options passed to landing to avoid location hostname conditional expressions.

👍

szkl (approved):

client/landing/site.landing/coffee/main.coffee
```diff
@@ -37,7 +37,7 @@ do ->
     new MainController group
 
   kd.config             or= {}
-  kd.config.environment   = if location.hostname is 'koding.com' then 'production' else 'development'
+  kd.config.environment   = window._runtimeOptions.environment
```
sinan: we should either go safe and place a landmine such as
`kd.config.environment  = window._runtimeOptions.environment or 'development'`
or we do `throw  unless kd.config.environment` to be safer since this is an important piece.

szkl: It is [defined](https://github.com/koding/koding/blob/master/config/main.dev.coffee#L24) as `dev`.

szkl: @sinan Is it possible to have landing loaded without `window._runtimeOptions` set in the page?

sinan: anything is possible :)

szkl: Only reference to `development` is going to be removed with this pull request. `kd.config.environment` will be compared to `production` only. It sounds good to me.

Issue #9891, opened 2016-12-05
Fix Rainforest QA test specifications for $5 / step


szkl:
0. Fork https://github.com/koding/koding repository
1. Ask for a Rainforest QA account
2. Pick a failed test in results [page](https://app.rainforestqa.com/runs)
3. Open the related test file in [this](https://github.com/koding/koding/tree/master/tests/spec/rainforest) directory. Test file has the same name as the title of the test 
3a. Click to "Go to test" link on top right corner
4. Click to "Preview Test" button
5. Select "Sandbox (default)" environment
6. Select "Google Chrome - OS X" browser
7. Click to "Open Preview" button
8. Click to "I Understand" button at the bottom
9. Follow the steps
10. Make the changes required to fix test specification when you see a mistake or test is not written properly. Make sure test step is written in simple and proper English
11. Continue testing remaining steps in the same test
12. Send a pull request to master branch in koding/koding repository when you're sure you have fixed all the errors
13. Wait until Rainforest test run is completed. Check the result of the test file that you have fixed
14. If all tests have passed then contact your supervisor. You'll receive $5 per step you have fixed. A single test file contains approximately 30 steps

Supervisors for this issue are @szkl @diclee @hakankaradis. Please contact a supervisor if you have any questions.

Comment on this issue regarding which test you are going to work on and your deadline estimation. This is necessary to avoid conflicts with multiple people working on the same test specification.

If you'd like to join this effort ask for a Rainforest account from one of the supervisors.

kwagdy:
finished: `sharedVMs_uploadFileByDragDrop.rfml`

kwagdy:
finished: `collaboration_reject_accept_invitation.rfml`

kwagdy:
started on: `sharedVMs_reject_inviteAgain_accept_checkTerminalActions`...

hectork01:
Working oin `skipGuide_custom_userData_variables_readme_templatePreviewClick`
Fixing instructions when looking for the aws key.

hectork01:
Working on: `createTeam_usingAlreadyRegisteredEmail_afterClickingFreshAccount`
Editing step 5, for some users is showing different message that 'Please enter your _sandbox.koding.com_ username & password.'

hectork01:
Working on `show_edit_remove_credentials` adding instructions to clarify the steps to follow

PR #9902, opened 2016-12-06
Fix/Reword email instructions for Rainforest test #49632

bitspill:
The usage of `rainforest+random.email` appears to have been causing troubles at Step.3 where it would claim an account already existed with that email, also adjusted wording later on when viewing the random.inbox

This pull request is an attempt at partially fulfilling #9891 and I would like to verify I'm doing what's expected before proceeding to any further changes. I can also be found on Slack as `@bitspill`

### DIFF
--- a/tests/spec/rainforest/joinByInvitation_logout_tryToLoginWith_NotAllowedEmail_InvalidPassword.rfml
+++ b/tests/spec/rainforest/joinByInvitation_logout_tryToLoginWith_NotAllowedEmail_InvalidPassword.rfml
@@ -7,7 +7,7 @@
 Click on the 'create a new team' link below the form where it says 'Welcome! Enter your team's Koding domain.'
 Do you see 'Let's sign you up!' form with 'Email Address' and 'Team Name' text fields and a 'NEXT' button?
 
-Enter "rainforest+{{ random.email }}" in the 'Email Address' and "{{ random.last_name }}{{ random.number }}" in 'Team Name' fields and click on 'NEXT' button
+Enter "{{ random.last_name }}{{ random.number }}+{{ random.email }}" in the 'Email Address' and "{{ random.last_name }}{{ random.number }}" in 'Team Name' fields and click on 'NEXT' button
 Do you see 'Your team URL' form with 'Your Team URL' field pre-filled?
 
 Click on 'NEXT' button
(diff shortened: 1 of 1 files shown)


tests/spec/rainforest/joinByInvitation_logout_tryToLoginWith_NotAllowedEmail_InvalidPassword.rfml
```diff
@@ -7,7 +7,7 @@
 Click on the 'create a new team' link below the form where it says 'Welcome! Enter your team's Koding domain.'
 Do you see 'Let's sign you up!' form with 'Email Address' and 'Team Name' text fields and a 'NEXT' button?
 
-Enter "rainforest+{{ random.email }}" in the 'Email Address' and "{{ random.last_name }}{{ random.number }}" in 'Team Name' fields and click on 'NEXT' button
+Enter "{{ random.email }}" in the 'Email Address' and "{{ random.last_name }}{{ random.number }}" in 'Team Name' fields and click on 'NEXT' button
```
diclee: `{{ random.email }}` doesn't solve the problems that `rainforest+{{ random.email }}` may cause. They have the same randomness. Also changing this value as `{{ random.email }}` breaks the next steps of the test where ``{{ random.email }}` is invited as a member (https://github.com/koding/koding/pull/9902/files?diff=split#diff-8f80098be99204fe0da8b9e9f3308253R25)

diclee (changes requested):

tests/spec/rainforest/joinByInvitation_logout_tryToLoginWith_NotAllowedEmail_InvalidPassword.rfml
bitspill: It's not that both emails are randomly being generated to be the same but that koding appears to only check prior to the `+` for uniqueness

https://github.com/koding/koding/blob/ba0fa3efcbcb26eb1ce7aa5612dc405db37e32a8/workers/social/lib/social/models/user/emailsanitize.coffee#L18
https://github.com/koding/koding/blob/ba0fa3efcbcb26eb1ce7aa5612dc405db37e32a8/workers/social/lib/social/models/user/emailsanitize.test.coffee#L38

Admittedly I'm not overly familiar with the codebase so those functions may not apply here but the behavior seems to fit.

diclee (approved):

tests/spec/rainforest/joinByInvitation_logout_tryToLoginWith_NotAllowedEmail_InvalidPassword.rfml
```diff
@@ -26,7 +26,7 @@ Enter "{{ random.email }}" in the 'Email' field that's highlighted, uncheck the
 Have you seen 'Invitation is sent to {{ random.email }}' message displayed?
 
 Open a new incognito window by clicking on the 3 dots on the top corner of the browser and selecting 'New Incognito Window' and then go to '{{ random.inbox }}' by pasting the url (using ctrl-v) in the address bar, wait ~1min and refresh the page
-Do you see 'You are invited to join a team on Koding' email in inbox that received a few minutes ago (and, probably, several older emails)?
+Do you see a newly recieved email with subject 'You are invited to join a team on Koding'? (may take a minute or so to appear)
```
szkl: Typo.

szkl (changes requested):

bitspill:
@szkl Fixed typo and condensed to a single commit

szkl (approved):

szkl:
Fix will be verified with next test runs.

szkl:
Fix for modified 2 steps are verified.

PR #9915, opened 2016-12-06
Appstorage Improvements

gokmen:
Fixes appStorage setter checks and uses synced version of getters in various places.

## Description
Since we provide appStorage data in payload, we don't need to re-fetch storage again.

## How Has This Been Tested?
Manually

## Types of changes


sinan (approved):
💯

PR #9919, opened 2016-12-06
github/scopes: remove ‘admin:org’ scope checks

mehmetalisavas:
Remove 'admin:org' check for terraform github plugin

Updated scope list is pasted below;
image


mehmetalisavas:
cc/ @gokmen

cihangir (approved):

PR #9932, opened 2016-12-07
Disable payment requirements on signup for Default Environment

usirin:
Until this PR we were forcing payment to appear on default environment as well (there were no extra precautions for environments other than `development` and `production`). This PR tries to make sure that open source users of Koding could still continue to use koding with their own installation.

/cc @gokmen

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -245,7 +245,9 @@ module.exports = utils = {
       return _.assign {}, data, { payment: { stripeToken } }
 
     if team = kd.team
-      if token = getPayment()?.token
+      token = getPayment()?.token
+      isDefaultEnv = kd.config.environment is 'default'
+      if not isDefaultEnv and token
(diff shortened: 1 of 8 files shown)


client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -283,10 +285,10 @@ module.exports = utils = {
...
   createTeam: (callbacks = {}) ->
 
-    # this part should never run actually because the signup flow in
-    # TeamAppController takes care of this. But i'll leave this here just in
-    # case.
-    unless token = utils.getPayment()?.token
+    isDefaultEnv = kd.config.environment is 'default'
+    token = utils.getPayment()?.token
+
+    if not isDefaultEnv and not token
```
szkl: unless isDefaultEnv and token

usirin: `unless ... and/or ...` usually makes me stop and try to understand the expression, i thought this was more readable, wdyt?

sinan (approved):

PR #9957, opened 2016-12-08
Account cleaner

mehmetalisavas:
Clean the postgre account that doesn't exist in mongodb.

### DIFF
--- a/go/src/socialapi/models/account.go
+++ b/go/src/socialapi/models/account.go
@@ -2,9 +2,13 @@ package models
 
 import (
 	"fmt"
+	"koding/db/mongodb/modelhelper"
 	"socialapi/request"
 	"strings"
 
+	mgo "gopkg.in/mgo.v2"
+
+	"github.com/hashicorp/go-multierror"
 	"github.com/jinzhu/gorm"
 	"github.com/koding/bongo"
 )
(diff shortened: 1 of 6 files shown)


go/src/socialapi/models/account.go
```diff
@@ -301,3 +305,84 @@ func FetchAccountsByNicks(nicks []string) ([]Account, error) {
...
+	_, err := modelhelper.GetAccountById(a.OldId)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return a.Delete()
+		}
+		return err
+	}
+
+	return nil
+}
```
cihangir: you could omit the if check & return err statement with 

```
func (a *Account) DeleteIfNotInMongo() error {
	if a.OldId == "" {
		return ErrOldIdIsNotSet
	}

	_, err := modelhelper.GetAccountById(a.OldId)
	if err == mgo.ErrNotFound {
		return a.Delete()
	}
	return err
}
```

go/src/socialapi/scripts/accountdeleter/accountdelete_test.go
```diff
@@ -0,0 +1,86 @@
...
+					account, err := rest.CreateAccount(account)
+					So(err, ShouldBeNil)
+					So(account, ShouldNotBeNil)
+					Convey("User should be in postgre when fetch from postgre", func() {
+						accounts, err := models.FetchAccountsWithBongoOffset(100, 0)
+						So(err, ShouldBeNil)
+						So(accounts, ShouldNotBeNil)
+						Convey("Created account should be in account list", func() {
+							var doesExist bool
+							for _, ac := range accounts {
```
cihangir: this test might fail if db has more than 100 accounts

cihangir (approved):
we wont be running the tests, so my comments are not required to be fixed

PR #9969, opened 2016-12-09
Add .nvmrc to .gitignore

usirin:
So that people can use their own versions if they want.


szkl (approved):

usirin:
Thanks, that was fast.

PR #9986, opened 2016-12-09
Show correct amount for free credit on billing dashboard

### LINKED ISSUE #9944: payment: credits are not reflected properly on payment dashboard
## Expected Behavior
Free credit should include account balance

## Current Behavior
it does not show

## Possible Solution
add account balance in negative form to credit.

1. add free credit to an account from cebeci
2.see if it is reflected on billing page

## Context
image

usirin:
It is computed from `customer.account_balance` property if it's less than `0`.

Alt text

/cc @cihangir 

Fixes #9944

### DIFF
--- a/client/app/lib/redux/modules/payment/customer.coffee
+++ b/client/app/lib/redux/modules/payment/customer.coffee
@@ -63,12 +63,21 @@ coupon = (state) -> state.customer?.discount?.coupon
 
 email = (state) -> state.customer?.email
 
+freeCredit = (state) ->
+  # negative account balance means user has credits.
+  if state.customer?.account_balance < 0
+  # the amount is in cents, we want dollars.
+  then ((state.customer.account_balance * -1) / 100).toFixed 2
+  # if the balance is greater than 0 that means 0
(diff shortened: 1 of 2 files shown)


client/app/lib/redux/modules/payment/customer.coffee
```diff
@@ -63,12 +63,21 @@ coupon = (state) -> state.customer?.discount?.coupon
 
 email = (state) -> state.customer?.email
 
+freeCredit = (state) ->
+  # negative account balance means user has credits.
+  if state.customer?.account_balance < 0
+  # the amount is in cents, we want dollars.
+  then (state.customer.account_balance * -1) / 100
```
sinan: `((state.customer.account_balance * -1) / 100).toFixed 2` just in case

sinan (approved):

client/app/lib/redux/modules/payment/customer.coffee
usirin: 👍

sinan (approved):

Issue #9980, opened 2016-12-09
Text change on collaboration invitation and joining modals.


diclee:
## Current Behavior

- [x] should be 'joining collaboration session'
![image](https://cloud.githubusercontent.com/assets/8261892/21049435/79c10b5a-be1d-11e6-9119-74647d7584f8.png)

- [x] should be 'collaborate on'
![image](https://cloud.githubusercontent.com/assets/8261892/21049440/80261198-be1d-11e6-8ea3-85f99bddc007.png)

diclee:
Joining modal is verified on latest but invitation modal still has old copy.

diclee:
Verify passed on latest.

Issue #9979, opened 2016-12-09
Select-box (next to 'Syntax' option on IDE file menu) alignment is a pixel off.


diclee:
## Current Behavior
also select box value text should be right aligned and sized dynamically by its content
![image](https://cloud.githubusercontent.com/assets/8261892/21048554/824671fc-be18-11e6-8a60-ec9ab4070f35.png)

diclee:
@hakankaradis can you please fix below minor issue when you have time?
![image](https://cloud.githubusercontent.com/assets/8261892/21662772/abf70a7a-d290-11e6-9334-8b0ffc051165.png)
Text in the select-box is not aligned correctly.

diclee:
Verify passed on sandbox.

Issue #9976, opened 2016-12-09
When you click "edit name" when you're on the other tabs on stack editor, it switches tab to stack tab instead of showing editing textbox.


diclee:
## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/21046678/72fe2f0a-be0e-11e6-9f90-f792be39e4fe.png)

diclee:
Verify passed on latest.

Issue #9953, opened 2016-12-08
'button is not defined' error message is displayed on the console when you click on 'Renew QR Code' button.


diclee:
## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/21021366/d6ada6a4-bd80-11e6-80a2-ad6cbf3df000.png)

![image](https://cloud.githubusercontent.com/assets/8261892/21021313/a40df708-bd80-11e6-8f1f-4dfa2fdba187.png)

![image](https://cloud.githubusercontent.com/assets/8261892/21021337/bc41ade2-bd80-11e6-88c7-1a8cf0202e15.png)

Seems to be working as expected but may break something else.

diclee:
Verify passed on latest.

PR #10000, opened 2016-12-12
Fix Some Release Bugs

### LINKED ISSUE #9941: 'Terminate' button should be updated as 'Log Out'
## Current Behavior
image

And the modal appearing after clicking on it:
image

Question here should be updated as 'Do you really want to log out from [Team Name]?'

hakankaradis:
Fixes: #9941, #9945, #9949, #9953, #9979, #9980, #9976 

- Members will not be allowed to clone stack when stack creation is disabled.
- Fix some button styles in `are you sure?` modals
- minor copy changes
- Edit name in stack editor section is clickable when the other tabs active
- Style fix for text box next to `Syntax` option in IDE menu items

### DIFF
--- a/client/app/lib/activity/sidebar/sidebarmachinesharepopup.coffee
+++ b/client/app/lib/activity/sidebar/sidebarmachinesharepopup.coffee
@@ -238,5 +238,5 @@ module.exports = class SidebarMachineSharePopup extends KDModalView
 
   TITLES =
     'shared machine' : 'wants to share their VM with you. <span class="footnote">* Rejecting this will destroy any existing collaboration sessions with this machine.</span>'
-    'collaboration'  : 'wants to collaborate with you on their VM.'
+    'collaboration'  : 'wants to collaborate on with you on their VM.'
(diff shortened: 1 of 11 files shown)


client/app/lib/activity/sidebar/sidebarmachinesharepopup.coffee
```diff
@@ -238,5 +238,5 @@ module.exports = class SidebarMachineSharePopup extends KDModalView
 
   TITLES =
     'shared machine' : 'wants to share their VM with you. <span class="footnote">* Rejecting this will destroy any existing collaboration sessions with this machine.</span>'
-    'collaboration'  : 'wants to collaborate with you on their VM.'
+    'collaboration'  : 'wants to collaborate on with you on their VM.'
```
sinan: that's correct copy, the one you need to change is not this one. Apply this `wants to collaborate with you on their VM.` to this https://github.com/koding/koding/issues/9980

sinan (changes requested):
one minor change.

PR #10002, opened 2016-12-12
New Stack editor with FlexSplit layout support

gokmen:
image


gokmen:
@sinan @usirin this is ongoing PR can u review and merge if it's ok? it's disabled by default and currently only loads the stack template. instructions on this to enable/disable https://github.com/koding/koding/pull/9740

client/new-stack-editor/lib/styl/main.styl
```diff
@@ -1,7 +1,181 @@
...
 }
 
+dark-panel-bg               = #2B2B30
+dark-tab-header-bg          = #131315
+dark-active-tab-color       = #9C9C9C
+dark-tab-color              = #5C5C5C
+
+resize-handle-color         = #6EA3E5
+stack-bg-color              = #232327
+stack-fg-color              = #59595C
```
usirin: [minor] prefixing variables with `$` makes it easier to distinguish from other ones.

client/new-stack-editor/lib/views/flexsplit/storage.coffee
```diff
@@ -0,0 +1,65 @@
...
+    then @adapter.get identifier, callback
+    else callback @storage[identifier]
+
+
+  set: (identifier, fractions, callback = kd.noop) ->
+    @storage[identifier] = fractions
+
+
+  store: (callback = kd.noop) ->
+    @adapter?.store @storage, callback
```
usirin: callback is not gonna be called here if `adapter` is undefined.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
```diff
@@ -0,0 +1,122 @@
...
+    @totalSize = @sizes[0] + @sizes[1] + @[@type.getter]()
+
+
+  limited = (num) ->
+    Math.min FlexSplit.MAX, Math.max FlexSplit.MIN, num
+
+
+  _updateFractions: (change = 0, set = yes) ->
+
+    for i in [0..1]
```
usirin: Why not `[0, 1]`? :)

client/new-stack-editor/lib/views/flexsplit/index.coffee
```diff
@@ -0,0 +1,89 @@
...
+      FlexSplit.EVENT_RESIZED
+      FlexSplit.EVENT_COLLAPSED
+    ]
+
+    @resizer.on FlexSplit.EVENT_EXPAND, ->
+      if view.parent instanceof FlexSplit
+        view.parent.emit FlexSplit.EVENT_EXPAND
+
+    @resizer.on FlexSplit.EVENT_COLLAPSE, ->
+      if view.parent instanceof FlexSplit
```
usirin: Can we distinguish with something else? `instanceof` checks are error prone when code is minified Maybe we can generalize this with a field on `kd.Object` like `_type`?

As a side effect, that would remove the need to `require` the file itself in places where we only need this type of check. (it would require you to know the type's string representation, if you don't want to require tho)

usirin: This would go to `constructor` not to the prototype of course.
```coffee
class FlexSplit extends kd.Object
  @_type = 'FlexSplit'
```

Than maybe `kd.Object` can export a function called `isInstance` that would help us to compare instances with their classes:

```coffee
split = new FlexSplit
kd.Object.isInstance split, 'FlexSplit'
# => true
```

Or maybe we can go one step further and make all the subclasses use this `is` function to check if an object is their instance:

```coffee
FlexSplit.isInstance split
# => true
```

client/new-stack-editor/lib/views/flexsplit/storage.coffee
```diff
@@ -0,0 +1,65 @@
...
+        FlexSplit.EVENT_EXPANDED
+        FlexSplit.EVENT_COLLAPSED
+      ], (fractions) =>
+
+        @eventCount++
+        @set identifier, fractions
+
+        if @eventCount is @viewCount - 1
+          @eventCount = 0
+          @store()
```
usirin: Can you explain this part? I don't quite understand what happens if `keepExpandStatus` is `true`.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
```diff
@@ -0,0 +1,122 @@
...
+
+
+  addView: (view) ->
+
+    @views.push view
+
+    view.on FlexSplit.EVENT_EXPAND, =>
+
+      return  if view.hasClass 'expanded'
+      view.setClass 'expanded'
```
usirin: Can we depend on something else other than a `css class`?

I am guessing this might be avoidable via putting this check into `view` class itself.

So this could be rewritten as:

```coffee
return  if view.isExpanded()
view.expand()
```

We will have the advantage of not changing anything if we want to change the way it's represented as `expanded` by only calling `isExpanded` and `expand` (presumably with `isCollapsed` and `collapse` helpers as well) on the view instance.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
```diff
@@ -0,0 +1,122 @@
...
+    super options, data
+
+    @views      = []
+    @_fractions = []
+    { @type, view } = @getOptions()
+
+    @addView view
+
+    @on 'DragFinished', @dragFinished
+    @on 'DragStarted',  @dragStarted
```
usirin: o,O how does this work without `@bound` helper?

client/new-stack-editor/lib/index.coffee
```diff
@@ -47,3 +62,15 @@ module.exports = class StackEditorAppController extends AppController
...
     log '::reloadEditor', templateId, skipDataUpdate
 
+
+  fetchStackTemplate: (templateId, callback) ->
+
+    if template = @templates[templateId]
+      return callback null, template
+
+    cc = kd.singletons.computeController
+    cc.fetchStackTemplate templateId, (err, template) =>
```
usirin: /cc @hakankaradis 

We need to be aware that this instance will not go into the flux stores.

client/app/lib/styl/require-styles.coffee
```diff
@@ -14,4 +14,5 @@ require 'ide/styl'
 # require 'pricing/styl'
 require 'stack-editor/styl'
 require 'new-stack-editor/styl'
+require 'new-stack-editor/views/flexsplit/styl'
```
usirin: This file only has requires to other app's top level `styl/index.coffee`.
let's move this require to `new-stack-editor/lib/styl/index.coffee`

usirin (approved):

client/new-stack-editor/lib/index.coffee
```diff
@@ -25,12 +33,19 @@ module.exports = class StackEditorAppController extends AppController
 
     super options, data
 
+    @templates = {}
```
usirin: Do you really need these to be saved locally? We already have a cache on `reactor` singleton, and all the other parts we have developed is using that as a source of truth. Loading stack templates here locally, will make other parts of the system unaware that this instance is loaded.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
gokmen: why not? O.o

gokmen: https://github.com/koding/kd/blob/master/lib/components/split/splitresizer.coffee#L19-L21

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
```diff
@@ -0,0 +1,122 @@
...
+    for view in @views
+      view.unsetClass 'expanded'
+      view.setClass 'ondrag'
+
+    @setClass 'ondrag'
+
+    @_updateViewSizes()
+
+
+  setFractions: (fractions, set = yes) ->
```
gokmen: it must be named as `updateViews` instead of `set` actually.

client/new-stack-editor/lib/views/localstorageadapter.coffee
```diff
@@ -0,0 +1,12 @@
+kd = require 'kd'
+
+module.exports = class LocalStorageAdapter
+
+  constructor: ->
+    @ls = kd.singletons.localStorageController.storage 'StackEditorLayout'
```
gokmen: if you look at it, these adapters are not part of `flex-split` it's implementation of `Koding` so it's part of stack-editor, I'll shape up constants only for `flex-split`

client/new-stack-editor/lib/index.coffee
gokmen: This was a temporary thing to show stack template contents with-in this PR, I'm planning to make a controller which uses a generic storage (it can use `reactor` singleton if it's already shaped up at there) safely ignore this part. I've some stuff around here as just "experimental"

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
gokmen: I thought about this again and kept it as is for now. Since I don't need any specific class to be as view, I don't want to create another dependency. Right now you can use and `kd.View` as `view` but if I require something like this it will add more over head than it's solving. `expand` class is dynamically added and it's ok if someone removes it, it will just re-trigger expand if they don't handle this status on their own, it won't break anything in `FlexSplit`.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
gokmen: this is why :)

`[0, 1]` produces following `js`

```js
var i, j, len, ref;

ref = [0, 1];
for (j = 0, len = ref.length; j < len; j++) {
  i = ref[j];
  foo(i);
}
```

but `[0..1]` produces following;

```js
var i, j;

for (i = j = 0; j <= 1; i = ++j) {
  foo(i);
}
```

client/new-stack-editor/lib/views/flexsplit/storage.coffee
gokmen: This is currently experimental and disabled by default, what it does is a bit complicated I know :) Let's assume you have nested flex splits; one vertical and inside of that vertical two horizontal flex splits. if one of the horizontal one requests expand this will cause a chain reaction to expand outer vertical one as well (as explained here https://github.com/koding/koding/blob/fde719cba0f3c752e65769a3672b50cbfae7b484/client/new-stack-editor/lib/views/flexsplit/resizer.coffee#L34-L45) this chain reaction will also cause all the affected flex-split's to emit `Flex.EVENT_EXPANDED` (https://github.com/koding/koding/blob/fde719cba0f3c752e65769a3672b50cbfae7b484/client/new-stack-editor/lib/views/flexsplit/resizer.coffee#L56) which will cause this store to fire `store` method `n` times (`n` as times of nested splits you have) to prevent calling this much `store` and call it once when all splits expanded/collapsed I'm counting the events here and comparing them with the number of views that I've in storage.

As I said, this is experimental, it may change or removed, it's not in use currently by StackEditor.

client/new-stack-editor/lib/index.coffee
```diff
@@ -47,3 +62,15 @@ module.exports = class StackEditorAppController extends AppController
...
+
+  fetchStackTemplate: (templateId, callback) ->
+
+    if template = @templates[templateId]
+      return callback null, template
+
+    cc = kd.singletons.computeController
+    cc.fetchStackTemplate templateId, (err, template) =>
+      return callback err  if err
+      return callback Errors.NotExists  unless template
```
hakankaradis: Minor thing, this line will not run, fetchStackTemplate is already checking the template existency. [here](https://github.com/koding/koding/blob/master/client/app/lib/providers/computecontroller.coffee#L1216)

client/new-stack-editor/lib/views/flexsplit/index.coffee
```diff
@@ -0,0 +1,89 @@
...
+        view.parent.emit FlexSplit.EVENT_COLLAPSE
+
+
+  setupViews: ->
+
+    { sizes, views, resizable } = @getOptions()
```
sinan: sum of `sizes` should be checked here or in constructor maybe:

```coffee
if sizes[0] + sizes[1] > 100
  # either, preferably throw
  throw 'learn basic arithmetic please'
  # or fix automagically
  # index == 0
  view.setCss 'flex-basis', "#{size}%"
  # index == 1
  view.setCss 'flex-basis', "#{100 - sizes.first}%"
```

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
sinan: context is same

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
sinan: or it can be done in two methods `_updateFractions` would do the first part and `setFractions` would do the actual setting.

client/new-stack-editor/lib/views/flexsplit/storage.coffee
```diff
@@ -0,0 +1,65 @@
...
+
+  constructor: (options = {}, data) ->
+
+    super options, data
+
+    @storage    = {}
+    @viewCount  = 0
+    @eventCount = 0
+
+    if adapterClass = @getOption 'adapter'
```
sinan: option name should also be `adapterClass` imo.

client/new-stack-editor/lib/views/flexsplit/resizer.coffee
```diff
@@ -0,0 +1,122 @@
...
+      @emit FlexSplit.EVENT_COLLAPSED, fractions
+      view._windowDidResize?()
+
+
+  _updateViewSizes: ->
+    @sizes = [
+      @views[0][@type.getter]()
+      @views[1][@type.getter]()
+    ]
+    @totalSize = @sizes[0] + @sizes[1] + @[@type.getter]()
```
sinan: don't end up like this :)

https://github.com/koding/kd/blob/master/lib/components/split/splitview.coffee#L124-L158

sinan (commented):
💯 thx @gokmen it looks pretty clean, some late improvement suggestions here :)

client/new-stack-editor/lib/views/flexsplit/index.coffee
gokmen: right, there are some places that I didn't put some basic checks like this, I'll add it now 👍

client/new-stack-editor/lib/views/flexsplit/storage.coffee
gokmen: I thought the same while implementing but then it didn't look good :) we can change it to `adapterClass` for sure.

Issue #9972, opened 2016-12-09
'original stack template has been updated' notification appears for the user that cloned also when he makes the changes himself.


diclee:
## Current Behavior
Only the clone stack template is changed, the original stack template is not changed in below example.
![gif](http://g.recordit.co/H0D7fLq2Jm.gif)
http://recordit.co/H0D7fLq2Jm

hakankaradis:
@diclee 
I think it is kind of expected behaviour. 
When you came to stack-editor page, it checks that the current stack template is clone of one of the shared stack template and if so, it compares its content with original stack template and if there is difference, the warning view is appeared.

I think having warning view is expected but copy is wrong. Because it say the original stack template has been updated, I think it should say `This stack template is different than original one` or something. I believe it will make more sense? 
@gokmen can you please share your thought for this bug please?

gokmen:
@hakankaradis this warning should only be visible if original content has changed not the cloned content.

diclee:
Verify passed on latest.

PR #10013, opened 2016-12-13
'original stack template has been updated' notification appears for the user that cloned also when he makes the changes himself. 

hakankaradis:
Fixes: #9972

When cloned stack template has update, we need to update its clonedSum info as well.
And while checking update for cloned stack template, we are looking for clonedSum information that we set when we clone the stack template.

http://recordit.co/SlHLMybrXI

### DIFF
--- a/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
+++ b/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
@@ -441,7 +441,8 @@ module.exports = class JStackTemplate extends Module
             # update clonedSum information in the template, if it is exists
             JStackTemplate.one$ client, { _id: originalId }, (err, template) ->
               if template and not err
-                data.config.clonedSum = template.template.sum
+                if template.template.sum is data.template.sum
+                  data.config.clonedSum = template.template.sum
               next()
         (next) =>
           query = { $set: data }
--- a/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
+++ b/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
@@ -435,7 +435,14 @@ module.exports = class JStackTemplate extends Module
 
             data['meta.modifiedAt'] = new Date
 
-            next()
+            originalId = data.config?.clonedFrom
+            return next()  unless originalId
+
+            # update clonedSum information in the template, if it is exists
+            JStackTemplate.one$ client, { _id: originalId }, (err, template) ->
+              if template and not err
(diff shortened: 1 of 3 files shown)


workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -435,7 +435,14 @@ module.exports = class JStackTemplate extends Module
...
             data['meta.modifiedAt'] = new Date
 
-            next()
+            originalId = data.config?.clonedFrom
+            return next()  unless originalId
+
+            # update clonedSum information in the template, if it is exists
+            JStackTemplate.one$ client, { _id: originalId }, (err, template) ->
+              if template and not err
+                data.config.clonedSum = template.template.sum
```
gokmen: why do we need this update? it will remain as is, isn't it?

hakankaradis: Let's say: Original Stack Template = A, cloned stack template = B, Stack A's sum = `xxx1`, Stack B's config's clonedSum = `xxx1`, Stack B's sum = `xxx1` as well.

We made Stack A's sum `xxx2` and then Stack B gets updates and we also update cloned stack, and made its sum `xxx2`. If we don't have above code, its config's clonedSum will remain `xxx1`.If so, it will be `xxx2`

Then we just update the Stack B, and made its sum `xxx3`. If Stack B's config's clonedSum is `xxx1` that means original stack template updated but cloned stack template has no idea about the new sum info.

So either we update the cloned stack template because of the original one updated or just the cloned one updated, we need to update its config's clonedSum info as well, otherwise it will read stale sum info.

hakankaradis: hmmm you update stack B while ignoring the update in stack A. Yes you are right, editor will think that I have same content with original one but it doesn't.
Whether I am updating the cloned stack template with original one or changes that I made, the config of clonedSum should be sync with original one. It only affects to show or not the warning view
I think the code should work with this scenario as well. I will give a try soon and update you.

hakankaradis:
@gokmen pr is updated.

http://recordit.co/u99O0XQrR8

As shown in screencast, the original one updated, then the cloned one got warning but I made the template content same by hand, then update it. Result, warning view is disappeared as expected.

http://recordit.co/Uupz6bPPq3
As shown in second screencast, the original one updated, then the clone one got warning and I add different line than the original one and update the template. Result, the warning view is still there as expected.

gokmen (approved):

Issue #9956, opened 2016-12-08
Adding a credit card doesn't renew the subscription for cancelled ones


usirin:
Credit card submission form now only updates the customer's default payment source, but it doesn't create a subscription, resulting in the website not being used at all.

Immediate fix is to make sure that, if there is no subscription, we create subscription with the same request as well.

- [x] **Disabled as it's currently not possible to make authorize request while logged in** _Make an `authorization` request using `/creditcard/auth` endpoint._ [ref](https://github.com/koding/koding/blob/master/client/landing/site.landing/coffee/core/utils.coffee#L521-L529)

  **Endpoint**: `/creditcard/auth`
  **Request Params**: `{ source: { token: String }, email: String }`
  **Response**: `200` if card is authorized, or relevant `4xx` Error code.
  **Action Type**: `creditCard.AUTHORIZE`
  **Action Creator**: 
  ```coffee
  creditCard.authorize: ({ source: { token: String }, email: String }) => AuthorizeAction
  ```
  - [x]  `payment service` exports a function called `authorize` to **make request** to payment backend. https://github.com/koding/koding/commit/b1632d033bb4b1d8ec516e52df8f37c1384f3065
    - `authorize` function accepts a valid **stripe token** as `source`, and a valid **email address** as email.
    - returns a `Promise` which resolves when request ends.
  - [x] `creditCard` module exports an `authorize` action, which wraps `authorize` function of `payment service`. https://github.com/koding/koding/commit/ee8585e2f616f5c34231eca3cde592560d6560de
  - [x] `authorize` action of `creditCard` module will return a `AuthorizeAction`, when dispatched it returns a `Promise` that **resolves** if `authorization` request is successful, or **rejects** if it fails.
- [x] Make a call to `createSubscription` after getting a success response from `/creditcard/auth` endpoint https://github.com/koding/koding/commit/78d2eeebf7d0eb1f4ed758100ea66e486f7e135a

usirin:
@cihangir When i wanted to make an authorization call to backend i got this error:

```
ERROR [response/response.go:74][response] Access Denied Err: logged out users only
```

We need to enable this endpoint for logged in users too.

usirin:
1 possible solution could be not sending the cookie with request and act like not logged in, but it feels a bit hacky to me. What do you think @cihangir ?

diclee:
Verify passed on latest.

PR #10014, opened 2016-12-13
socialapi/payment: remove requirement of being logged out while auth …

cihangir:
We should validate the CC while changing it.


cihangir:
related with https://github.com/koding/koding/issues/9956

usirin (approved):

Issue #9889, opened 2016-12-05
Add kd machine ssh subcommand 


ppknap:
Old kd ssh command uses Kontrol based machines and it will not be compatible with new Kloud based implementation

rjeczalik:
It should also support `/-/discover/ssh` route for tunnels (supported in current implementation - see [tunnelproxy/discover](https://github.com/koding/koding/tree/master/go/src/koding/kites/tunnelproxy/discover)).

PR #10018, opened 2016-12-13
klientctl: rework klientctl/ssh package and make it business logic agnostic

ppknap:
This PR separates SSH-related logic, stored in `klientctl/ssh` package, from klient's business logic part.

Part of #9889

## Description
The old implementation of `klientctl/ssh` package (now in `sshold.go` file) mixes ssh local machine logic with remote klient logic and remote machine discovery. The new one (`ssh.go` && `ssh_test.go`) contains only local SSH related operations (generating keys etc.). Because of this, the package can be moved out of `klientctl` subdirectory(or even to `koding/sshkey` repository) since its logic is not related with `klientctl` anymore.

## Motivation and Context
It was impossible to use(without ugly hacks) existing `kd ssh` logic in `kd machine ssh` 

## How Has This Been Tested?
Unit tests.

## Types of changes

**NOTE**: `sshold.go` should not be reviewed. It's just old `ssh.go` file renamed to `sshold.go`. Git didn't resolve this as file move.

### DIFF
--- a/go/src/koding/klientctl/ssh/sshold.go
+++ b/go/src/koding/klientctl/ssh/sshold.go
@@ -1,0 +1,408 @@
+// The core logic for the `kd ssh` command.
+//
+// TODO(ppknap) this logic is deprecated. Remove it once we switch to new
+// machine commands entirely.
+package ssh
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"net"
+	"os"
+	"os/exec"
+	"os/user"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"koding/kites/tunnelproxy/discover"
+	"koding/klient/remote/req"
+	"koding/klientctl/config"
+	"koding/klientctl/klient"
+	"koding/klientctl/klientctlerrors"
+	"koding/klientctl/list"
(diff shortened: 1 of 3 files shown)


go/src/koding/klientctl/ssh/sshold.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+
+		if klientctlerrors.IsDialFailedErr(err) {
+			return ErrRemoteDialingFailed
+		}
+
+		return ErrFailedToGetSSHKey
+	}
+
+	args := []string{
```
rjeczalik: Can we add here something like `-o StrictHostKeychecking=no` and `-o UserKnownHostsFile=/dev/null`, in case of ephemeral IPs vms can re-use a host that was already added, which makes the ssh command to fail. Happened to me a number of times.

rjeczalik (commented):
lgtm

rjeczalik (approved):

cihangir (approved):

go/src/koding/klientctl/ssh/sshold.go
rjeczalik: oh, yeah, I have always problems with rtfm

PR #10023, opened 2016-12-14
Various fixes and improvements for providers and credentials

gokmen:
## Description
Currently we are not keeping information of which provider is selected for given stack template, we're guessing that information from stack template content. And the way we were doing it was relying on parsing variables with in stack template by using known provider slugs. Since `vagrant` provider doesn't have such information there were no way to understand which provider is using by given stack template. I've updated parsing flow to include `#{provider}_` keys as well. It's not the perfect solution, we need to move all these operations to backend to make things easier and available for `kd` operations as well. I'll make those changes with in new Stack Editor, this one is enough to fix current issues that we've in production right now.

## How Has This Been Tested?
Manually; I've created stacks with/without `userInputs`, `custom variables` with `vagrant`, `digitalocean`, `aws` providers and verified them they are working fine.

## Types of changes

## Checklist:

### DIFF
--- a/client/app/lib/kite/kites/kloud.coffee
+++ b/client/app/lib/kite/kites/kloud.coffee
@@ -7,10 +7,7 @@ globals = require 'globals'
 
 module.exports = class KodingKiteKloudKite extends require('../kodingkite')
 
-  SUPPORTED_PROVIDERS = do ->
-    { providers } = globals.config
-    (Object.keys providers).filter (provider) ->
-      providers[provider].supported
+  SUPPORTED_PROVIDERS = globals.config.providers._getSupportedProviders()
 
   debugEnabled = ->
     kd.singletons.computeController._kloudDebug
--- a/client/app/lib/providers/computecontroller.coffee
+++ b/client/app/lib/providers/computecontroller.coffee
@@ -37,7 +37,7 @@ require './config'
 module.exports = class ComputeController extends KDController
 
 
-  @providers = globals.config.providers
+  @providers = globals.config.providers._getSupportedProviders()
   @Error     = {
     'TimeoutError', 'KiteError', 'NotSupported'
     Pending: '107', NotVerified: '500'
@@ -161,9 +161,9 @@ module.exports = class ComputeController extends KDController
     if method?
       switch method
         when 'reinit'
-          return NotSupported  if provider in ['aws', 'vagrant']
+          return NotSupported
         when 'createSnapshot'
-          return NotSupported  if provider in ['aws', 'softlayer', 'vagrant']
+          return NotSupported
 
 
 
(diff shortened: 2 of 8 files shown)


gokmen:
@rjeczalik there is another issue came up after fixing some of the issues in this PR. In current stack-editor if you put some custom or userInputs as commented;
```
${var.custom_foo}
# ${var.custom_bar}
```
we were taking into account as a requirement as well (in this case `bar` will be requested), and it was showing up in custom variables warning as a requirement. But in this PR I've fixed that by ignoring commented out lines https://github.com/koding/koding/pull/10023/commits/4bcf8f22e0522110ef306db1a3c36d2289c73885 even tough I've realized same issue exists on Kloud/Terraform side as well. I'm not sure if it's an expected behaviour or our issue in variable parser;

As example given stack template for `vagrant`;

```
# Here is your stack preview
# You can make advanced changes like modifying your VM,
# installing packages, and running shell commands.

resource:
  vagrant_instance:
    # this is the name of your VM
    vagrant-instance:
      # define your vm specs here, 2 cpus, 2GB of memory etc.
      cpus: 2
      memory: 2048
      # select your image (optional) eg. ubuntu/trusty64 (it should be based on ubuntu 14.04)
      box: ubuntu/trusty64
      user_data: |-
        # let's create a file on your root folder:
        echo "hello world!" >> /helloworld.txt
        echo ${var.custom_foo}
        # echo ${var.custom_bar}
```
won't ask me for `${var.custom_bar}` after this PR with following logs;
```
[13:42:02] Parsing template for credential requirements...
[13:42:02] Following credentials are required:
[13:42:02] - [
	"vagrant"
]
[13:42:02] Parsing template for additional requirements...
[13:42:02] Following information will be fetched from variables section:
[13:42:02] [
	"foo"
]
[13:42:02] Template content saved.
```
but when I try to initialize it, it's failing with the error below;
```
You got an error:
apply failed with code: 1, output: o:There are warnings and/or errors related to your configuration. Please
fix these before continuing.

e:Errors:

e:  * 1 error(s) occurred:

* module root: 1 error(s) occurred:

* resource 'null_resource.vagrant-instance' config: unknown variable referenced: 'custom_bar'. define it with 'variable' blocks
 (31460310dc07d951)
```
my custom variable section at the time was only including `foo: 'gokmen'` as data.

As a final note, interestingly the same example won't let me initialize the stack and gives the same error in stack editor for `digitalocean` provider;

```
[13:57:40] Something went wrong with the template:
[13:57:40] apply failed with code: 1, output: o:There are warnings and/or errors related to your configuration. Please
fix these before continuing.

e:Errors:

e:  * 1 error(s) occurred:

* module root: 1 error(s) occurred:

* resource 'null_resource.do-instance' config: unknown variable referenced: 'custom_bar'. define it with 'variable' blocks

[13:57:40] Parsing failed, please check your template and try again
```

rjeczalik (approved):

rjeczalik:
> even tough I've realized same issue exists on Kloud/Terraform side as well. I'm not sure if it's an expected behaviour or our issue in variable parser;

It's an issue in variable parser, afair tf has no concept of bash comments, it treats user_data as one big string and looks for vars there.

gokmen:
> It's an issue in variable parser, afair tf has no concept of bash comments, it treats user_data as one big string and looks for vars there.

can we simply remove all comments before passing to tf somehow?

rjeczalik:
> can we simply remove all comments before passing to tf somehow?

Yeah, I've edited my comment to say exactly that. Instead of removing lines (someone could have valid comments which one would not want to lose), we can just escape tf variables as shown above. wdyt?

szkl (approved):

PR #10043, opened 2016-12-16
Replace embedded Ace with Browserified version

gokmen:
This PR removes old Ace we were using and serving under `thirdparty` folder. It's using the latest stable version of Ace (03.12.16 build) v1.2.6 from https://github.com/koding/brace repo which is based on https://github.com/thlorenz/brace . Our implementation includes missing snippets in `brace` (there is a pr for that [in upstream](https://github.com/thlorenz/brace/pull/75)) but we will keep using our fork since we are adding our own themes/modification for now.

It's currently covering all the existing features we provide in current Ace implementation (including snippets and Emmet support) plus this one uses latest stable Ace with more improvements in performance.

There is a downside of this change, this one uses `require`s instead of getting required modes/themes/keybindings or snippets asynchronously. Which I made some tests to see the change on generated `bundle.js` (there is an unknown issue with the current production or development version which you'll realize in report);

## bundle.js on production
image

## bundle.js on localhost _(with `this` branch)_
image

difference between **production** vs. **this branch** is approx. **`300kb` which is reasonable**. but interesting thing happening here on my test on localhost with current `master` 
(**all tests relies on `make dist` under `client`**)

## bundle.js on localhost _(with `master` branch)_
image

it's way lower than others. if this is the correct one and if we are doing something wrong in current `production` we need to fix that one and look at this one after all.

I'll try to find a way to `async` require some of the external requirements of `Ace` with this  implementation.


szkl (approved):
LGTM

usirin (approved):

PR #10055, opened 2016-12-19
klient: restore stored state on klient initialization

ppknap:
This PR allow to restore klient state after restart/reinitialization.

## Motivation and Context
Until now, `machine list` started clients, they must be started without this command to make `machine mount` work correctly.

~Depends on: #10054~

## How Has This Been Tested?
Unit tests.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #10056, opened 2016-12-19
klient: update machine group internal state after machine list command

ppknap:
This PR closes clients and does clean ups for machines not present in Kloud.

## Description
Depends on: ~#10055~

## Motivation and Context
We don't want to maintain state of non existing machines.

## How Has This Been Tested?
Unit tests.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinetest/machinetest.go
+++ b/go/src/koding/klient/machine/machinetest/machinetest.go
@@ -11,7 +11,7 @@ import (
 )
 
 // DynamicClientOpts creates test-friendly options for dynamic client.
-func DynamicClientOpts(s *Server, b *NilBuilder) machine.DynamicClientOpts {
+func DynamicClientOpts(s *Server, b *ClientBuilder) machine.DynamicClientOpts {
 	return machine.DynamicClientOpts{
 		AddrFunc:        s.AddrFunc(),
 		Builder:         b,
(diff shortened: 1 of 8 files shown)


rjeczalik (approved):

go/src/koding/klient/machine/machinetest/machinetest.go
```diff
@@ -145,6 +153,17 @@ func (n *NilBuilder) WaitForBuild(timeout time.Duration) error {
...
 
 // BuildsCount returns how many times Build method was invoked.
-func (n *NilBuilder) BuildsCount() int {
+func (n *ClientBuilder) BuildsCount() int {
 	return int(atomic.LoadInt64(&n.buildsN))
 }
+
+// WaitForContextClose waits until context is done. It times out after specified
+// duration.
+func WaitForContextClose(ctx context.Context, timeout time.Duration) error {
```
cihangir: why did you choose this over "context.WithTimeout" ?

cihangir (approved):

go/src/koding/klient/machine/machinetest/machinetest.go
ppknap: i wanted to separate businesses logic from its tests. If I used `context.WithTimeout`:

```go
ctx, _ := context.WithTimeout(foobar.Context(), time.Second)
<-ctx.Done()
if err := ctx.Err(); err == context.DeadlineExceeded {
   // Context timed out either because test timeout or parent context timed out.
}
```
I would need to know the underlying logic to write correct tests. Even if i knew it, I wouldn't be able to distinguish who closed the context (parent context timed out(not an error) or testing context timed out(error)).

Currently, I'm only `cancel` context so the error will be `context.Cancelled` and `context.WithTimeout` solution is going to work. However, If anyone adds timeouts to the logic, this may trigger false negative tests results.

Issue #9643, opened 2016-11-21
kd: add auth login subcommand / rework installation


rjeczalik:
Makes it possible to login (create jSession and fetch kite.key)

TODO

- [x] kites/config: rework to use koding base url
  - [x] PR #10003 
- [x] kd: add support for --koding flag (multi-configuration) (like `kd auth login --koding sandbox.koding.com`)
  - [x] kites/config/configstore: client package
  - [x] kites/config/configstore: implicit migration for single-configuration
  - [x] kites/config/configstore: rework kd/klient to use new client package
  - [x] kites/config/configstore: store kite.key in bolt + implicit migration
  - [x] kites/config/configstore: rework metadata dump + registration
  - [x] kites/config/configstore: kd config list
  - [x] kites/config/configstore: kd config use
  - [x] kloud: new backend for config update
  - [x] kites/config/configstore: store all configuration in bolt
  - [x] kites/config/configstore: kd config update
  - [x] PR #10021
- [x] kloud/kontrol: add API for getting Koding Metadata (e.g. kite.key)
- [x] kd: add frontend for `auth login` - it requires kite.key for kloud and clientId for remote.api
- [x] klient: rework installation to be able to `brew install kd` - relates to https://github.com/koding/koding/issues/8965, https://github.com/koding/koding/issues/8936 and https://github.com/koding/koding/issues/9250 (installing dependencies etc.)
  - [x] kd: initial auth login support
    - [x] PR #9870
    - [x] PR #9851
  - [x] kloud: add auth.login method
    - [x] PR #9863 

Implements

`kd auth login`

rjeczalik:
TODO

- [ ] fix presence/ping

```
2016-12-21 00:05:31 [klient] ERROR    failed to ping: Get http://rjeczalik.koding.team/api/social/presence/ping: key not found
```

- [ ] improve tunnel retry - do not retry on key mismatch error (ignore invalid cache entry)

```
2016-12-20T23:04:59.653Z CRITICAL [tunnelclient   ][PID:40477][logging/context.go:23] [transport] client fetch identifier error: authenticationError: kiteKey: invalid kontrol key found (00e2ae560f8a90a7)
```

- [ ] fix kd config use

```
$ kd config list
ID        KODING URL
137cb791  http://rjeczalik.koding.team
7a823edf  https://koding.com
```
```
$ kd config use 7a823edf
Configuration "7a823edf" was not found. Please use "kd config list" to list available configurations.
```

[...]

PR #10069, opened 2016-12-19
kd: improvements for configuration

rjeczalik:
pic
(http://recordit.co/nmVUGj2Yn5)


rjeczalik:
@cihangir @ppknap Ptal, there's going to be part 2 that addresses [issues listed here](https://github.com/koding/koding/issues/9643#issuecomment-268386462). They already exist on master but klient/kd is not deployed to production channel so it's only problem for sandbox at the moment.

go/src/koding/kites/config/config.go
```diff
@@ -104,6 +104,36 @@ func (u *URL) WithPath(paths ...string) *URL {
 	return &URL{URL: &ur}
 }
 
+// StringOrBool implements json.Unmarshaler, which can parse
+// either boolean type or converts string to bool.
+type StringOrBool bool
+
+var _ json.Unmarshaler = (*StringOrBool)(nil)
+
+func (b *StringOrBool) UnmarshalJSON(p []byte) error {
```
cihangir: ParseBool should be enough

```
func (b *StringOrBool) UnmarshalJSON(p []byte) error {
	v, err := strconv.ParseBool(string(p))
	*b = StringOrBool(v)
	return err
}
```

rjeczalik: It's not, input can be either `[]byte("true")` (boolean) or `[]byte("\"true\"")` (string), thus the unmarshalling. Alternatively I can use strconv.Unquote which could spare two lines or three, up to you :-)

go/src/koding/kites/kloud/stack/team.go
```diff
@@ -39,3 +42,35 @@ func (k *Kloud) TeamList(r *kite.Request) (interface{}, error) {
...
+		ClientURL:   r.Client.URL,
+		Environment: r.Client.Environment,
+	}
+
+	group, err := modelhelper.LookupGroup(opts)
+	if err != nil {
+		return nil, models.ResError(err, "jGroup")
+	}
+
+	members, _ := group.Counts["members"].(int)
```
cihangir: drop support for "members" it will be removed.

cihangir (approved):

go/src/koding/kites/kloud/stack/team.go
rjeczalik: Done.

@ppknap This should also be dropped from `machine.list`. In separate PR. Does anything depends on it?

go/src/koding/kites/config/config.go
cihangir: under normal circumstances it should :) tbh we shouldnt need write our own marshaller for parsing bool, we could just use the string tag in json tags.  

that sounds like an input error to me. Where are you getting the config from? We may need to fix the places where we are writing to it instead of trying to adjust ourselves to it - if it is already not too late :)

Using "Unquote" could create problems where the input is `[]byte("true")` we can leave it as is

rjeczalik: This type here is explicitly for the purpose of accepting boolean value of its stringified encoding.

`kd config set debug true` - the argument here is always a string, so it must be possible to unmarshal `{"debug": "true"}` onto *config.Konfig value. I could make the field `Debug string` instead, but then I would need to parse it either way in every place I would want to use it.

So the StringOrBool basically implements an union - I was loosely inspired by [IntOrString](https://github.com/kubernetes/kubernetes/blob/81d788dd6e0748e5d53a62c16d933c5f7a0718af/staging/src/k8s.io/client-go/pkg/util/intstr/intstr.go#L43) wrapper.

cihangir: If i understand it correctly you could just use the json string tag then
It would read a string into a bool and write a bool as string
And you would not need to parse it every time you use it.

```
type Konfig struct {
       ......
        Debug StringOrBool `json:"debug,string"`
        .....
}

```

rjeczalik: TIL, done. 💯

go/src/koding/kites/kloud/stack/team.go
rjeczalik: @ppknap then `team.list` does

ppknap (approved):

PR #10079, opened 2016-12-20
Stack Editor

### LINKED ISSUE #10060: Lots of 'Mixed Content' error displayed on the console
## Current Behavior
image
image

Also 'This page is trying to load scripts from unauthenticated sources' notification appears on address bar.

gokmen:
ref: http://recordit.co/dCR7Mj5heP

Improvements on StackEditor continues, layout requirements, toolbar and statusbar basic structure is ready and all main editor requirements done with this PR.

I've also changed styling on sidebar for stack titles when collapsed;

image

it's as is on expanded mode

image

This is not intended to be a final design we will update the sidebar with new design later. I made this for now since the current version looks terrible. cc/ @ftasdelen @sinan 

fixes: #10060


client/new-stack-editor/lib/views/statusbar.coffee
```diff
@@ -0,0 +1,18 @@
...
+  constructor: (options = {}, data) ->
+
+    options.cssClass = kd.utils.curry 'statusbar', options.cssClass
+    data ?= { row: 1, column: 1, title: '...' }
+
+    super options, data
+
+
+  pistachio: ->
+    '{h3{#(title)}} <caret>{{#(row)}}:{{#(column)}}</caret>'
```
sinan: do we need a custom html tag here? `<caret>`

client/new-stack-editor/lib/views/toolbar.coffee
```diff
@@ -0,0 +1,28 @@
...
+JView = require 'app/jview'
+
+
+module.exports = class Toolbar extends JView
+
+
+  constructor: (options = {}, data) ->
+
+    options.cssClass = kd.utils.curry 'toolbar', options.cssClass
+    data ?= { title: '...' }
```
sinan: `…`

client/new-stack-editor/lib/views/statusbar.coffee
```diff
@@ -0,0 +1,18 @@
...
+JView = require 'app/jview'
+
+
+module.exports = class Statusbar extends JView
+
+
+  constructor: (options = {}, data) ->
+
+    options.cssClass = kd.utils.curry 'statusbar', options.cssClass
+    data ?= { row: 1, column: 1, title: '...' }
```
sinan: `…` -> `option + ;`

sinan (approved):
looks good 👍 minor improvements noted.

client/new-stack-editor/lib/views/statusbar.coffee
gokmen: this is `data` which will be set after so in `option` there is no requirement to have `title`

client/new-stack-editor/lib/views/toolbar.coffee
gokmen: I can replace this as `Loading...` it won't be visible most of the time though

client/new-stack-editor/lib/views/statusbar.coffee
gokmen: I thought `<div class='caret'> ... </div>` vs. `<caret> ... </caret>` looks better :)

sinan: I agree, but then you never know how other browsers will behave. Just recently fixed IE edge problems with the `<main>` tag that's why commented. You can do sth like `'{h3{#(title)}} <span>{{#(row)}}:{{#(column)}}</span>'`

Issue #10087, opened 2016-12-21
klient/machine: do not log kite warnings when machine is offline


rjeczalik:
On `kd machine ls` when vm is offline pinger will spam log with:

```
2016-12-21T14:53:07.772Z WARNING  [machine        ][PID:38114][logging/context.go:35] [machines][5859b85a68795bff132571b3][monitor] Machine ping error: no kites availabile
2016-12-21T14:54:07.887Z WARNING  [machine        ][PID:38114][logging/context.go:35] [machines][5859b85a68795bff132571b3][monitor] Machine ping error: no kites availabile
2016-12-21T14:54:37.769Z WARNING  [machine        ][PID:38114][logging/context.go:35] [machines][5859b85a68795bff132571b3][monitor] Machine ping error: no kites availabile
```

It should not.

rjeczalik:
More logs:

```
2016-12-23T09:52:43.893Z WARNING  [machine        ][PID:13584][logging/context.go:35] [machines][585c0617fe196e586009f759][monitor] Machine ping error: address not found
2016-12-23T09:52:43.894Z WARNING  [machine        ][PID:13584][logging/context.go:35] [machines][585c0df1c601062c41a3b911][monitor] Machine ping error: address not found
2016-12-23 10:52:43 [klient] INFO     Kite 'rjeczalik/development/kd' is disconnected
2016-12-23 10:52:43 [klient] INFO     Start disconnection timer with 1 minute delay.
2016-12-23T09:52:44.046Z WARNING  [machine        ][PID:13584][logging/context.go:35] [machines][5859b85a68795bff132571b3][monitor] Machine ping error: no kites availabile
2016-12-23 10:52:44 [klientpool] INFO     Creating new klient connection to ///////4d411f74-7770-4c13-ad96-c30750d057ed
2016-12-23T09:52:44.765Z INFO     [machine        ][PID:13584][logging/context.go:47] [machines][585c0d9dc601062c41a3b90b][monitor] Reinitializing client with kite address: ///////4d411f74-7770-4c13-ad96-c30750d057ed
2016-12-23T09:52:58.967Z WARNING  [machine        ][PID:13584][logging/context.go:35] [machines][585c0617fe196e586009f759][monitor] Machine ping error: address not found
2016-12-23T09:52:58.967Z WARNING  [machine        ][PID:13584][logging/context.go:35] [machines][585c0df1c601062c41a3b911][monitor] Machine ping error: address not found
[...]

PR #10104, opened 2016-12-23
klient: limit ping error logs to not print them periodically

ppknap:
Fixes: #10087

### DIFF
--- a/go/src/koding/klient/machine/dynamic.go
+++ b/go/src/koding/klient/machine/dynamic.go
@@ -160,22 +160,57 @@ func (dc *DynamicClient) cron() {
 		pingTick    = time.NewTicker(dc.opts.PingInterval)
 	)
 
-	curr, islog := Addr{}, true
-	dc.tryUpdate(&curr, &islog)
+	addr, err := Addr{}, error(nil)
+
+	// tryUpdate uses client builder to ping the machine and updates dynamic
+	// client if machine address changes.
+	tryUpdate := func() {
+		stat, a, e := dc.opts.Builder.Ping(dc.opts.AddrFunc)
+		if e != nil {
+			if err == nil {
+				dc.log.Warning("Machine ping error: %s", e)
+			}
+			err = e
+			return
+		}
+		err = nil
+
+		if a.Network == addr.Network && a.Value == addr.Value {
+			// Client address did not change.
+			return
+		}
+
+		// Create new client.
+		dc.log.Info("Reinitializing client with %s address: %s", a.Network, a.Value)
+		ctx, cancel := context.WithCancel(context.Background())
+		c := dc.opts.Builder.Build(ctx, a)
+
+		// Update current address.
+		addr = a
+
+		dc.mu.Lock()
+		if dc.cancel != nil {
+			dc.cancel()
+		}
+		dc.c, dc.stat, dc.ctx, dc.cancel = c, stat, ctx, cancel
+		dc.mu.Unlock()
+	}
+
+	tryUpdate()
 	for {
 		select {
 		case <-dynAddrTick.C:
 			// Look address cache for new addresses. This does not require
 			// pinging remote machines because it only checks current address
 			// book state. Thus, it may be run more frequently than ping.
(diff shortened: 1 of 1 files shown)


go/src/koding/klient/machine/dynamic.go
```diff
@@ -192,10 +192,14 @@ func (dc *DynamicClient) cron() {
 
 // tryUpdate uses client builder to ping the machine and updates dynamic client
 // if machine address changes.
-func (dc *DynamicClient) tryUpdate(addr *Addr) {
+func (dc *DynamicClient) tryUpdate(addr *Addr, islog *bool) {
 	stat, a, err := dc.opts.Builder.Ping(dc.opts.AddrFunc)
 	if err != nil {
-		dc.log.Warning("Machine ping error: %s", err)
+		if *islog {
```
cihangir: there is 26% chance of failure with pointer panic


>
<<<~spoiler~>>>
http://www.ign.com/videos/2016/12/12/rogue-one-a-star-wars-story-chance-of-failure-clip
<<<~/spoiler~>>>

ppknap: There is 1e-84 chance of failure with pointer panic :smile:. `islog` is always non-nil and this is under private interface.

cihangir: there is nothing prevents someone from calling `tryUpdate` outside of cron() ;) i would accept that argument if it were in `cron()` as anon/inline func. 🙋‍♂️

rjeczalik: I would rather add a ThrottledLogger type and used it globally - the logic whether to display a log belongs to Logger not a DynamicClient.

```go
type ThrottledLogger struct {
   Throttle time.Duration
   Log      logging.Logger
}

func (th *ThrottledLogger) Error(fmt string, args ...interface{}) {
    th.logOnce(th.log.Error, fmt, args...)
}

...
```

This way we could get rid of the bool patching here.

rjeczalik: s/globally/in every place in machine the logs are produced/

ppknap: > there is nothing prevents someone from calling tryUpdate outside of cron() ;) i would accept that argument if it were in cron() as anon/inline func. :raising_hand_man:

Initially it was inlined, I moved it to separate method just to improve readability. And I think that private functions are implementation details to the package so we don't need defensive programming for them. I can find many places in the code base where you did the same [assumptions](https://github.com/koding/koding/blob/f365afd539f663052f5d9c2a60fe2fba749c0436/go/src/koding/workers/removenonexistents/main.go#L229) and, as I said, I don't treat them as errors ([because they are not](https://golang.org/src/net/http/client.go?s=6486:6538#L209)).

rjeczalik (approved):

go/src/koding/klient/machine/dynamic.go
cihangir: Just to note that, the code you pointed is a one-off script which is run manually. It is not deployed, it is not monitored, it is not in any critical path of our production workload or any user facing features/infrastructure. Please check tunnelproxymanager as an example that contains many pointer variables because of aws-sdk-go's structs.

It is always better to be constructive than destructive. Pointing fingers is not from the constructive ones.

The code you linked also irrelevant to your proposal. It is an exposed function.

cihangir (approved):

go/src/koding/klient/machine/dynamic.go
ppknap: > The code you linked also irrelevant to your proposal. It is an exposed function.

I was tricked by golang URLs format. And changed:
https://golang.org/src/net/http/client.go?s=6486:6538#L189
to
https://golang.org/src/net/http/client.go?s=6486:6538#L209
but it still links to line 189. Should have checked this.

I meant line [209](https://golang.org/src/net/http/client.go?#L209), where req is not checked for `nil`.

You are right, pointing to someone's code is generally destructive. Unless it's pointing to code of people you respect as deveopers - I pointed yours and these written by go authors:)

PR #10101, opened 2016-12-23
kd: improvements for configuration

rjeczalik:
Last round of improvements for kd / klient configuration. Next step is to finalise `kd auth login` / `kd stack create` and start adding integration tests for kd.

PR #10105, opened 2016-12-23
klient: remove nondeterministic check from client tests

ppknap:
This PR removes nondeterministic check which can result in false negative tests. See [comment](https://github.com/koding/koding/pull/10101#issuecomment-268965682)

## Description
Following test waits for `Build` method then, it checks  `dc.stat` member which is set **after** Build method. This is done on separate goroutine thus, we cannot see `dc.stat` value just by checking `Build` method.

```go
c := dc.opts.Builder.Build(ctx, a)

// Update current address.
*addr = a

dc.mu.Lock()
if dc.cancel != nil {
	dc.cancel()
}
dc.c, dc.stat, dc.ctx, dc.cancel = c, stat, ctx, cancel
dc.mu.Unlock()
```

## Motivation and Context
False negative test results.

### DIFF
--- a/go/src/koding/klient/machine/dynamic.go
+++ b/go/src/koding/klient/machine/dynamic.go
@@ -126,19 +126,23 @@ func (dc *DynamicClient) Client() Client {
 }
 
 // Context returns current client's context. If client change, returned context
-// will be canceled. If there is no clients available in dynamic client. Already
-// canceled context is returned.
-func (dc *DynamicClient) Context() context.Context {
+// will be canceled. If there is no clients available in dynamic client. New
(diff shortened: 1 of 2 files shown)


cihangir (approved):

rjeczalik:
I was thinking what about adding WaitForStatus with some small interval? This would block & poll on status, so we ensure the status evetually reaches desired state.

ppknap:
@rjeczalik status is changed along with context so we can use WaitForContextClose as I did now. This makes these test deterministic.

@cihangir ptal again since I changed the solution.

rjeczalik (approved):

PR #10108, opened 2016-12-23
cleanup | workers: remove unused workers

cihangir:


config/main.default.coffee
```diff
@@ -41,7 +41,6 @@ Configuration = (options = {}) ->
   options.sendEventsToSegment = yes
   options.scheme = 'http'
   options.suppressLogs = no
-  options.paymentBlockDuration = 2 * 60 * 1000 # 2 minutes
```
cihangir: cc/ @usirin

usirin: 👍  

I was using this for blocking people who tried to enter a failing card more than 3 times. But since we have a auth system in backend now, i don't think this will be necessary.

gokmen (approved):

PR #10111, opened 2016-12-23
klient: do not check for kite.key

rjeczalik:
Fixes the following error while connecting a VM to koding (on sandbox / dev env):

```
An existing kite.key for http://szkl.koding.team detected. Type 'yes' to override and continue: yes
Authenticated successfully
+ sudo '[' '!' -f /etc/kite/kite.key ']'
+ echo 'Error: Critical component missing. Aborting installation.'
Error: Critical component missing. Aborting installation.
+ exit -1
```

Btw we should get rid of connected vm and use kd instead.

/cc @szkl

### DIFF
--- a/go/src/koding/klient/install.sh
+++ b/go/src/koding/klient/install.sh
@@ -421,14 +421,7 @@ EOF
 	fi
 
 
-	if sudo [ ! -f /etc/kite/kite.key ]; then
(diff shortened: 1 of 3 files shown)


cihangir (approved):
SGTM

rjeczalik:
@ppknap ping

ppknap (approved):

PR #10122, opened 2016-12-27
more explicit, non-dev error messaging.

devrim:
customer facing messages should conform to https://en.wikipedia.org/wiki/Basic_English as much as possible due to international nature of our product.

### DIFF
--- a/go/src/koding/kites/kloud/provider/aws/aws.go
+++ b/go/src/koding/kites/kloud/provider/aws/aws.go
@@ -202,10 +202,10 @@ func (meta *Cred) Valid() error {
 		return errors.New("aws: secret key is empty")
 	}
 	if !reAccessKey.MatchString(meta.AccessKey) {
-		return errors.New("aws: access key is ill-formed")
+		return errors.New("aws: access key is invalid")
 	}
 	if !reSecretKey.MatchString(meta.SecretKey) {
-		return errors.New("aws: secret key is ill-formed")
+		return errors.New("aws: secret key is invalid")
 	}
 	return nil
 }
--- a/go/src/koding/kites/kloud/provider/google/google.go
+++ b/go/src/koding/kites/kloud/provider/google/google.go
@@ -91,7 +91,7 @@ func (c *Cred) Valid() error {
 	}{}
 
 	if err := json.Unmarshal([]byte(c.Credentials), &acckey); err != nil {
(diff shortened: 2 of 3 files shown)


go/src/koding/kites/kloud/provider/aws/aws.go
```diff
@@ -202,10 +202,10 @@ func (meta *Cred) Valid() error {
 		return errors.New("aws: secret key is empty")
 	}
 	if !reAccessKey.MatchString(meta.AccessKey) {
-		return errors.New("aws: access key is ill-formed")
+		return errors.New("aws: access key seems wrong")
```
rjeczalik: I think `seems wrong` suggests we're not entirely sure they key is bad - what about `is invalid`?

rjeczalik: Or `is wrong`.

devrim: i thought that was the intention (we don't know if it's wrong or not) - sure, "is invalid" is good.

ppknap: > i thought that was the intention (we don't know if it's wrong or not)

Actually, If this sentence was `access key format seems wrong`, It would be correct, since Amazon can change their key formats anytime.

rjeczalik: @ppknap Two things:

1) the implementation does not consider, neither for aws nor any other provider, that checks are fragile or can be invalidated - otherwise those would be only warnings and we would not be aborting credential verification if the check fail; moreover did any of the format actually changed till now?
2) wording for such error both for aws and google uses the word `ill-formed` - this issue is about using simpler language not about changing the meaning of it

I'll change `ill-formed` to `invalid` for all the provider after #10127 is merged.

ppknap: > this issue is about using simpler language not about changing the meaning of it.

I know and that's why I left my comment: `ill-formed` suggests that the format of provided credential is invalid, however `invalid` may suggest that we actually checked the credentials (by calling AWS API) and we were denied.

I'm fine with `s\ill-formed\invalid`, just think that the word `format` was important here.

cihangir (changes requested):
rebase needed.

rjeczalik (approved):

cihangir (approved):

PR #10123, opened 2016-12-27
aws-instance right under aws_instance is confusing

devrim:
copy change


gokmen:
@devrim all other provider instances named with their provider name, `my-vm` not fits correctly.

PR #10129, opened 2016-12-27
Events email

### LINKED ISSUE #10125: Make sure events are sent with user.email
So that we can help our users with onboarding emails, sending them tips and tricks as they attempt to build stacks, add cloud keys etc.

probably a small fix to this file:
https://github.com/koding/koding/blob/master/workers/social/lib/social/models/tracker.coffee

image

mehmetalisavas:
check user email in events & add if it doesn't exist

fixes https://github.com/koding/koding/issues/10125

### DIFF
--- a/go/src/socialapi/workers/email/emailsender/emailsender.go
+++ b/go/src/socialapi/workers/email/emailsender/emailsender.go
@@ -56,20 +56,16 @@ func Send(m *Mail) error {
 // and sends the message according to the mail adress
 // its a helper method to send message
 func (c *Controller) Process(m *Mail) error {
-	var to = m.To
-
-	if c.forcedRecipientEmail != "" {
-		to = c.forcedRecipientEmail
-	}
-
-	user := &eventexporter.User{Email: to}
 	if m.Properties == nil {
 		m.Properties = NewProperties()
 	}
 
-	user.Username = m.Properties.Username
-	if c.forcedRecipientUsername != "" {
-		user.Username = c.forcedRecipientUsername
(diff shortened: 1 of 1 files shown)


go/src/socialapi/workers/email/emailsender/emailsender.go
```diff
@@ -56,20 +59,16 @@ func Send(m *Mail) error {
...
 		m.Properties = NewProperties()
 	}
 
-	user.Username = m.Properties.Username
-	if c.forcedRecipientUsername != "" {
-		user.Username = c.forcedRecipientUsername
+	user, err := c.getUserInfo(m)
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			c.log.Error("could not determine the user info for %+v, skipping this event", m)
```
cihangir: it seems you forgot `return nil ` here

go/src/socialapi/workers/email/emailsender/emailsender.go
```diff
@@ -97,6 +94,30 @@ func (c *Controller) Process(m *Mail) error {
...
+			return nil, err
+		}
+
+		m.To = u.Email
+	}
+
+	user := &eventexporter.User{Email: m.To}
+
+	if c.forcedRecipientEmail != "" {
+		m.To = c.forcedRecipientEmail
```
cihangir: you should set user.Email instead m.To

cihangir (changes requested):
user.Email should be fixed.

cihangir (approved):

PR #10130, opened 2016-12-27
Algolia account deleter

mehmetalisavas:
Clean up algolia data that belongs to non existing accounts in mongo


mehmetalisavas:
@cihangir PTAL

cihangir:
you already deleted all the non existing accounts right?

cihangir (approved):
there are some things to comment in this PR but it is one-off script, we should clean them all at some point ~ 3-4 months later?

mehmetalisavas:
@cihangir we've cleaned existing accounts and indexes on algolia.

PR #9796, opened 2016-11-28
klient: add machine group

ppknap:
This PR adds `machinegroup` package that handles operations on remote machine groups.

Depends on: #9793 #9794 #9795 

## Description
This is reworked architecture that is going to handle all `kd machine *` CLI commands. It *partially* deprecates `klient/remote` packages. There is a number of current issues which these changes are intended to solve:

- Support for multiple mounts per machine.
- Correct handling of destroyed machines.
- Correct handling of machines that changed their IP address.
- Correct handling of machines that changed their Kite.key.
- Correct handling of rsync mount processes.
- Clean ups after errors.
- Prevent machine duplication during `kd machine list` 
- Businnes logic and Kite transort separation.
- etc.

## Architecture
image

- **Machine Group** - object that can manage multiple remote machines.
- **Kite Handlers** - a set of factory methods that binds Kite server to `MachineGroup` instance.

## How Has This Been Tested?
Unit tests

## Screenshots (if appropriate):
none

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #10133, opened 2016-12-28
klient/mount: abstract mount types

ppknap:
This PR adds `mount` abstract types.

Depends on: -

## Description

 - etc.

## Architecture
image

## How Has This Been Tested?
Only data structures - nothing to test.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/mount.go
+++ b/go/src/koding/klient/machine/mount/mount.go
@@ -1,0 +1,10 @@
+package mount
+
+// ID is a unique identifier of the mount.
+type ID string
+
+// Mount stores information about a single local to remote machine mount.
+type Mount struct {
+	Path  string `json:"path"`  // Mount point.
+	RPath string `json:"rpath"` // Remote directory path.
+}
--- a/go/src/koding/klient/machine/mount/mount.go
+++ b/go/src/koding/klient/machine/mount/mount.go
@@ -5,6 +5,6 @@ type ID string
 
 // Mount stores information about a single local to remote machine mount.
 type Mount struct {
-	Path  string `json:"path"`  // Mount point.
-	RPath string `json:"rpath"` // Remote directory path.
+	Path       string `json:"path"`       // Mount point.
+	RemotePath string `json:"remotePath"` // Remote directory path.
 }


go/src/koding/klient/machine/mount/mount.go
```diff
@@ -0,0 +1,10 @@
+package mount
+
+// ID is a unique identifier of the mount.
+type ID string
+
+// Mount stores information about a single local to remote machine mount.
+type Mount struct {
+	Path  string `json:"path"`  // Mount point.
+	RPath string `json:"rpath"` // Remote directory path.
```
rjeczalik: When I see `rpath` I think more about [this](https://en.wikipedia.org/wiki/Rpath) than remote path. What about just `RemotePath string`?

cihangir:
- [ ]  What is `Machine Group`
- [ ] Are we planning to get the index first from remote machine than the files' contents itself?
- [ ] could you elaborate the interaction between remote->local & local->remote changes?

cihangir (approved):

ppknap:
>  What is `Machine Group`

All components not described here are already [implemented](https://github.com/koding/koding/pull/9796).

> Are we planning to get the index first from remote machine than the files' contents itself?

The workflow is:
- user on local machine creates a mount.
- local klient calls remote one.
- remote klient generates index cache and sends the index back to local.
- local kd:
   - informs the user about the size and number of files to be fetched.
   - based on index data creates FUSE mount.
   - starts RSync process which is intended to fetch the actual content. 

I want to fetch index first because it's basically the only thing I need to create FUSE mount(actually the directory view of not existing files that I can "lazy" fetch when requested(future performance improvements))

> could you elaborate the interaction between remote->local & local->remote changes?

This will depend on Notifiers strategy. For now I'm working on two:

- `FUSE` - it's obviously local->remote change notifier. When you interact with FUSE mounted dir, it will generate index.Change "event" and send it to Sync object(which is responsible for throttling duplicated changes and sending them to syncers). Sync will also maintain priority queue(used as producer for Syncer workers) and ensure that large files won't block synchronization. FUSE changes will have higher priority than these produced by other Notifiers -> the local changes is going to override these made on remote machine.

- `Pooler` - it's responsible for remote->local and partially local->local notifications. It will be asynchronous worker that(by comparing remote and local indexes) tries as much as it can to make remote and local indexes identical(or at least make their entropy as small as possible). It will occasionally compare local index with local directory (partial local->local notifications) to catch events not correctly handled by FUSE. However, this most likely will be moved to separate type that satisfies Notifier interface(just to KISS).

My initial goal is to create very extensible POC which will left us a window for numbers of improvements. There are too many use cases and I can't cover them all. That's why I'm using incremental approach.

rjeczalik (approved):

PR #10134, opened 2016-12-28
klient/mount: machine mount index

ppknap:
This PR adds index type.

:warning: NOTE: this version is **NOT** optimized. 

Depends on: -

**Code coverage**: 79.7% of statements

## Description

 - etc.

## Architecture
image

## How Has This Been Tested?
Unit tests.

none

## Types of changes


go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+// Count returns the number of entries stored in index. Only items which size is
+// below provided value are counted. If provided argument is negative, this
+// function will return the number of all entries.
+func (idx *Index) Count(maxsize int64) (count int) {
+	idx.mu.RLock()
+	defer idx.mu.RUnlock()
+
+	if maxsize < 0 {
+		return len(idx.entries)
+	} else if maxsize == 0 {
```
rjeczalik: question: Why do we treat `0` as nop?

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+	idx.mu.RLock()
+	defer idx.mu.RUnlock()
+
+	return json.Marshal(idx.entries)
+}
+
+// UnmarshalJSON satisfies json.Unmarshaler interface. It is used to unmarshal
+// data into private index fields.
+func (idx *Index) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &idx.entries)
```
rjeczalik: nit: It should take lock for consistency with MarshalJSON.

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+func (cs ChangeSlice) Swap(i, j int)      { cs[i], cs[j] = cs[j], cs[i] }
+func (cs ChangeSlice) Less(i, j int) bool { return cs[i].Name < cs[j].Name }
+
+// Index stores a virtual working tree state. It recursively records objects in
+// a given root path and allows to efficiently detect changes on it.
+type Index struct {
+	mu      sync.RWMutex
+	entries map[string]*Entry
+}
+
```
rjeczalik: nit: You can show here explicitly that Index implements `json.Marshaler`/`json.Unmarshaler`.

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+// ChangeMeta indicates what change has been done on a given file.
+type ChangeMeta uint32
+
+const (
+	ChangeMetaUpdate ChangeMeta = 1 << iota // File was updated.
+	ChangeMetaRemove                        // File was removed.
+	ChangeMetaAdd                           // File was added.
+
+	ChangeMetaLarge ChangeMeta = 1 << (8 + iota) // File size is above 4GB.
```
rjeczalik: Is there any reason this is separate from the ChangeMeta group?

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+	CTime int64       `json:"ctime"` // Metadata change time since EPOCH.
+	MTime int64       `json:"mtime"` // File data change time since EPOCH.
+	Mode  os.FileMode `json:"mode"`  // File mode and permission bits.
+	Size  int64       `json:"size"`  // Size of the file.
+	SHA1  []byte      `json:"sha1"`  // SHA-1 hash of file content.
+}
+
+// NewEntryFile creates new Entry from a file stored under path argument.
+// Info is optional and, if given, should store LStat result on the given file.
+func NewEntryFile(root, path string, info os.FileInfo) (e *Entry, err error) {
```
rjeczalik: From what I understand:

- we store relative paths, relative to the mount dir, so it's possible to use the same index for two mounts with different mount directories; the mount dir here is root (can we add a comment for this argument?)
- we store paths converted to Unix paths, so it'd possible to use window index on linux

Is it correct?

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+// ChangeSlice stores multiple changes.
+type ChangeSlice []Change
+
+func (cs ChangeSlice) Len() int           { return len(cs) }
+func (cs ChangeSlice) Swap(i, j int)      { cs[i], cs[j] = cs[j], cs[i] }
+func (cs ChangeSlice) Less(i, j int) bool { return cs[i].Name < cs[j].Name }
+
+// Index stores a virtual working tree state. It recursively records objects in
+// a given root path and allows to efficiently detect changes on it.
+type Index struct {
```
rjeczalik: question: Is it possible for a single index to index files from different mounts? I'm trying to find out whether it would be feasible to make`root` a field of index instead of passing it to most functions as an argument.

rjeczalik:
I really your diagrams and designs, makes reading PR much more pleasurable.

I have 2 questions if you wouldn't mind:

> NOTE: this version is NOT optimized.

1) What optimisations do you have in mind? Just curious.
2) How we would handle corrupted mount dir? E.g. a directory tree on remote has one directory with no access rights so it's not possible to index it and user wants to mount it locally - do we want to show it's not possible to have a full mount due to access problems? Right now we ignore all errors.

go/src/koding/klient/machine/mount/index/index.go
ppknap: because we know that that there are no files(on UNIX) with `size == 0` this is a small optimization made to not iterate over the index.

rjeczalik: I meant `0` and `< 0` arguments are treated as `all` in stdlib. Tried to understand why we treat `-1` and `0` differently here.

go/src/koding/klient/machine/mount/index/index.go
ppknap: I made this for future usage. I plan to add more change types like `ChangeMetaChmod` `ChangeMetaMovedFrom/To`. This will allow to select better sync strategy based on observed operation. Eg. RSync is not very good with moves, so `Sync` can use KD/Kite(remotely) & mv(locally) instead.

go/src/koding/klient/machine/mount/index/index.go
ppknap: @rjeczalik you are correct. `root` is different on remote machine than on local one. And yes, the plan is to make possible to create mounts between Windows <--> Linux. 

 I took `root` name from https://golang.org/pkg/path/filepath/#Walk

go/src/koding/klient/machine/mount/index/index.go
ppknap: A: Everything is possible:). But, do we want one index for all mounts? It is guarded by mutex so eg. unmounting will stop all other mounts making it slow.

Other reason is that when index get broken we can just drop it and recreate. It's better to make this for a single mount rather than all of them(some of other ones might be even unreachable at that moment - machine is powered off).

rjeczalik: I don't want to store multiple mounts in a single index. Can we make a root member of index?

ppknap:
> What optimisations do you have in mind? Just curious.

- when scanning the directory for changes we don't need to check directory when its mtime didn't change. (idea is taken from git index - however, not all operating systems update mtime of directory when file inside it changes). This can greatly reduce the number of files we need to stat.
- binary format of index/compression - will reduce the amount of data we need to transfer over the network.
- custom `filepath.Walk` function - we don't need to walk directories in lexical order.
- if we have two indexes to sync. we can send just SHA-1 of the entry, if second index doesn't have it, this means that we need to sync it. This gives 200Kb over network for index that manages 10000files.

I'm going to check more ways for opt after POC.

>  How we would handle corrupted mount dir? E.g. a directory tree on remote has one directory with no access rights so it's not possible to index it and user wants to mount it locally - do we want to show it's not possible to have a full mount due to access problems? Right now we ignore all errors.

All corrupted indexes can be dropped and recreated without harm to the actual mount. If there is a directory with no access rights it won't be added to the index. If someone change its permissions on the remote machine and it become accessible, Pooler sync will eventually fetch the changes to local machine.

rjeczalik:
> binary format of index/compression - will reduce the amount of data we need to transfer over the network.

We should introduce msgpack2 to koding/kite instead and agree on used encoding and transport during handshake (https://github.com/koding/kite/issues/158).

go/src/koding/klient/machine/mount/index/index.go
ppknap: I took the idea from strings(regex) packages eg. https://golang.org/pkg/strings/#Replace where only negative numbers are treated as `all`. If this is a case I can change it.

rjeczalik: If a nop would be handy (e.g. no need for conditional Count) then let it be. Did not think of it as an optimisation.

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+	// Get relative file name.
+	name, err := filepath.Rel(root, path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Compute file's SHA-1 sum.
+	var sum []byte
+	if !info.IsDir() {
+		if sum, err = readSHA1(path); err != nil {
```
cihangir: lets use crc32 ( i have tried with IEEETable ) here instead of sha1, it is way faster than any of crc64 ( 6x ) , sha1( 11x), md5 (11x), sha256 (22x) both in big files & small files

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -0,0 +1,328 @@
...
+	return ChangeMetaLarge
+}
+
+// Apply modifies index according to provided changes. This function doesn't
+// guarantee that changes from Compare function applied to the index will
+// result in actual directory state.
+func (idx *Index) Apply(root string, cs ChangeSlice) {
+	for i := range cs {
+		switch {
+		case cs[i].Meta&(ChangeMetaUpdate|ChangeMetaAdd) != 0:
```
cihangir: could you explain the cases as comments

go/src/koding/klient/machine/mount/index/index_std.go
```diff
@@ -0,0 +1,14 @@
+// +build !darwin,!linux
```
cihangir: it seems this file is not needed, if fi.Sys returns nil, linux implementation will result with the same, if sometime in the future Sys is implemented any other OS, it will work without requiring any code change

rjeczalik: This file stubs `ctimeFromSys` for platforms other than darwin/linux (e.g. Windows). Just to make it possible to build the index package.

cihangir: ok i missed index_darwin file, what keeps us from implementing for others or using a package for that ( i remember there was a package just for unifying them )

cihangir: found https://github.com/djherbis/times

cihangir (changes requested):
Could you update the description with the coverage results?

go/src/koding/klient/machine/mount/index/index_std.go
ppknap: @cihangir thanks!

go/src/koding/klient/machine/mount/index/index.go
ppknap: Two identical indexes can use different roots. That's why I didn't add `Root` field to Index struct. Eg.
you're sending Index from remote machine to local. Root value from remote will be (almost) always incorrect on local.

go/src/koding/klient/machine/mount/index/index.go
ppknap: I'd like to keep SHA-1 here because I have some interest in keeping hashes consistent with these stored in git index(which means that I need SHA-1).

I have (long-term) plan to add another `Syncer` interface implementation. Lets call it `GitSync` that can detect git repository in mounted folder and improve git operations on it. Eg.

User runs `git checkout my_other_branch` on mounted directory. This can create a lot of FUSE changes and require as much of Rsync syncer work. Now, assuming we have `GitSync` that can examine git index/reflog/etc. and detect eg. checkout. Thus, it will be able to just run `git checkout mu_other_branch` on remote machine without copying anything.

These are just plans but, if we ever start implementing this, it would be nice to have SHA-1 sums in order to perform sanity checks on git index.

ppknap:
@cihangir Done

go/src/koding/klient/machine/mount/index/index.go
rjeczalik: Oh, got it. I assumed remote paths are already relative, so it's easier to work with index locally if the side owning an index presents paths in relative form.

go/src/koding/klient/machine/mount/index/index.go
cihangir: you are calculating the hash just from the file content, git hashing uses other kind of info ( commit message, author, committer, previous hash etc ), how are you planning to keep them insync? or from another angle, git has history of commit orders, this implementation of index does not have any kind of ordering/sorting. Another point for git indexes would be, files could be in the folder (uncommitted) without a hash value, but in any case you need to keep them in the index for various reasons.

I dont have a strong opinion at this moment since we will see what is waiting for us in the future, i just wanted to give you a speed bump of 10x :)

cihangir (approved):

go/src/koding/klient/machine/mount/index/index.go
ppknap: > I dont have a strong opinion at this moment since we will see what is waiting for us in the future..

Exactly, I have the same feeling and I don't mind changing this to CRC32 when SHA-1 turns out to be useless :)

> you are calculating the hash just from the file content, git hashing uses other kind of info ( commit message, author, committer, previous hash etc ),

You are partially right I'm speaking of git index itself which stores SHA-1 of file [content](http://stackoverflow.com/questions/4084921/what-does-the-git-index-contain-exactly/25806452#25806452)

```
20 byte SHA-1 over the previous content of the entry. Note that according to my experiments with the assume valid flag, the flags that follow it are not considered in this SHA-1.
```

rjeczalik (approved):

Issue #9988, opened 2016-12-10
kd: add `kd machine addresses` command.


ppknap:
The goal of this issue is to add `kd machine addresses` command which will show all machine addresses known to `kd`.

Currently, `kd machine list` IP column shows tunnel address for machines that don't have public IP. Obviously, IP != tunnel address. Thus, the information in IP column is wrong. 

Proposed solution is to replace tunnel address with `<tunnel>` string and get access to it by `kd machine addresses` command. Klient will store such addresses in `Addr` structure with `tunnel` network.

PR #10141, opened 2016-12-28
kloud: kd machine list - add support for managed machines.

ppknap:
Machine.GeneratedFrom field is `null` for managed machines. This causes `kd machine ls` fail for them, since we try to bson-unmarshal nil pointer to non-pointer field:

```sh
$ kd machine ls
Error communicating with Koding: ObjectIDs must be exactly 12 bytes long (got 0) (329309e5469ed4b2)
error executing "list" command: ObjectIDs must be exactly 12 bytes long (got 0) (329309e5469ed4b2)
```

After the fix:
```sh
kd machine ls
ID                        ALIAS         TEAM  STACK                     PROVIDER  LABEL             OWNER  AGE     IP                                         STATUS
584ffc262586bbae18308d44  olive_raisin  ssh   Alakazam Aws Stack        aws       aws-instance             15d5h   35.156.209.39                              online (15d5h)
584ffd312586bbae18308d4b  fuchsia_kiwi  ssh   Aerodactyl Vagrant Stack  vagrant   vagrant-instance         15d5h   upsvfffe9f69.pawel.1d543013.dev.koding.me  offline (7d23h: Machine is marked as Stopped)
58640b4f569ccef36b598315  black_banana  -     -                         managed   archpk                   18m56s  c1feb3e08536.pawel.1d543013.dev.koding.me  <unknown> (-)
```

This PR also fixes false negative race condition in machine dynamic client tests.

## Motivation and Context
Bug fix in db/models

## How Has This Been Tested?
Manually

## Types of changes

### DIFF
--- a/go/src/koding/kites/kloud/machine/mongo.go
+++ b/go/src/koding/kites/kloud/machine/mongo.go
@@ -107,19 +107,10 @@ func (m *MongoDatabase) Machines(f *Filter) ([]*Machine, error) {
 		groupTitles[st.Id] = [2]string{st.Group, st.Title}
 	}
 
-	var team, stack string
 	machines := make([]*Machine, len(machinesDB))
 	for i, mdb := range machinesDB {
-		team, stack = "", ""
-		if mdb.GeneratedFrom != nil {
-			team = groupTitles[mdb.GeneratedFrom.TemplateId][0]
-			stack = groupTitles[mdb.GeneratedFrom.TemplateId][1]
-		}
-
 		machines[i] = &Machine{
 			ID:          mdb.ObjectId.Hex(),
-			Team:        team,
-			Stack:       stack,
 			Provider:    mdb.Provider,
 			Label:       mdb.Label,
 			IP:          hostOnly(mdb.IpAddress),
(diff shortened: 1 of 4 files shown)


go/src/koding/kites/kloud/machine/mongo.go
```diff
@@ -105,12 +107,19 @@ func (m *MongoDatabase) Machines(f *Filter) ([]*Machine, error) {
...
+			team = groupTitles[mdb.GeneratedFrom.TemplateId][0]
+			stack = groupTitles[mdb.GeneratedFrom.TemplateId][1]
+		}
+
 		machines[i] = &Machine{
 			ID:          mdb.ObjectId.Hex(),
-			Team:        groupTitles[mdb.GeneratedFrom.TemplateId][0],
-			Stack:       groupTitles[mdb.GeneratedFrom.TemplateId][1],
+			Team:        team,
+			Stack:       stack,
```
rjeczalik: Nit: You could assign to machine[i].Stack directly without using temporary vars.

rjeczalik (approved):

go/src/koding/kites/kloud/machine/mongo.go
ppknap: Done.

Thanks @rjeczalik , nice catch!

gokmen (approved):
that field is only for machines which generated from a `stackTemplate` so 👍

cihangir:
Not about this PR:
What about changing `IP` to `Address` in the output of ls command? There are tunnel urls of vagrant & managed vms. Address might cover both cases

cihangir (approved):

ppknap:
@cihangir yes that's another good idea among others: #9988 . We will need to do something with this because tunnel address obviously is not of IP form.

PR #10155, opened 2016-12-29
go/makefile: add cover to test makefile

mehmetalisavas:
coverage for tests
- [x] write a new script for test files to test with coverage 
- [x] bash script should be able to take parameters for test directory & config directory
- [x] re-write socialapi/makefile with coverage & bash script
- [x] all coverages should be reported


go/src/socialapi/Makefile
```diff
@@ -99,7 +99,7 @@ testcollaboration:
 
 testmodels:
 	@echo "$(OK_COLOR)--> model tests... $(NO_COLOR)"
-	@`which go` test -c models/*.go
+	@`which go` test -race -coverprofile=profile.out -covermode=atomic -c models/*.go
```
rjeczalik: Would it be possible to use package names instead of file paths?

```
@`which go` test -race -coverprofile=profile.out -covermode=atomic -c socialapi/models
```

mehmetalisavas: yeap, working on it 👍

wercker.yml
```diff
@@ -89,7 +89,7 @@ build:
         code: go/build.sh
     - script:
         name: test go koding config
-        code: scripts/wercker/run-command ./run exec go test -v --race koding/kites/config
+        code: scripts/wercker/run-command ./run exec go test -v --race -coverprofile=profile.out -covermode=atomic koding/kites/config
```
cihangir: you could use `scripts/wercker/run-command /scripts/gotests.sh koding/kites/config`

scripts/gotests.sh
```diff
@@ -0,0 +1,57 @@
...
+
+function runAll () {
+    compile $@
+    run $@
+    clean $@
+}
+
+if [ "$1" == "socialapi" ]; then
+  shift
+  export RUN_FLAGS=${RUN_FLAGS:-"-c=./go/src/socialapi/config/dev.toml"}
```
cihangir: could you make `-race` default for all tests after we merge this PR?

mehmetalisavas:
ping

szkl (approved):

scripts/gotests.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
```
rjeczalik: How about adding `set -euo pipefail`? So e.g. if `compile` at L37 fails we do not `run` binary, which may be stale.

rjeczalik (approved):

PR #10180, opened 2016-12-30
Speed up page loading - part 1

cihangir:
Various optimizations...


cihangir:
@gokmen this PR is getting bigger, i have removed some of the unused and unnecessary paths from initial page load. We can check this in and i will send separate prs for other stuff

servers/lib/server/helpers.coffee
```diff
@@ -152,29 +151,13 @@ serveHome = (req, res, next) ->
 
 
 isLoggedIn = (req, res, callback) ->
-
-  { JName } = koding.models
-
   findUsernameFromSession req, res, (err, username) ->
+    return callback null, no, null  unless username
```
gokmen: `return callback err  if err` would be nice

gokmen (approved):
it LGTM but there are some minors, I'll merge this one can u update later? thanks!

PR #10181, opened 2016-12-30
Stack Editor: Logs view, editor improvements and stack create flow

gokmen:


gokmen:
fixes: #10184

gokmen:
fixes: #10186

client/app/lib/providers/computecontroller.coffee
```diff
@@ -37,7 +37,7 @@ require './config'
 module.exports = class ComputeController extends KDController
 
 
-  @providers = globals.config.providers._getSupportedProviders()
+  PROVIDERS  = globals.config.providers._getSupportedProviders()
```
usirin: [minor] maybe extracting this to a util?

client/new-stack-editor/lib/controllers/base.coffee
```diff
@@ -0,0 +1,10 @@
+kd = require 'kd'
+
+
+module.exports = class BaseController extends kd.Object
+
+  constructor: (options = {}, data) ->
+
+    super options, data
+
+    @editor = @getOption 'editor'
```
usirin: Right now this only creates just another indirection with any visible gain (`@getOption 'editor'` is always available without this base class), but i am guessing you are thinking about extending this one's functionality.

client/new-stack-editor/lib/index.coffee
```diff
@@ -36,27 +43,25 @@ module.exports = class StackEditorAppController extends AppController
     @stackEditor.on 'InitializeRequested', @bound 'initializeStack'
 
 
-  openEditor: (templateId) ->
+  openEditor: (templateId, reset = no) ->
```
usirin: Can we change the 2nd argument to be an object, that way, we will have to explicitly write `{ reset: no }`. IMO it improves its readability, and removes the need for checking the implementation to understand what `no` or `yes` is used for at 2nd arg.

client/new-stack-editor/lib/index.coffee
```diff
@@ -36,27 +43,25 @@ module.exports = class StackEditorAppController extends AppController
...
+    markAsLoaded()
 
 
-  reloadEditor: (templateId, skipDataUpdate) ->
+  reloadEditor: (templateId) ->
 
-    console.trace()
-    log '::reloadEditor', templateId, skipDataUpdate
+    delete @templates[templateId]
+    @openEditor templateId, reset = yes
```
usirin: aha, you already felt the same but applied another solution 🤔 
i liked it, but i am still in favor of using an object, to be less error prone.

client/new-stack-editor/lib/views/editor.coffee
```diff
@@ -18,15 +21,40 @@ module.exports = class Editor extends BaseView
...
+
   focus: -> @ready =>
     @aceView.ace.focus()
 
 
+  setReadOnly: (state) -> @ready =>
+
+    @aceView.ace.setReadOnly state, no
+    state = if state then 'none' else 'block'
+    @aceView.ace.editor.renderer.$cursorLayer.element.style.display = state
```
usirin: (titredim...) :D

client/new-stack-editor/lib/views/wizard/providerselectionview.coffee
```diff
@@ -0,0 +1,67 @@
...
+
+    @createProviders()
+
+
+  createProviders: ->
+
+    { providers } = globals.config
+
+    supportedProviders = (Object.keys providers).filter (p) ->
+      return p  if providers[p].supported
```
usirin:
```coffee
supportedProviders = (Object.keys providers).filter (p) -> providers[p].supported
```

actually with lodash there is an easier and -IMO- more readable way ([link to doc](https://lodash.com/docs/4.17.3#pickBy)):

```coffee
_.pickBy providers, (provider) -> provider.supported
```

All three of them do the same thing.

usirin: btw, i don't have any preferences, i just wanted to share my opinions. I hate doing `Object.keys ....` everytime.

client/new-stack-editor/lib/views/wizard/providerselectionview.coffee
```diff
@@ -0,0 +1,67 @@
...
+        click      : =>
+          return  if extraClass is 'coming-soon'
+
+          Tracker.track Tracker["STACKS_WIZARD_SELECTED_#{provider.toUpperCase()}"]
+
+          providerView.setClass 'selected'
+          @selected?.unsetClass 'selected'
+
+          @selected = if @selected is providerView then null else providerView
+          @emit 'SelectedProviderChanged', @selected
```
usirin: Recently i have started to extract out this kind of hardcoded values into a `constants.coffee` right next to the files that uses them. It's especially useful when one wants to change the value of the event, change in one place, and don't think about where it was being used.

client/new-stack-editor/lib/views/wizard/stackwizard.coffee
```diff
@@ -0,0 +1,51 @@
...
+
+
+  pistachio: ->
+
+    '''
+      {{> @providerSelectionView}}
+      <footer>
+        {{> @createButton}}
+        {{> @cancelButton}}
+      </footer>
```
usirin: Unnecessary indentation.

client/new-stack-editor/lib/views/wizard/stackwizardmodal.coffee
```diff
@@ -0,0 +1,44 @@
...
+      @on 'KDObjectWillBeDestroyed', -> kd.singletons.router.back()
+
+    @addSubView wizard
+
+
+  createStackTemplate: (selectedProvider) ->
+
+    { router } = kd.singletons
+
+    unless selectedProvider
```
usirin: Why and how do we need this case?

usirin (approved):
My comments are mostly minors/nitpicks. We can tackle those issues later. Merging.

client/new-stack-editor/lib/index.coffee
gokmen: I usually not prefer that if there is only one argument, if there will be more than I can update.

client/new-stack-editor/lib/views/editor.coffee
gokmen: this is the only way to hide cursor on Ace, it doesn't look good I agree. there are terrible solutions for that one believe me :) https://github.com/ajaxorg/ace/issues/266

client/new-stack-editor/lib/views/wizard/providerselectionview.coffee
gokmen: I agree but I usually prefer to use the way everybody can understand without a knowledge. TIL `pickBy` we can use it further.

client/new-stack-editor/lib/views/wizard/stackwizardmodal.coffee
gokmen: my bad, in the previous example there was a case for this one but it's not valid anymore.

client/app/lib/providers/computecontroller.coffee
gokmen: most of the time where I need supported providers list I also need providers as well so I kept it in the same object for consistency.

client/new-stack-editor/lib/controllers/base.coffee
gokmen: in this file I may re-set the option and delete editor from there after I've assigned it to local variable. so, there are some plans on `@editor` variable for the future.

PR #10188, opened 2016-12-31
Machine mount index cache

ppknap:
Cache is used to retrieve mounted directory index without recomputing it each time. 

Depends on: ~#10134~

**Code coverage**: 84.6% of statements.

## Description

 - etc.

## Architecture
image

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/index/cache.go
+++ b/go/src/koding/klient/machine/mount/index/cache.go
@@ -1,0 +1,172 @@
+package index
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const tempIndexDirPrefix = "koding_index_"
+
+// GetCachedIndex returns index that describes the current directory state. This
+// function stores the resulting index in temporary file in order to not
+// recompute it each time when the index is requested.
+func GetCachedIndex(root string) (*Index, error) {
+	var cs ChangeSlice
+
+	// Load or create index.
+	idx, path, err := getCachedIndex(root)
+	if err != nil {
+		// Generate new index.
+		if idx, err = NewIndexFiles(root); err != nil {
+			return nil, err
+		}
+	} else {
+		// Update loaded index.
+		cs = idx.Compare(root)
+		idx.Apply(root, cs)
+	}
+
+	// If index changed or was generated, save it.
+	if path == "" || len(cs) != 0 {
+		if path == "" {
+			if path, err = createTempPath(root); err != nil {
+				return nil, err
+			}
+		}
+
+		if err = cacheIndex(idx, path); err != nil {
+			return nil, err
+		}
+	}
+
+	return idx, nil
+}
+
+// HeadCachedIndex gets cached index or creates a new one and returns the number
+// and the overall size of stored files.
+func HeadCachedIndex(root string) (count int, diskSize int64, err error) {
+	idx, err := GetCachedIndex(root)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return idx.Count(-1), idx.DiskSize(-1), nil
+}
+
+// getCachedIndex looks up for index stored in one of temporary directories.
+// If provided index is found, it will be loaded to memory and returned.
+func getCachedIndex(root string) (idx *Index, path string, err error) {
+	name := hashSHA1(root)
+	// Find index in temporary directories.
+	for _, tempdir := range indexTempDirs(-1) {
+		path = filepath.Join(tempdir, name)
+		if _, err = os.Stat(path); err == nil {
+			break
+		}
+	}
+
+	if path == "" || err != nil {
+		return nil, "", errors.New("index file not found")
+	}
+
+	// Read index content.
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, "", err
+	}
+	defer f.Close()
+
+	idx = NewIndex()
+	if err = json.NewDecoder(f).Decode(idx); err != nil {
+		return nil, "", err
+	}
+
+	return idx, path, nil
+}
+
+// cacheIndex atomically saves the provided index under a given path. If the
+// path is empty, this function will create or use existing index temporary
+// directory.
+func cacheIndex(idx *Index, path string) (err error) {
+	dir, name := filepath.Split(path)
+	f, err := ioutil.TempFile(dir, name+"_")
+	if err != nil {
+		return err
+	}
+
+	if err := json.NewEncoder(f).Encode(idx); err == nil {
+		err = f.Sync()
+	}
+
+	if cerr := f.Close(); cerr != nil && err == nil {
+		err = cerr
+	}
+
+	if err == nil {
+		err = os.Rename(f.Name(), path)
+	}
+
+	if err != nil {
+		os.Remove(f.Name())
+	}
+
+	return err
+}
+
+// createTempPath creates a valid path to index file.
+func createTempPath(root string) (path string, err error) {
+	if dirs := indexTempDirs(1); len(dirs) > 0 {
+		path = dirs[0]
+	} else {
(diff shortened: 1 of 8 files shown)


go/src/koding/klient/machine/mount/index/cache.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return err
+}
+
+// createTempPath creates a valid path to index file.
+func createTempPath(root string) (path string, err error) {
+	if dirs := indexTempDirs(1); len(dirs) > 0 {
+		path = dirs[0]
+	} else {
+		if path, err = ioutil.TempDir("", tempIndexDirPrefix); err != nil {
```
rjeczalik: Do we have guarantee /tmp is persistent on Linux? If it's tmpfs, then it's going to be wiped on reboot. What happens when index file is removed while klient is running with active mounts?

What about ~/.cache/koding, supplementary to ~/.config/koding?

go/src/koding/klient/machine/mount/index/index.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/djherbis/times"
 )
 
+// Version stores current version of index.
+const Version = 1
```
rjeczalik: Is it going to be manually bumped or set during build with ldflags? If the latter, then it should be var.

go/src/koding/klient/machine/mount/index/index_unix_test.go
```diff
@@ -0,0 +1,12 @@
+// +build !windows
```
rjeczalik: 👍

go/src/koding/klient/machine/mount/index/cache.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+
+	return dirs
+}
+
+// hashSHA1 converts a string to hex representation of its SHA-1 checksum.
+func hashSHA1(val string) string {
+	h := sha1.New()
+	io.WriteString(h, val)
```
rjeczalik:
```go
h := sha1.Sum([]byte(val))
return hex.hex.EncodeToString(h[:])
```

You can spare one line 😎

go/src/koding/klient/machine/mount/index/cache_test.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		t.Fatalf("want err = nil; got %v", err)
+	}
+
+	// Check index integrity.
+	count, diskSize, err := HeadCachedIndex(root)
+	if err != nil {
+		t.Fatalf("want err = nil; got %v", err)
+	}
+	if count != idx.Count(-1) {
```
rjeczalik: `if n := idx.Count(-1); n != count {`

If `Count` is buggy you have no guarantee second invocation of it will return the same value as the first one.

go/src/koding/klient/machine/mount/index/cache.go
ppknap: > Do we have guarantee /tmp is persistent on Linux? If it's tmpfs, then it's going to be wiped on reboot. What happens when index file is removed while klient is running with active mounts?

We don't care at all, if index is removed while running on active mount it will be regenerated. The cache on remote machine is only to speed things up.

go/src/koding/klient/machine/mount/index/index.go
ppknap: I'm not going to set this on build time nor should we - this won't be changed along with kd version. I added this field only to provide a way to support incompatible changes in the future(git's index is versioned in the same way).

rjeczalik (approved):

rjeczalik:
Really liked this PR, especially the index tests. 💯

go/src/koding/klient/machine/mount/index/cache.go
gokmen: if we are going to keep using `root` user for `kd` `/tmp` is the best choice then.

go/src/koding/klient/machine/mount/index/cache.go
```diff
@@ -0,0 +1,170 @@
...
+	if cerr := f.Close(); cerr != nil && err == nil {
+		err = cerr
+	}
+
+	if err == nil {
+		err = os.Rename(f.Name(), path)
+	}
+
+	if err != nil {
+		os.Remove(f.Name())
```
cihangir: 👍

go/src/koding/klient/machine/mount/index/cache.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil
+	}
+	defer d.Close()
+
+	if n <= 0 {
+		n = -1
+	}
+
+	var dirs []string
+	for len(dirs) != n {
```
cihangir: if there are 101 items in the folder this will loop infinitely

go/src/koding/klient/machine/mount/index/cache.go
```diff
@@ -0,0 +1,170 @@
...
+		if path, err = ioutil.TempDir("", tempIndexDirPrefix); err != nil {
+			return "", err
+		}
+	}
+
+	return filepath.Join(path, hashSHA1(root)), nil
+}
+
+// indexTempDirs reads all directory names that could be created by index cache.
+func indexTempDirs(n int) []string {
```
cihangir: could you write test for this function?

go/src/koding/klient/machine/mount/index/cache_test.go
```diff
@@ -0,0 +1,140 @@
...
+				return nil
+			}
+
+			idxs = append(idxs, path)
+		}
+
+		return nil
+	}
+
+	if err := filepath.Walk(os.TempDir(), walkFn); err != nil {
```
cihangir: could you change the structure here to use one specific folder per test run and clean up only that one? If we start running tests in parallel they might remove other test's folders.

cihangir (changes requested):

go/src/koding/klient/machine/mount/index/cache.go
rjeczalik: It won't, we'll break the loop at line 159 on second iteration with err being io.EOF.

cihangir: I didnt know readdirnames was keeping track of the offset, it is not written anywhere in the docs too(at least i couldnt find). But it is https://play.golang.org/p/yU7UOLzjAK TIL

rjeczalik: I also needed to lookup`os.Readdirnames` since the behaviour is not obvious at all, just by looking at implementation.

The comment could be improved:

```
// Limit the number of read directories. If no more directories are left to read, err is io.EOF.
```

go/src/koding/klient/machine/mount/index/cache_test.go
ppknap: Good point.

go/src/koding/klient/machine/mount/index/cache.go
ppknap: this function is already tested (indirectly). I'm usually try to test only public interfaces.

go/src/koding/klient/machine/mount/index/cache.go
ppknap: Yeah, Go's official documentation is not very clear here. I took the info from https://golang.org/pkg/os/#File.Readdir and by directly looking at implementation.

ppknap:
@cihangir done

go/src/koding/klient/machine/mount/index/index_unix_test.go
ppknap: https://github.com/rjeczalik/notify/blob/master/sync_unix_test.go

:smile:

ppknap:
@cihangir I'm merging this now(need to fix #10220), If you request other changes, I will add them to new PR.

cihangir (approved):

PR #10190, opened 2017-01-01
social: add doc for JStackTemplate.create

rjeczalik:


workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -162,6 +162,26 @@ module.exports = class JStackTemplate extends Module
...
+  #
+  # @example api
+  #
+  #   {
+  #     "template": "{\"provider\": {\"aws\": {}}}",
+  #     "title": "My AWS template",
+  #     "credentials": {}
+  #   }
+  #
+  @create = ->
```
rjeczalik: /cc @gokemn

cihangir (approved):

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
gokmen: actually, this should work;

```
...
"template": {
  "provider": {
    "aws": {}
  }
},
...
```

gokmen (approved):

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
rjeczalik: I think it's not the same - the template field is expected to be a string, meaning it's stringified JSON. In your example template is going to be an object, isn't it?

rjeczalik: The requirement for template to be a stringified JSON is that jStackTemplate.RawContent field is a string.

gokmen: ah ok, you're right. I thought there was an issue with doc parser and you tried to fix that one. ignore my comment then.

PR #10193, opened 2017-01-02
servers: separate functionality out of helpers.coffee

cihangir:


sinan (approved):
💯  I am not sure about removing `verifytoken`, but I think you did your research.

cihangir:
@sinan i have searched usages of `/Verify/` endpoint but couldn't find any

PR #10206, opened 2017-01-04
Mount book - manage multiple mounts on remote machine. 

ppknap:
Mount book is used to store different mounts that synchronize between the same remote machine.

Depends on: -

**Code coverage**: 88.1% of statements

## Description

 - etc.

## Architecture
image

- **Mount Book** - group of mounts.

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/addr.go
+++ b/go/src/koding/klient/machine/addr.go
@@ -37,6 +37,11 @@ type AddrBook struct {
 	addrs []Addr
 }
 
+var (
+	_ json.Marshaler   = (*AddrBook)(nil)
+	_ json.Unmarshaler = (*AddrBook)(nil)
+)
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/addr.go
```diff
@@ -37,6 +37,11 @@ type AddrBook struct {
 	addrs []Addr
 }
 
+var (
+	_ json.Marshaler   = (*AddrBook)(nil)
+	_ json.Unmarshaler = (*AddrBook)(nil)
```
rjeczalik: 👍

rjeczalik (approved):

cihangir (approved):

PR #10207, opened 2017-01-04
Extra padding on top of each stack and teammates section on dashboard 

### LINKED ISSUE #10203: Extra padding on top of each stack section on dashboard
## Current Behavior
image

hakankaradis:
Even we have this code for `borderizeListItems`
https://github.com/koding/koding/blob/master/client/app/styl/commons/appfn.styl#L340
with the description of `:first-of-type` that represents the first element among siblings of its element type. But the first element's of the list `padding-top` wasn't set correctly. 
Strangely, it is working for some other components...

My first solution was adding a style for this class to generalize the solution https://github.com/koding/koding/blob/master/client/app/lib/components/list/index.coffee#L136
That cause other styling problems in other components because of some custom implementation.

So I had to customize the all elements in list but first.
Not sure that it is a correct solution.

Fixes #10203

### DIFF
--- a/client/home/lib/styl/homestacks.styl
+++ b/client/home/lib/styl/homestacks.styl
@@ -67,6 +67,11 @@ $borderColor                = rgba(151, 151, 151, .52)
 .ListView-section.HomeAppViewStackSection
   .ListView-row
     borderizeListItems      25px, 1px solid $borderColor
+  .ListView-row
+    padding-top             0px !important
+  .ListView-row ~ .ListView-row
+    padding-top             25px !important
+
 
 .HomeAppView--scroller.stacks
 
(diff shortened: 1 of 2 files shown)


sinan:
ℹ️ `first/last-of-type` don't consider classnames as types, they just work properly with the `<tagName>` so if there is another div in the same parent with a different class it would fail.

sinan:
i can't really say it is the right solution w/o seeing this live. If you tested thoroughly and think that this works and doesn't break anything else, we can take in.

hakankaradis:
I have checked all other pages, and my changes are specific for these components. They will not affect to other components. But tomorrow I will check it again and make sure it doesn't break anything and let you know.

hakankaradis:
@sinan I have checked the all other components, it looks good to me. I think it is ready to go.

sinan (approved):

PR #10219, opened 2017-01-05
scripts/api-generator: improvements for method parameters

rjeczalik:
A number of improvements for swagger.json:

- scripts/api-generator: make DefaultSelector a generic object

Setting properties of DefaultSelector to single field ("_id") makes
the DefaultSelector to be just a struct with single field, which is
useless in majority of cases.

- scripts/api-generator: make it possible to send payload for update methods

This PR adds generic bodyParam definition for each instance methods, so
it's possible to set a body for e.g. JStackTemplate.update request,
which currently is not possible.

### DIFF
--- a/scripts/api-generator.coffee
+++ b/scripts/api-generator.coffee
@@ -154,7 +154,7 @@ generateMethodPaths = (model, definitions, paths, docs) ->
   methods = model.getSharedMethods()
 
   schema = if definitions[name]
-  then { $ref: "#/definitions/#{name}" }
+  then { allOf: [ { $ref: "#/definitions/#{name}" }, { $ref: '#/definitions/DefaultResponse' } ] }
   else { $ref: '#/definitions/DefaultResponse' }
 
   for method, signatures of methods.statik
--- a/scripts/api-generator.coffee
+++ b/scripts/api-generator.coffee
@@ -194,7 +194,7 @@ generateMethodPaths = (model, definitions, paths, docs) ->
 
   for method, signatures of methods.instance
 
-    parameters = [ { $ref: '#/parameters/instanceParam' } ]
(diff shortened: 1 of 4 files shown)


scripts/api-generator.coffee
```diff
@@ -199,7 +194,7 @@ generateMethodPaths = (model, definitions, paths, docs) ->
 
   for method, signatures of methods.instance
 
-    parameters = [ { $ref: '#/parameters/instanceParam' } ]
+    parameters = [ { $ref: '#/parameters/instanceParam' }, { $ref: '#/parameters/bodyParam' } ]
```
rjeczalik: /cc @gokmen

scripts/api-generator.coffee
```diff
@@ -43,11 +43,6 @@ swagger =
           example: 'KodingError'
     DefaultSelector:
       type: 'object'
-      properties:
-        _id:
-          type: 'string'
-          description: 'Mongo Object ID'
-          example: '582c21d43bf248161538450b'
```
rjeczalik: /cc @gokmen

cihangir (approved):

rjeczalik:
One more:

- 	scripts/api-generator: make it possible to read error object

The remote.api endpoint on 200 OK returns either response object
if request was successful, or returns an error object to indicate
an error.

This makes it impossible to inspec error object for methods that
don't use DefaultResponse schema.

This PR makes it possible to read both - response object or
eventual error.

@cihangir Please take a look again.

rjeczalik:
blocked by https://github.com/go-swagger/go-swagger/issues/844

gokmen (approved):
💯

rjeczalik:
Tested, works when generating client with [fixed swagger](https://github.com/go-swagger/go-swagger/pull/845).

Next part of this PR - #10223.

@cihangir @gokmen ptal

gokmen (approved):

PR #10223, opened 2017-01-05
remoteapi: regenerate swagger client

rjeczalik:
This PR regenerates swagger client for remote.api after #10219.

In order to make it possible, firstly update go-openapi/* packages, since code generated by swagger from master has backward-incompatible changes.

Depends on #10219.


cihangir (approved):
merging this since the code is autogenerated

PR #10225, opened 2017-01-06
klient/mount: add remote index kite handlers

ppknap:
This PR adds kite handlers to client so the local klient can call remote one when it wants remote index data.

Depends on: #10188

## Motivation and Context
local -> remote machine sync

## How Has This Been Tested?
No test written - these are only handlers to already tested functions + simple path validation.

### DIFF
--- a/go/src/koding/klient/machine/mount/index/kite.go
+++ b/go/src/koding/klient/machine/mount/index/kite.go
@@ -1,0 +1,55 @@
+package index
+
+import (
+	"github.com/koding/kite"
+)
+
+// KiteHandlerHead creates a kite handler function that, when called, invokes
+// index package Head method.
+func KiteHandlerHead() kite.HandlerFunc {
+	return func(r *kite.Request) (interface{}, error) {
+		req := &Request{}
+
+		if r.Args != nil {
+			if err := r.Args.One().Unmarshal(req); err != nil {
+				return nil, err
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/mount/index/kite.go
```diff
@@ -0,0 +1,55 @@
+package index
+
+import (
+	"github.com/koding/kite"
+)
+
+// KiteHandlerHead creates a kite handler function that, when called, invokes
+// index package Head method.
+func KiteHandlerHead() kite.HandlerFunc {
```
rjeczalik: I'm thinking whether having single `kiteHandler` would be more economical (if we're going to have 10+ kite methods then probably so) in terms of loc, something like:

```go
func KiteHandlerHead() kite.HandlerFunc {
    return kiteHandler(func(*Request)(interface{}, error) { return Head() })
}
```

🤔

ppknap: It would be cleaner and I'm aware of this. However, I am planning to go even further and shorten other [kite.go](https://github.com/koding/koding/blob/master/go/src/koding/klient/machine/machinegroup/kite.go) file with the one generic handler "factory" for all functions.

Code coverage will make such changes more beneficial.

cihangir (approved):

rjeczalik (approved):

PR #10229, opened 2017-01-06
kd: add stack subcommand

rjeczalik:
This PR adds `stack` subcommand.

I'm going to give a rough overview of existing kd commands workflow, in order to get feedback and improve things when necessary.

### Login workflow

In order to create a stack user needs to be logged in into Koding and have a verified credential.

### Logging to Koding / kd.io

In order to log in to kd.io one needs to:

```
$ kd auth login
```

Currently logging in to kd.io does not work in dev environments, as there's no kd-io team created by default, use `--team` switch to log in to your team instead.

For the first time kd will ask for username / password in order to obtain kite.key from kontrol. Once kite.key is saved, consecutive `kd auth login --team x` commands will proceed without asking for password. 

In order to login to your dev koding use `--baseurl` flag:

```
$ kd auth login --baseurl http://rjeczalik.koding.team --team foobar
Logging to http://rjeczalik.koding.team ...
Successfully logged in to the following team: foobar
```

If you need to switch between kodings during development, use `kd config list` to list them:

```
$ kd config list
ID        KODING URL
10bc5796  https://sandbox.koding.com
1d1f3321  http://dev.koding.com:8090
79154da7  https://koding.com
824f84b3  http://rjeczalik.koding.team
```

And `kd config use` or `kd auth login --baseurl` to make the switch:

```
$ kd config use 824f84b3
Switched to http://rjeczalik.koding.team.

Please run "sudo kd restart" for the new configuration to take effect.
```

A todo here could be to add some aliasing support, so it's possible to `kd config use production`, `kd config use rjeczalik` etc.

### Stack workflow

In order to create a stack, user needs to have a valid credential. In order to list credentials, do:

```
$ kd credential list
ID                                TITLE                                              TEAM  PROVIDER  USED
5320460345e6b0ce20c8c64ff8afa777  kdev                                                     aws       -
aa5705ac2f689d6ecd05ccf5dd696e41  khaled                                                   aws       -
f41f95a5553614c9a9df1c490141218d  khaled2                                                  aws       -
8aa8da8016ae0e5a125d41e6ba2dacd1  khaled-test                                              aws       -
06e4df2e93c6ed47b49fecac20099c01  khaled3                                                  aws       -
35c9e3f8d948e3f31bc2f434483943b9  af                                                       aws       -
8e0bb69117abbd6f73f17b79895f4a63  asdasdasd                                                aws       -
58695fb7b8966365413d4cf0          rjeczalik 2017-01-01 19:59:51.521225533 +0000 UTC        aws       -
58696068b8966365413d4cf3          rjeczalik 2017-01-01 20:02:48.006556361 +0000 UTC        aws       -
586960b0b8966365413d4cf6          rjeczalik 2017-01-01 20:04:00.067389374 +0000 UTC        aws       default
6642a8525ef16a3d20d75bc824cad669  dev                                                      marathon  default
b5c4da7a77199924d3e4852d0acccb0d  rafal                                                    vagrant   default
```

It is possible to set a default credential for each of the provider with `kd credential use`. Default credential is the one that is used by stack creation when no explicit credential was provided. When no explicit credential was provided and no default credential is set, `kd stack create` command will fail.

The simplest way to create a stack is to:

```
$ kd stack create
```

This command will use `kd.yml` from the current working directory and create a stack template out of it, create a stack with default credential and build it. The command returns as soon as the stack starts to build. A todo is a `--wait` flag, that will make the command not return until the stack finishes building.

The command also accepts `--file` flag, being either custom path to a template file or `-` - which makes it read the template content from stdin. The file can be either a json, yaml or hcl.

A todo is to use existing stack template with `--template, -t` flag, like:

```
$ kd stack create --template rjeczalik/stack 
```

This will be introduced after `kd template` command is implemented.

### DIFF
--- a/go/src/koding/klientctl/auth.go
+++ b/go/src/koding/klientctl/auth.go
@@ -115,7 +115,7 @@ func AuthLogin(c *cli.Context, log logging.Logger, _ string) (int, error) {
 		enc.SetIndent("", "\t")
 		enc.Encode(resp)
 	} else {
-		fmt.Fprintf(os.Stdout, "Successfully logged in to the following team: %s.\n", resp.GroupName)
+		fmt.Fprintln(os.Stdout, "Successfully logged in to the following team:", resp.GroupName)
 	}
 
 	return 0, nil
--- a/go/src/koding/klientctl/auth.go
+++ b/go/src/koding/klientctl/auth.go
@@ -92,6 +92,8 @@ func AuthLogin(c *cli.Context, log logging.Logger, _ string) (int, error) {
 		}
 	}
 
+	fmt.Fprintln(os.Stderr, "Logging to", kodingURL, "...")
+
 	resp, err := authClient.Login(opts)
 	if err != nil {
 		return 1, fmt.Errorf("error logging into your Koding account: %v", err)
@@ -113,7 +115,7 @@ func AuthLogin(c *cli.Context, log logging.Logger, _ string) (int, error) {
 		enc.SetIndent("", "\t")
 		enc.Encode(resp)
 	} else {
-		fmt.Fprintf(os.Stdout, "Successfully logged in to %q team.\n", resp.GroupName)
+		fmt.Fprintf(os.Stdout, "Successfully logged in to the following team: %s.\n", resp.GroupName)
 	}
 
 	return 0, nil
--- a/go/src/koding/klientctl/endpoint/stack/stack.go
+++ b/go/src/koding/klientctl/endpoint/stack/stack.go
@@ -1,6 +1,7 @@
 package stack
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -130,7 +131,7 @@ func (c *Client) jsonReencode(data []byte) ([]byte, error) {
 	var ymlv interface{}
 
 	if err := yaml.Unmarshal(data, &ymlv); err == nil {
-		return json.Marshal(fixYAML(ymlv))
+		return jsonMarshal(fixYAML(ymlv))
 	}
 
 	var hclv interface{}
@@ -138,7 +139,7 @@ func (c *Client) jsonReencode(data []byte) ([]byte, error) {
 	if err := hcl.Unmarshal(data, &hclv); err == nil {
 		fixHCL(hclv)
 
-		return json.Marshal(hclv)
+		return jsonMarshal(hclv)
 	}
 
 	return nil, errors.New("unknown encoding")
@@ -167,6 +168,19 @@ func (c *Client) readProviders(data []byte) ([]string, error) {
 // Create
 func Create(opts *CreateOptions) (*stack.ImportResponse, error) {
 	return DefaultClient.Create(opts)
+}
+
+func jsonMarshal(v interface{}) ([]byte, error) {
+	var buf bytes.Buffer
+
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
+
+	if err := enc.Encode(v); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
 }
 
 // fixYAML is a best-effort of fixing representation of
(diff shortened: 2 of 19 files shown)


cihangir:
> For the first time kd will ask for username / password in order to obtain kite.key from kontrol. Once kite.key is saved, consecutive kd auth login --team x commands will proceed without asking for password.

What about the cases where a team member is kicked out?

cihangir:
<3

```kd config use 824f84b3
kd credential use.```

**Review by cihangir — APPROVED** [review]
LGTM, most prob "login" issue is another thing we should take a look, not related with this PR

**Comment by rjeczalik** [comment]
> What about the cases where a team member is kicked out?

There is no change here - works as previously.

To summarize:

Kite.key and clientId are used to authenticate - kite.key for kites, clientId for socialapi/socialworker. Password-login is used to obtain kite.key, having kite.key we can obtain for clientId (`kloud.auth` kite method).

With kite.key we authenticate to

- kloud

Kloud does authorize. For every stack related methods kloud authorizes user - if user does not belong to a group (kicked out) or have insufficient jPermissions (no admin etc.) then the request is denied which makes `kd stack create` to fail.

- kontrol

Afair there's no authorization in kontrol, since kite.key has no team information kicking out user from a team has no effect - user can still register to kontrol and GetKites etc. No handling here.

- klient

Also no team authorization - klient allows requests from external user (e.g. webterm.connect for a shared sessions) if the external user is saved locally in klient.bolt file (whitelist). 

The part I'm sure about is socialworker and socialapi - in particular whether they do authorize. The access to these two services is not changed by this PR.

The missing team authorization in klient is going to be taken care of by this epic:

https://github.com/koding/koding/issues/10154

So not only we add authorization for team-wide access, we would also authorize single users whether they are team members, however I'm not sure the latter is possible without having `team` info in JWT claims (kite.key).

**Comment by rjeczalik** [comment]
> however I'm not sure the latter is possible without having team info in JWT claims

On a second though for this particular use-case it is possible - if external user is locally authorized in klient.bolt we can ask koding whether it belongs to the same team as we. My talking about adding `team` to kite.key is about moving team validation from authorization phase to authentication one.

**Comment by cihangir** [comment]
just side note, we delete 'clientId' s after 14 days

File: go/src/koding/klientctl/auth.go [hunk 1]
```diff
@@ -92,6 +92,8 @@ func AuthLogin(c *cli.Context, log logging.Logger, _ string) (int, error) {
 		}
 	}
 
+	fmt.Fprintln(os.Stderr, "Logging to", kodingURL, "...")
```
ppknap: It's better to have `os.Stdout` here.

go/src/koding/klientctl/endpoint/stack/stack.go
```diff
@@ -35,22 +36,21 @@ func (opts *CreateOptions) Valid() error {
...
 
-// Create
 func (c *Client) Create(opts *CreateOptions) (*stack.ImportResponse, error) {
 	if err := opts.Valid(); err != nil {
 		return nil, err
 	}
 
+	log.Println("BEFORE RE-ENCODE")
 	data, err := c.jsonReencode(opts.Template)
+	log.Println("AFTER RE-ENCODE")
```
ppknap: I might be wrong but this looks like debug prints. Are you going to leave them here?

ppknap (approved):
Just minor things

go/src/koding/klientctl/endpoint/stack/stack.go
rjeczalik: Done 🤦‍♂️

go/src/koding/klientctl/auth.go
rjeczalik: I don't share this opinion, I took an approach that a final message that is the result of the command get printed to os.Stdout, anything that is printed before are just logs and printed to os.Stderr.

Without this assumption it won't be possible to write a test.

PR #10238, opened 2017-01-08
vendor: update hashicorp/hcl

rjeczalik:
A note about the mcov package:

In order to write tests for kd command line we used [TestHelperProcess trick](https://github.com/golang/go/blob/master/src/os/exec/exec_test.go#L624) ([TestMainHelper in kd](https://github.com/koding/koding/blob/f17125854a56aff971a5965bf87025c9894e6bd5/go/src/koding/klientctl/main_test.go#L18)) - the TestMainHelper wraps main function, which [is then called](https://github.com/koding/koding/blob/f17125854a56aff971a5965bf87025c9894e6bd5/go/src/koding/klientctl/main_test.go#L60) from within a test itself. E.g. a stack test runs `kd credential create` + `kd credential use` + `kd stack create` and inspects the output of each asserting certain content.

This is needed because of two reasons:

- main.go contains a certain global state + it sets up the command line interface - we wanted to have end-to-end tests that treat main.go as a black box, because
- we have an idea to replace the cli package with spf13/cobra and while doing that, rework all the commands to be more composable (so calling ~main would be no longer needed for an end-to-end test)

Before we reach there, we're going to collect test coverage reports form existing kd. Since most of the actual code is executed in a subprocess, it's not recorded in the original coverprofile file.

The mcov package allows for gathering coverage profiles from subprocesses and merge them all back to a single file per package.


cihangir:
mcov is not needed i beileve, https://github.com/koding/koding/pull/10155 already handles that

cihangir:
+1 for spf13/cobra

rjeczalik:
> mcov is not needed i beileve, #10155 already handles that

I think it does not. I've researched how codecov uploads the coverage reports - it just concatenates the files:

https://gist.github.com/rjeczalik/78eb285f0140e067cbd090811780d8f1#file-codecov-bash-L1063

Concatenating multiple profiles, which are produced as described above, leads to wrong results.

cihangir:
Yes it merges them into one file but it seems it works correctly, while @mehmetalisavas was working on that task i told him to add "merging" cover profiles support but it turned out it is not needed because their backend seems like handles it properly. Please let me know if you find any anomalies in the result set of test runs in the `codecov.io` reports, because i couldnt find any in the places i've looked.

just fyi: coverpkg flag in the respective PR is not added intentionally for now.

rjeczalik:
Yes, it works correctly when you have 1 profile per package. As described above, kd is different - it has multiple profiles per 1 package - and those needs to be merged, otherwise for majority of command functions (`StackCreate`, `CredentialCreate` etc.) the coverage will be displayed as `0%`.

Let's revisit this issue once we have codecov.io integrated, for now I removed mcov.

cihangir (approved):

ppknap (approved):

PR #10240, opened 2017-01-08
IDE: Syntax-Selector, Text in the select-box is not aligned correctly (vertically)

### LINKED ISSUE #9979: Select-box (next to 'Syntax' option on IDE file menu) alignment is a pixel off.

hakankaradis:
Syntax-selector, vertical alignment fix

Fixes: #9979

### DIFF
--- a/client/ide/lib/styl/ide.styl
+++ b/client/ide/lib/styl/ide.styl
@@ -718,6 +718,9 @@ body.ide .dark .ide-files-tab .kdtabhandle.close-handle .icon
         height              20px !important
         padding             0
 
+      span.title
+        padding-top         2px
+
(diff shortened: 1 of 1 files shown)


usirin:
I am not sure if this is the proper fix for given issue. It points out to the right alignment of the selectbox with other items, and this PR only has a change on `title`. Am I missing something?

hakankaradis:
@usirin The issue is about the vertical alignment of `Text` in the selectbox, we already fixed the issue that you mentioned in #10000.

sinan (approved):

PR #10247, opened 2017-01-09
Remove unnecessary cookies - part 2

### LINKED ISSUE #10165: Remove _hp2_id.* cookies (set by heap)

### LINKED ISSUE #10167: Remove __hstc cookie ( from hubspot )

### LINKED ISSUE #10169: Remove cookies ending with mixpanel (mp_91ac80db7e24a52e75851f9d431198b5_mixpanel)

### LINKED ISSUE #10170: Remove cookies starting with _ob_pub (_ob_pub-95fb772a45399d22b903c934ac1e894a) - unknown origin

cihangir:
fixes https://github.com/koding/koding/issues/10165
fixes https://github.com/koding/koding/issues/10167
fixes https://github.com/koding/koding/issues/10169
fixes https://github.com/koding/koding/issues/10170


client/app/lib/util/cleanup.coffee
```diff
@@ -3,16 +3,63 @@ kookies = require 'kookies'
...
+  'hubspot.hub.id'
+  'hsfirstvisit'
+  'hubspotauth'
+  'hubspotauthcms'
+  'hubspotauthremember'
+  'hubspotutktzo'
+  '_hs_opt_out'
+  '__hluid'
+
+  # kissmetrics
```
cihangir: @sinan @devrim do we use kissmetrics & olark & mixpanel? there are many integrations on segment and i am not sure which ones are in use

sinan: we do use kissmetrics but not olark and mixpanel.

sinan: from segment I removed mixpanel and heap integrations.

cihangir:
@sinan removed kissmetrics cookies from the list, PTAL

client/app/lib/util/cleanup.coffee
```diff
@@ -3,16 +3,55 @@ kookies = require 'kookies'
...
+  'hubspotauthcms'
+  'hubspotauthremember'
+  'hubspotutktzo'
+  '_hs_opt_out'
+  '__hluid'
+
+  # font?
+  'ajs_anonymous_id'
+  'ajs_group_id'
+  'ajs_user_id'
```
sinan: these are probably segment's, i think we should keep them as well. see: https://segment.com/docs/sources/website/analytics.js/#anonymous-id

client/app/lib/util/cleanup.coffee
```diff
@@ -3,16 +3,55 @@ kookies = require 'kookies'
...
 toBeDeleted = [
   'clientIPAddress' # an old cookie set by koding backend.
+  'kdproxy-preferred-domain' # old koding cookie
   'connect.sid'     # express cookie - infact this is session cookie, here for just in case.
-  'pnctest'         # set by PubNub for testing.
-  'hubspotutk'      # hubspot
-  'hsfirstvisit'    # hubspot
+  # 'pnctest'       # set by PubNub for testing. it is set on every load, so not necessary to remove.
   '__ssid'          # google map
+  'gsScrollPos'     # session cookie, wont be cleaned, not necessary
```
sinan: that is probably a chat widget's scroll position cookie, safe to remove but the comment on the right is wrong.

client/app/lib/util/cleanup.coffee
```diff
@@ -38,3 +77,9 @@ module.exports = ->
     for option in options
       if kookies.expire key, option
         console.log "#{key} deleted"
+
+  for exp in regexWith
```
sinan: you can do this with only one regex:

```regex
^(?:_hp2_(?:id|ses_props)|_ob_pub)|.*_mixpanel$
```

regex

if you think this is more readable, that's fine. But either way you need to change `_mixpanel$` to `.*_mixpanel$`

sinan (changes requested):
👍 thx @cihangir

client/app/lib/util/cleanup.coffee
cihangir: which part is wrong?

client/app/lib/util/cleanup.coffee
cihangir: nice! i wasnt sure about the source of these, came across to "set by custom font provider" in many places that is why i wrote `font?` removing immediately.

client/app/lib/util/cleanup.coffee
cihangir: What is wrong with `_mixpanel$` ? seems like works [here](http://coffeescript.org/#try:alert%20%2F%2F%2F_mixpanel%24%2F%2F%2F.test%20%22mp_113c2731b47a5151f4be44ddd5af0e7a_mixpanel%22)

sinan: thought it should work too but when I tested, the engine I used returned false, so I made it fail safe. but as you say it works, we can keep as is.

client/app/lib/util/cleanup.coffee
sinan: the comment says 'session cookie' which is too generic, thought it might be better saying:

'gsScrollPos'     # 3rd party scroll position cookie, safe to remove`

cihangir: aggreed it is too generic :) already created confusion but i am not the name father, "session cookie" is its attribute http://snpy.in/fZhdSY browsers cant alter that value, hence wont be cleaned

cihangir: removed it completely, since it was there just to show it is not required by our system.

sinan (approved):

PR #10256, opened 2017-01-09
scripts: support -coverpkg flag for tests.sh script

mehmetalisavas:


szkl (approved):
Looks good to me.

I've noticed gotest.sh passes all arguments to functions. We can improve that by wrapping `$@` expression in double quotes. That'd make it avoid unwanted expansion and substitution. Although, functions only use first positional argument by referencing `$1` directly. Other arguments would be discarded. Those can benefit from `"$@"` for future compatibility.

PR #10266, opened 2017-01-10
stacktemplate: setAccess, Fix callback when there is no group.slug data

hakankaradis:
Minor update in setAccess function of stack template
Return correct callback when there is not group.slug data provided

### DIFF
--- a/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
+++ b/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
@@ -373,9 +373,9 @@ module.exports = class JStackTemplate extends Module
       @update query, (err) =>
 
         return callback err  if err
-        return  unless group.slug
 
-        return callback null, this  if currentAccessLevel is accessLevel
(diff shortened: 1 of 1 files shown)


workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -373,9 +373,9 @@ module.exports = class JStackTemplate extends Module
       @update query, (err) =>
 
         return callback err  if err
-        return  unless group.slug
 
-        return callback null, this  if currentAccessLevel is accessLevel
+        if currentAccessLevel is accessLevel or not group.slug
+          return callback null, this
```
hakankaradis: @gokmen I just thought that shouldn't we throw error when there is no group.slug data provided?
update shouldn't work if we are not able to get group.slug data 😕

gokmen: it doesn't need to be, we can leave this as is.

gokmen (approved):

PR #10268, opened 2017-01-10
Remove unnecessary operations from page load flow

cihangir:
Main reason for this PR to remove socialapi cache from the page load but i didnt skimp on it :)


go/src/koding/kites/kloud/team/mongo.go
```diff
@@ -87,11 +85,9 @@ func groups2teams(groups ...*models.Group) []*Team {
 			continue
 		}
 
-		members, _ := group.Counts["members"].(int)
```
cihangir: cc/ @ppknap

workers/social/lib/social/models/group/index.coffee
```diff
@@ -495,18 +475,8 @@ module.exports = class JGroup extends Module
...
-    #   return callback err  if err
-    #   return callback new KodingError "Subscription is not found"  unless subscription
-    #   subscription.debitPack tag: "group", (err) =>
-    #     return callback err  if err
     @create client, formData, delegate, (err, group) ->
-      return callback err if err
-      # group.addSubscription subscription, (err) ->
-      #   return callback err  if err
-      callback null, { group }
+      return callback err, group
```
gokmen: just this one questioned me, checked it out site-wide it seems it's never used like `group.group` not sure how it was working :)

gokmen (approved):
💯

go/src/koding/kites/kloud/team/mongo.go
ppknap: Could you please also rm two things:

1) `Members` field from `Team` [here](https://github.com/koding/koding/blob/e19251e0ef2b53c1a485ac62aa8a9f6bfd09e67f/go/src/koding/kites/kloud/team/client.go#L9) 

2) `MEMBERS` column [here](https://github.com/koding/koding/blob/e19251e0ef2b53c1a485ac62aa8a9f6bfd09e67f/go/src/koding/kites/kloud/team/client.go#L9)

Seems like I forgot to do this before because we were talking about this..

PR #10273, opened 2017-01-10
Add bongo caching

cihangir:
Shared caching support with magic functions in bongo.


workers/social/lib/social/models/group/index.coffee
```diff
@@ -260,6 +260,7 @@ module.exports = class JGroup extends Module
         type        : String
         validate    : require('../name').validateName
         set         : (value) -> value.toLowerCase()
+        cacheable   : yes
```
cihangir: when you add a new property with "cacheable" it will automatically add `fetchFromCacheByXxx` function to the model (it will be only static function since fetching from instance is meaningless)

```
JGroup.fetchFromCacheBySlug 'koding', (err, data) ->
    console.log err, "data", data.getId()
```

gokmen (approved):

PR #10275, opened 2017-01-11
StackTemplate: slug support, create with defaults

### LINKED ISSUE #10269: mongo: add jStackTemplate.slug
In order to be able address stack templates basing on a unique slug - add it to the jStackTemplate documents.

```
$ kd template show -t foobar/mysql
```

Naming

The slugs needs to be unique, however it should be possible for a different users to use the same names for their template. That is UserA and UserB should be able to create a mysql template, without needing to pick an available name. Thus, the jStackTemplate should be in team/stack format, e.g. UserA creates a mysql template for foobar team, the jStackTemplate.slug would be "foobar/mysql". The names should be:

- case-insensitive ("FooBar/MySQL" == "foobar/mysql")
- url safe

For now we could create slugs out of titles with some heuristics, example:

- `Default stack template` -> `template`
- `Default stack template2` -> `template-2`
- `MongoDB Stack` -> `mongo`
- `Golduck Google Stack` -> `golduck-google`
- `Plantilla estandard de ambiente de desarrollo` -> `plantilla-estandard-de-ambiente-de-desarrollo`

Migrations

For all existing stack templates we should generate slugs. Above example mapping could be used

TODO

- [ ] socialworker: set slugs during JStackTemplate.create / update swagger.json
- [ ] migrations: add workers/migrations/0025-jstacktemplate-slug.js

/cc @gokmen @cihangir

gokmen:
- [x] `slug` field needs to be created with unique index on `group`, `originId` and `slug`
- [x] `slug` should be created automatically from `title` if not provided
- [x] `title` should be auto generated like in client side if not provided
- [x] all required fields of `config` should be auto generated if not provided
- [x] `slug` should automatically provide incremental alternatives (`foo, foo-1, foo-2` and so on)

fixes: #10269 
rel: https://github.com/koding/koding/pull/9488/files#diff-888f0565303232623f005f3516ced158

### DIFF
--- a/workers/migrations/0015-create_default_indexes.js
+++ b/workers/migrations/0015-create_default_indexes.js
@@ -78,15 +78,6 @@ var indexes = {
       "_id": 1
     },
     "ns": "koding.jStackTemplates"
-  }, {
-    "name": "groupName_originId_slug",
-    "key": {
-      "slug": 1,
-      "group": 1,
-      "originId": 1
-    },
-    "ns": "koding.jStackTemplates",
-    "unique": true
   }],
   "jSecretNames": [{
     "name": "_id_",
--- a/workers/migrations/0015-create_default_indexes.js
+++ b/workers/migrations/0015-create_default_indexes.js
@@ -78,6 +78,15 @@ var indexes = {
       "_id": 1
     },
     "ns": "koding.jStackTemplates"
+  }, {
+    "name": "groupName_originId_slug",
+    "key": {
+      "slug": 1,
+      "group": 1,
+      "originId": 1
+    },
+    "ns": "koding.jStackTemplates",
+    "unique": true
   }],
   "jSecretNames": [{
     "name": "_id_",
--- a/workers/migrations/0025-stacktemplate-index.js
+++ b/workers/migrations/0025-stacktemplate-index.js
@@ -1,0 +1,20 @@
+
+var mongodb = require('mongodb');
+
+exports.up = function(db, next){
+  var index = {
+    "name": "groupName_originId_slug",
+    "key": {
+      "slug": 1,
+      "group": 1,
+      "originId": 1
+    },
+    "ns": "koding.jStackTemplates",
+    "unique": true
+  }
+  db.ensureIndex("jStackTemplates", index.key, index, next);
+};
+
+exports.down = function(db, next){
+  next();
+};
--- a/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
+++ b/workers/social/lib/social/models/computeproviders/stacktemplate.coffee
@@ -180,17 +180,30 @@ module.exports = class JStackTemplate extends Module
     return config
 
 
-  validateTemplate = (templateData, group, callback) ->
+  validateTemplateData = (data, group, callback) ->
 
-    unless templateData
+    unless data
       return callback new KodingError 'Template data is required!'
 
-    { template } = templateData
+    { config = {}, slug, title, template, rawContent, templateDetails } = data
+
+    if not template or not rawContent
+      return callback new KodingError 'Template content is required!'
+
+    template = generateTemplateObject template, rawContent, templateDetails
+    config   = updateConfigForTemplate config, template.rawContent
+    title   ?= generateTemplateTitle config.requiredProviders[0]
+    slug     = slugify slug ? title
+
+    replacements = { template, config, title, slug }
+
     limitConfig  = helpers.getLimitConfig group
-    return callback null  unless limitConfig.limit # No limit, no pain.
+    unless limitConfig.limit # No limit, no pain.
+      return callback null, replacements
 
     ComputeProvider = require './computeprovider'
-    ComputeProvider.validateTemplateContent template, limitConfig, callback
+    ComputeProvider.validateTemplateContent data.template, limitConfig, (err) ->
+      callback err, replacements
 
 
   generateTemplateTitle = (provider) ->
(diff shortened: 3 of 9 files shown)


rjeczalik:
While trying to save the template I'm getting:

```
pq: duplicate key value violates unique constraint "account_nick_key"
```

Just heads up, I realize it's still wip - wanted to be an alpha user 😎

cihangir:
@rjeczalik prob your postgres docker is kinda dead or migrations did not work properly, recreation might solve it if there isnt a problem with this branch itself - didnt check the code, just saw your comment and wanted to reply.

gokmen:
This ready to review, I've decided to write migrate script for updating existing stacktemplates to have slugs. cc/ @rjeczalik @cihangir

workers/migrations/0015-create_default_indexes.js
```diff
@@ -78,6 +78,15 @@ var indexes = {
       "_id": 1
     },
     "ns": "koding.jStackTemplates"
+  }, {
```
cihangir: updating existing migrations is not effective, please create another migration

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -160,19 +173,72 @@ module.exports = class JStackTemplate extends Module
...
     config.requiredData      = requirementsParser template
     config.requiredProviders = providersParser template, supportedProviders
 
     return config
 
 
-  validateTemplate = (template, group, callback) ->
+  generateSlugFromTitle = ({ originId, group, title, index }, callback) ->
+
+    unless title
```
cihangir: lets put a termination statement here ie `return callback errTooBig if index > 20`

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -160,19 +173,72 @@ module.exports = class JStackTemplate extends Module
...
+      replacements = { template, config, title, slug }
+
+      if not data.template     or \
+         not limitConfig.limit or \
+             limitConfig.limit is 'unlimited' # No limit, no pain.
+        return callback null, replacements
+
+      ComputeProvider = require './computeprovider'
+      ComputeProvider.validateTemplateContent \
+        data.template, limitConfig, (err) -> callback err, replacements
```
cihangir: returning replacements is not required ( from the places i've checked ) on error cases

workers/social/lib/social/models/computeproviders/stacktemplate.test.coffee
```diff
@@ -52,6 +54,39 @@ runTests = -> describe 'workers.social.models.computeproviders.stacktemplate', -
             expect(template.template.sum)     .to.exist
             done()
 
+      it 'should generate a title/slug if not provided', (done) ->
```
cihangir: 👍

cihangir:
https://github.com/koding/koding/pull/10275#issuecomment-272376140

@gokmen prob on side of the world wont be online while we are deploying to latest/prod, lets have the lazy migrations of slugs in the code, otherwise there might be gaps in the data before/during/after the deploy/switch operation. Lets try not to synchronize as much as possible.

cihangir (changes requested):
migration file needs to be separated

rjeczalik (approved):

workers/migrations/0015-create_default_indexes.js
gokmen: I know how migration works :) I thought since we didn't use it on prod db yet we don't need to pollute the migration stack and therefore it will work for dev environments since we're resetting db each time when we do `./run buildservices`, so it will work in both cases.

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
gokmen: right, but lets keep it in here in case of feature requirements. we may need to log the `replacements` once an error occurred.

gokmen:
https://github.com/koding/koding/pull/10275#issuecomment-272384018

@cihangir I thought the same thing but it won't be an issue since we don't have any piece of code yet relying on `slug` field. Instead of doing an update operation on each fetch request we can do it at once even after we deploy and switch this one. New documents will have slugs, old one doesn't but as said current codebase not relying that field at all.

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
cihangir: lets add it if we need it, also it is against our flow to return both error and the possible partial response - in most cases

workers/migrations/0015-create_default_indexes.js
cihangir: not running it on prod does not allow us to update migration files, also we wont be running these first migrations in prod, since they are not meant for prod. 

migration files should be written for keeping development env runnning flawlessly, so i shouldnt be need to run buildservices to have migrations applied. so it wont work.

cihangir:
https://github.com/koding/koding/pull/10275#issuecomment-272529507
if you say so, that is fine with me

gokmen:
@cihangir all done, ptal.

usirin (approved):

gokmen:
@cihangir I had to merge this one, please do not hesitate to add your review if needed.

workers/migrations/0025-stacktemplate-index.js
```diff
@@ -0,0 +1,20 @@
+
+var mongodb = require('mongodb');
+
+exports.up = function(db, next){
+  var index = {
+    "name": "groupName_originId_slug",
+    "key": {
+      "slug": 1,
```
cihangir: with the current code base it seems having slug at the end could be better use of the index - since there are other queries that uses originId and group. as you know property order matters in mongo indexes

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -188,6 +188,10 @@ module.exports = class JStackTemplate extends Module
     slug = _title = if index? then "#{title} #{index}" else title
     slug = slugify slug
 
+    if index >= 20
+      return callback new KodingError \
```
cihangir: 👍

Issue #10205, opened 2017-01-04
SEND INVITES does not look like a CTA


diclee:
## Current Behavior
People are having hard time locating it

![image](https://cloud.githubusercontent.com/assets/8261892/21627634/0a86c2b0-d1ce-11e6-8f40-a77c013e6fba.png)

ftasdelen:
We can add a button instead of the link. We can make it faded if user doesn't enter any recipient. When they enter any, we can make the button 100% visible.
![image](https://cloud.githubusercontent.com/assets/306264/21657453/2f098d3a-d277-11e6-97f0-c7413bc8bcda.png)

cc @diclee

diclee:
Verify passed on sandbox & latest.

PR #10288, opened 2017-01-12
Send Invites button and become active dynamically

andrewjcasal:
## Description

References https://github.com/koding/koding/issues/10205.
- Send invite button starts as inactive.
- With one valid e-mail, send invite button becomes active.
- With more valid e-mails, send invite button count reflects that number.

No inputs filled in:
image

Some e-mail addresses:
image

In action:
http://screencast.com/t/2j1Q8Xg3jZcg

## Types of changes

## Checklist:

### DIFF
--- a/client/home/lib/myteam/components/hometeamsendinvites/view.coffee
+++ b/client/home/lib/myteam/components/hometeamsendinvites/view.coffee
@@ -2,7 +2,7 @@ kd = require 'kd'
 React = require 'app/react'
 List = require 'app/components/list'
 CheckBox = require 'app/components/common/checkbox'
-
+isEmailValid = require 'app/util/isEmailValid'
 
 module.exports = class HomeTeamSendInvitesView extends React.Component
 
@@ -35,11 +35,14 @@ module.exports = class HomeTeamSendInvitesView extends React.Component
         onClick={@props.onInputChange.bind(null, rowIndex, 'canEdit', { target: { value: not checked}})}/>
     </div>
 
+
   renderEmptySectionAtIndex: -> <div> No data found</div>
 
 
   render: ->
-    count = _.sum @props.filledInputsObj
+    count = _.sum this.props.inputValues.toList().toArray().map (inputRow) ->
+       if isEmailValid inputRow.get 'email' then 1 else 0
+
     if count > 0
       buttonTitle = "Send #{count} Invite#{if count > 1 then 's' else ''}"
       buttonEnabled = yes
--- a/client/home/lib/myteam/components/hometeamsendinvites/view.coffee
+++ b/client/home/lib/myteam/components/hometeamsendinvites/view.coffee
@@ -43,10 +43,7 @@ module.exports = class HomeTeamSendInvitesView extends React.Component
     </div>
 
   onInputChange: (rowIndex, fieldName, evt) ->
-    if isEmailValid evt.target.value
-      @state.filledInputCount[rowIndex] = 1
-    else
-      @state.filledInputCount[rowIndex] = 0
+    @state.filledInputCount[rowIndex] = if isEmailValid evt.target.value then 1 else 0
 
     @props.onInputChange(rowIndex, 'email', evt)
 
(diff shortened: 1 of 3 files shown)


client/home/lib/myteam/components/hometeamsendinvites/view.coffee
```diff
@@ -25,7 +32,7 @@ module.exports = class HomeTeamSendInvitesView extends React.Component
     userEmailClassName = 'kdinput text user-email'
 
     <div className='kdview invite-inputs'>
-      <input type='text' className={userEmailClassName} placeholder='mail@example.com' value={inviteInput.get 'email'} onChange={@props.onInputChange.bind(this, rowIndex, 'email')} />
+      <input type='text' className={userEmailClassName} placeholder='mail@example.com' value={inviteInput.get 'email'} onChange={@onInputChange.bind(this, rowIndex, 'email')} />
```
sinan: @hakankaradis @usirin can you approve this line?

sinan (approved):
👍

client/home/lib/myteam/components/hometeamsendinvites/view.coffee
```diff
@@ -2,10 +2,17 @@ kd = require 'kd'
...
+isEmailValid = require 'app/util/isEmailValid'
 
 
 module.exports = class HomeTeamSendInvitesView extends React.Component
 
+  constructor: (props) ->
+
+    super props
+
+    @state = { filledInputCount : {} }
```
sinan: pls rename the key, it's not holding a decimal anymore so you can't name it as `...Count`

sinan (changes requested):

client/home/lib/myteam/components/hometeamsendinvites/view.coffee
```diff
@@ -35,11 +42,22 @@ module.exports = class HomeTeamSendInvitesView extends React.Component
         onClick={@props.onInputChange.bind(null, rowIndex, 'canEdit', { target: { value: not checked}})}/>
     </div>
 
+  onInputChange: (rowIndex, fieldName, evt) ->
+    @state.filledInputCount[rowIndex] = if isEmailValid evt.target.value then 1 else 0
```
usirin: assigning to state should be done via `setState` method, but it will trigger a new render, so to make it safer we should use the completion handler of `setState` method (a callback passed as 2nd argument):

```coffee
filledInputCount = _.assign {}, @state.filledInputCount
filledInputCount[rowIndex] = if isEmailValid evt.target.value then 1 else 0

@setState { filledInputCount }, => 
  @props.onInputChange rowIndex, 'email', evt
```

But this will trigger 2 render cycles, so i am not sure if this is the best way to approach this problem. I think moving `filledInputCount` to props, and managing it from the parent component would be a better approach.

client/home/lib/myteam/components/hometeamsendinvites/view.coffee
usirin: There is nothing wrong with this line, but the handler might need changes. I added my comments over there.

hakankaradis (commented):
@sinan @umut Whenever you enter anything to that section `@props.inputValues` are filled so we just can check `@props.inputValues` to be able to get emails count, no need to hold another state?
or we can add another getter function that returns only emails so button visibility and count can be checked from this getter?

andrewjcasal:
Ready for review again @hakankaradis, @sinan, @usirin

hakankaradis (approved):
LGTM

client/home/lib/myteam/components/hometeamsendinvites/view.coffee
```diff
@@ -40,6 +40,15 @@ module.exports = class HomeTeamSendInvitesView extends React.Component
 
 
   render: ->
+    count = _.sum this.props.inputValues.toList().toArray().map (inputRow) ->
+       if isEmailValid inputRow.get 'email' then 1 else 0
```
sinan: `~~isEmailValid inputRow.get 'email'` yours is more readable tho :)

sinan (approved):

PR #10297, opened 2017-01-12
kd/template: add list / show / delete subcommands

rjeczalik:
~Depends on: https://github.com/koding/koding/pull/10296~

This PR adds the following commands:

- kd template list

```
$ kd template list
ID                        TITLE              SLUG               OWNER      TEAM    ACCESS   MACHINES
587777ba112c2d2217cab6aa  Diglett Aws Stack  diglett-aws-stack  rjeczalik  foobar  private  0
```

- kd template show

```
$ kd template show -t rjeczalik/diglett-aws-stack
provider:
  aws:
    access_key: ${var.aws_access_key}
    secret_key: ${var.aws_secret_key}
resource:
  aws_instance:
    aws-instance:
      ami: ""
      instance_type: t2.nano
      tags:
        Name: ${var.koding_user_username}-${var.koding_group_slug}
      user_data: |2

        echo "hello world!" >> /helloworld.txt
```
```
$ kd template show -t rjeczalik/diglett-aws-stack -json
{
	"provider": {
		"aws": {
			"access_key": "${var.aws_access_key}",
			"secret_key": "${var.aws_secret_key}"
		}
	},
	"resource": {
		"aws_instance": {
			"aws-instance": {
				"ami": "",
				"instance_type": "t2.nano",
				"tags": {
					"Name": "${var.koding_user_username}-${var.koding_group_slug}"
				},
				"user_data": "\necho \"hello world!\" >> /helloworld.txt\n"
			}
		}
	}
}
```
```
$ kd template show -t rjeczalik/diglett-aws-stack -hcl
"provider" "aws" {
  "access_key" = "${var.aws_access_key}"

  "secret_key" = "${var.aws_secret_key}"
}

"resource" "aws_instance" "aws-instance" {
  "tags" = {
    "Name" = "${var.koding_user_username}-${var.koding_group_slug}"
  }

  "instance_type" = "t2.nano"

  "ami" = ""

  "user_data" = "\necho \"hello world!\" >> /helloworld.txt\n"
}
```

- kd template delete

```
$ kd template delete -t rjeczalik/diglett-aws-stack
Please type "yes" to confirm you want to delete the resource []: yes
Stack template with "587777ba112c2d2217cab6aa" ID deleted successfully.
```
```
$ kd template show -t rjeczalik/diglett-aws-stack -json
error executing "show" command: resource was not found
```

### DIFF
--- a/go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go
+++ b/go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go
@@ -7,6 +7,7 @@ import (
 	"koding/klientctl/config"
 	"koding/klientctl/endpoint"
 	"koding/klientctl/endpoint/kloud"
+	"koding/klientctl/endpoint/team"
 	"koding/remoteapi"
 	"koding/remoteapi/client"
 )
@@ -15,15 +16,17 @@ var DefaultClient = &Client{}
 
 type Client struct {
 	Kloud *kloud.Client
+	Team  *team.Client
 
-	once sync.Once // for c.init()
-	c    *remoteapi.Client
+	once   sync.Once // for c.init()
+	api    *remoteapi.Client
+	client *client.Koding
 }
 
 func (c *Client) New(user *api.User) *client.Koding {
 	c.init()
 
-	return c.c.New(user)
+	return c.api.New(user)
 }
 
 func (c *Client) init() {
@@ -31,10 +34,14 @@ func (c *Client) init() {
 }
 
 func (c *Client) initClient() {
-	c.c = &remoteapi.Client{
+	c.api = &remoteapi.Client{
 		Transport: endpoint.Transport(c.kloud()),
 		Endpoint:  config.Konfig.Endpoints.Remote().Public.URL,
 	}
+	c.client = c.api.New(&api.User{
+		Username: c.kloud().Username(),
+		Team:     c.team().Used().Name,
+	})
 }
 
 func (c *Client) kloud() *kloud.Client {
@@ -44,4 +51,9 @@ func (c *Client) kloud() *kloud.Client {
 	return kloud.DefaultClient
 }
 
-func New(user *api.User) *client.Koding { return DefaultClient.New(user) }
+func (c *Client) team() *team.Client {
+	if c.Team != nil {
+		return c.Team
+	}
+	return team.DefaultClient
+}
--- a/go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go
+++ b/go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go
@@ -1,10 +1,13 @@
 package remoteapi
 
 import (
+	"errors"
 	"sync"
+	"time"
 
 	"koding/api"
 	"koding/klientctl/config"
+	"koding/klientctl/ctlcli"
 	"koding/klientctl/endpoint"
 	"koding/klientctl/endpoint/kloud"
 	"koding/klientctl/endpoint/team"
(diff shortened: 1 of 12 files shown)


cihangir:
what does `/` do in front of the template names?

rjeczalik:
> what does / do in front of the template names?

@cihangir Unhandled empty username. The effect is the following command:

>kd template list -t /diglett-aws-stack -json

Queries for JStackTemplates which slug is `diglett-aws-stack`. I believe in such case, without setting OriginID, remote.api will return all stacks named `diglett-aws-stack` from all users, the one making query has access to.

I was going to forbid empty usernames, but while I was using it, it feels to me it's rather feature than a bug. I'm open to opinions.

rjeczalik:
@gokmen While working at it I think now it would be easier to use JStackTemplate.username than JStackTemplate.originId. Basically in every client - being kd, kloud etc. - I need to setup a JAccount.profile.nickname lookup backed with cache. I think every external user of JStackTemplate.one will end up reimplementing this.

gokmen:
https://github.com/koding/koding/pull/10297#issuecomment-272435476

@rjeczalik I tried to tell this many times in our conversation :) it's your choice :p

rjeczalik:
mistakes were made

gokmen:
@rjeczalik I can update current implementation for slugs to use `username` field (which doesn't exists) instead of `originId` for new documents. and in migration script I can add missing `username` fields while setting `slug`s. and then we can remove all `originId` fields once we make the switch. cc/ @cihangir

rjeczalik:
For now I'm going to use JStackTemplate.originId with lookups and for future we can switch to JStackTemplate.username - so I think it is not urgent.

cihangir:
One thing i am missing is what happened to the idea of a unified API endpoint and that endpoint handling all that jazz?

rjeczalik:
@cihangir Which one should serve as the one API endpoint? Currently we there is two endpoints that kd interacts with (there's also kontrol, but let's neglect it):

- kloud - for terraform-related operations, like build credential, build stack etc.
- remote.api - for mongo-related operations, like create template doc, find template doc, delete template doc, find group docs, list stacks

If we'd want kd to use only one endpoint, we'd need to reimplement part of kloud in remote.api or part of remote.api in kloud. Wdyt?

cihangir:
@rjeczalik i was thinking to have the listing/filtering (by the given string ie <team>/<user>/<name>) logic handled by remoteapi.

All ops could start with search/get operation from the remote.api 
ie:
 (taken from the description)

  ```kd template delete -t rjeczalik/diglett-aws-stack```

could fetch the info of the stack first then deletes it  with a response of

```Stack template with "587777ba112c2d2217cab6aa" ID deleted successfully.```

all RUD ops could follow the same pattern.

Does that sound plausible?

**Comment by rjeczalik** [comment]
@cihangir All the operations here are handled by remote.api. Also all `kd machine list`, `kd credential list` and alike operations are going to be migrated to remote.api in future (see #10333).

**Comment by rjeczalik** [comment]
@cihangir I probably did not address this part:

> could fetch the info of the stack first then deletes it with a response of

Yes, all `kd template *` lookup results are cached, e.g. once you lookup `rjeczalik/diglett-aws-stack` (either with list, create ops), the originID of `rjeczalik` is going to be cached in bolt. And every another operations does not lookup the originID anymore if it exists in cache.

Cache eviction is missing (e.g. when the lookup using cached value gives 404), but at this point I do not expect the originID to change (these are fixed and unique values).

**Comment by cihangir** [comment]
> I do not expect the originID to change

the context and the use case is not clear from the sentence but if a user deletes the account and re-register with the same username ( which was a common op once a while back ), it can change

**Review by cihangir — APPROVED** [review]

**Comment by rjeczalik** [comment]
@cihangir The context was missing cache invalidation mechanism, the use-case was the exact as you described - if user deleted an account, created the same account again, and then tried to `kd template list user/stack`, the command would fail with 404 since it used old cached originId. I just mentioned this shortcomming in my implementation in my comment above.

For the short term running `kd uninstall` / `kd install` will fix the problem , for longer term using JStackTemplate.username instead of JStackTemplate.originId.

File: go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go [hunk 1]
```diff
@@ -1,39 +1,154 @@
...
+	}
+	return err
+}
+
+// AccountByUsername looks up the jAccount by the given username.
+func (c *Client) AccountByUsername(username string) (*models.JAccount, error) {
+	c.init()
+
+	// TODO(rjeczalik): Make a wrapper type for c.accounts and move the lookup there.
+	for _, account := range c.accounts {
```
ppknap: Just a note: is hard to follow the code when you shadow `account` package.

ppknap (approved):

go/src/koding/klientctl/endpoint/remoteapi/remoteapi.go
rjeczalik: Why? The scope is only 3 lines.

ppknap: Entire `AccountByUsername` and `Account` methods use `account` token in a staggered manner. (L73 vs L84)

rjeczalik: Gotcha.

PR #10306, opened 2017-01-12
kd/machine: Add supervised client.

ppknap:
Supervised client type is meant to handle temporary network issues or cases when underlying client is not set up yet. Eg. when KD started but haven't made connection to remote machine.

## How Has This Been Tested?
Unit tests.

Current package coverage: 59.1% of statements

Question: is seems that `klientctl` and `klient` tests haven't been handled by `codecov` yet. @mehmetalisavas are these packages going to be scanned by `codeconv`? 

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/client/supervised.go
+++ b/go/src/koding/klient/machine/client/supervised.go
@@ -112,7 +112,7 @@ func (s *Supervised) call(f func(Client) error) error {
 	ctx, cancel := context.WithTimeout(ctx, s.timeout)
 	defer cancel()
 
-	if _ = <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
+	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
 		// Client is still disconnected. Return it as is.
 		return nil
 	}
--- a/go/src/koding/klient/machine/client/supervised.go
+++ b/go/src/koding/klient/machine/client/supervised.go
@@ -12,7 +12,7 @@ type DynamicClientFunc func() (Client, error)
 
 // Supervised is a decorator type for Clients which will wait for valid client
 // if the current client is Disconnected. This type is meant to handle temporary
-// network issues or cases when underlaying client is not set up yet. Eg. when
+// network issues or cases when underlying client is not set up yet. Eg. when
 // program started but haven't made connection to remote machine yet.
 type Supervised struct {
 	dcf     DynamicClientFunc
@@ -109,7 +109,7 @@ func (s *Supervised) call(f func(Client) error) error {
 		timer := time.NewTimer(s.timeout)
 		select {
 		case <-ctx.Done():
-			// Client changed. Use new client.
+			// Client changed. Use a new one.
 			timer.Stop()
 			c, err = s.dcf()
 			if err != nil {
--- a/go/src/koding/klient/machine/client/supervised.go
+++ b/go/src/koding/klient/machine/client/supervised.go
@@ -88,14 +88,13 @@ func (s *Supervised) MountGetIndex(path string) (idx *index.Index, err error) {
 // canceled context.
 func (s *Supervised) Context() context.Context {
 	c, err := s.dcf()
-	if err == nil {
-		return c.Context()
+	if err != nil {
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		return ctx
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	return ctx
+	return c.Context()
 }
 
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/client/supervised.go
```diff
@@ -0,0 +1,126 @@
...
+		select {
+		case <-ctx.Done():
+			// Client changed. Use a new one.
+			timer.Stop()
+			c, err = s.dcf()
+			if err != nil {
+				return err
+			}
+			return f(c)
+		case <-timer.C:
```
cihangir: instead of using another timer what about creating the context with a deadline?

go/src/koding/klient/machine/client/supervised.go
```diff
@@ -0,0 +1,126 @@
...
+// there is an error during client retrieving, this function will return
+// canceled context.
+func (s *Supervised) Context() context.Context {
+	c, err := s.dcf()
+	if err == nil {
+		return c.Context()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
cihangir: this context is already cancelled when the function returns

rjeczalik: I guess this is due to unreadable branch - one would expect to handle error not success. How about:

```go
func (s *Supervised) Context() context.Context {
	c, err := s.dcf()
	if err != nil {
		ctx, cancel := context.WithCancel(context.Background())
		cancel()
		return ctx
	}

	return c.Context()
}
```

go/src/koding/klient/machine/client/supervised.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (s *Supervised) call(f func(Client) error) error {
+	c, err := s.dcf()
+	if err != nil {
+		return err
+	}
+
+	ctx := c.Context()
+	if err = f(c); err == ErrDisconnected {
+		timer := time.NewTimer(s.timeout)
```
rjeczalik: defer timer.Stop()

go/src/koding/klient/machine/client/supervised.go
ppknap: yes, this is intended to return cancelled context which will prevent any other logic to run (since we don't have a valid client). This type is also going to be extended with cancellation logic so the method will change a little.

ppknap:
@rjeczalik @cihangir gentle ping

go/src/koding/klient/machine/client/supervised.go
```diff
@@ -0,0 +1,127 @@
...
+// will wait until valid client is available or timeout is reached.
+func (s *Supervised) CurrentUser() (user string, err error) {
+	if e := s.call(func(c Client) error {
+		user, err = c.CurrentUser()
+		return err
+	}); e != nil {
+		return "", e
+	}
+
+	return
```
rjeczalik: nit: formatting, feels to me claustrophobic, wdyt?

```go
func (s *Supervised) CurrentUser() (user string, err error) {
    fn := func(c Client) error {
        user, err = c.CurrentUser()
        return err
    }

    _, err = s.call(fn)
    return
}
```

go/src/koding/klient/machine/client/supervised_test.go
```diff
@@ -0,0 +1,67 @@
...
+		_, err = spv.CurrentUser()
+		hitC <- time.Now()
+	}()
+
+	close(startC)
+	serv.TurnOn()
+	select {
+	case <-hitC:
+		if err == client.ErrDisconnected {
+			t.Fatalf("want err != %[1]v; got %[1]v", client.ErrDisconnected)
```
rjeczalik: "want err != client.ErrDisconnected" is enough msg, if it fails it's obvious why

(the real reason is that I always need to lookup the doc whether positional arguments are 0- or 1-index thus I try to avoid it if possible)

ppknap: Actually that's easy to remember:
```go
t.Fatalf(arg0, arg1, arg2, arg3) // everything always starts from 0.

// arg0 is a formatting string so: %[1]v <=> arg1
```

I will change this anyway.

rjeczalik (approved):

mehmetalisavas:
@ppknap i forgot to mention for you that is normally should cover, but there is a problem i think. Im still investigating

ppknap:
@mehmetalisavas no problem, just noticed that and wanted to let you know :+1:

rjeczalik (approved):

rjeczalik:
Merging for now.

@cihangir If you'd have any comments, they're going to be addressed in a separate PR.

cihangir:
LGTM 👍

PR #10320, opened 2017-01-14
Kite api access over social api

gokmen:
Based on the requirements for secure `remote.api` calls I've decided to support `JApiToken`'s with `remote.api` see #10359 Which will provide us better management over api calls. `JApiToken` was designed to serve as team itself not as user but since it has `originId` as creator's `JAccount._id` we can use it for the same purpose. But, to prevent any possible security issues I've updated fetchers for apiTokens to shadow token codes which are not belong to current user. Current user as an admin can revoke (delete) anyone else's token but they cannot see them. For this purpose I've updated some UI around that as well.

image

There is still a limit of `5` token per team but there is no limitation per user, a user can have `10` tokens for different teams also a user can have more than `1` token per team as well.

Once this is done, I'll add support tokens in api routes support into this one which will allow us to use these tokens in anywhere we want. Final result will be like;

`https://your-team.koding.com/remote.api/xxx.../JStackTemplate.build/{JStackTemplate._id}`

where `xxx...` will be the token you've created under Api Tokens section.

I've also updated the format of api tokens as I'm using `uuid-v4` for now and I'll decline requests for other types of apiTokens (current apiToken list in db will be removed)

### DIFF
--- a/workers/social/lib/social/main.coffee
+++ b/workers/social/lib/social/main.coffee
@@ -138,7 +138,7 @@ do ->
 
   options = { rateLimitOptions : KONFIG.nodejsRateLimiter }
 
-  app.post '/remote.api/:model/:id?', (require './remoteapi') koding
+  app.post '/remote.api/:token?/:model/:id?', (require './remoteapi') koding
 
   app.get  '/remote.api', (req, res) ->
     res.send 'REST API is OK'
--- a/workers/social/lib/social/models/apitoken.coffee
+++ b/workers/social/lib/social/models/apitoken.coffee
@@ -126,7 +126,10 @@ module.exports = class JApiToken extends jraphical.Module
   remove$: permit
     advanced: PERMISSION_EDIT_GROUPS
     success: (client, callback) ->
-      @remove callback
+      JSession.remove { clientId: @getAt 'code' }, (err) =>
+        return callback err  if err
+        @remove callback
+
 
   @createSessionByToken = (token, callback) ->
 
--- a/workers/social/lib/social/models/apitoken.coffee
+++ b/workers/social/lib/social/models/apitoken.coffee
@@ -1,4 +1,4 @@
-hat           = require 'hat'
+uuid          = require 'uuid'
 async         = require 'async'
 JGroup        = require './group'
 JAccount      = require './account'
@@ -40,7 +40,7 @@ module.exports = class JApiToken extends jraphical.Module
       code             :
         type           : String
         required       : yes
-        default        : hat
+        default        : uuid.v4
       group            :
         type           : String
         required       : yes
@@ -89,7 +89,6 @@ module.exports = class JApiToken extends jraphical.Module
       (next) ->
         # creating token
         token = new JApiToken
-          code     : hat()
           group    : group
           originId : account.getId()
 
--- a/workers/social/lib/social/models/apitoken.coffee
+++ b/workers/social/lib/social/models/apitoken.coffee
@@ -1,7 +1,5 @@
 uuid          = require 'uuid'
 async         = require 'async'
-JGroup        = require './group'
-JAccount      = require './account'
 jraphical     = require 'jraphical'
 KodingError   = require '../error'
 { permit }    = require './group/permissionset'
@@ -12,6 +10,8 @@ KodingError   = require '../error'
 
 module.exports = class JApiToken extends jraphical.Module
 
+  JGroup     = require './group'
+  JAccount   = require './account'
   Validators = require './group/validators'
 
   @API_TOKEN_LIMIT = 5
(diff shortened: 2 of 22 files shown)


workers/social/lib/social/models/apitoken.coffee
```diff
@@ -120,4 +126,56 @@ module.exports = class JApiToken extends jraphical.Module
...
+          next()
+
+      (next) ->
+        sessionData =
+          clientId  : token
+          username  : account.getAt 'profile.nickname'
+          groupName : group.getAt 'slug'
+          data      : { apiSession: yes }
+
+        JSession.createNewSession sessionData, (err, sessionObj) ->
```
cihangir: instead of `createNewSession` could we use `fetchSessionByData`?

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
+{ Base, secure, signature } = require 'bongo'
+
+KONFIG       = require 'koding-config-manager'
+async        = require 'async'
```
cihangir: async and underscore are not used

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
...
+_            = require 'underscore'
+
+KodingError  = require '../../error'
+KodingLogger = require '../kodinglogger'
```
cihangir: KodingError & KodingLogger not used

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
...
+
+  do ->
+
+    options          =
+      url            : "http://localhost:#{KONFIG.kloud.port}/kite"
+      autoConnect    : yes
+      autoReconnect  : yes
+      transportClass : SockJs
+      auth           :
+        type         : 'kloudctl'
```
cihangir: this isnt kloudctl so we shouldnt use it. @rjeczalik which auth should we use here?

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
...
+        return res
+      .timeout TIMEOUT
+      .catch (err) ->
+        callback err
+
+
+  do ->
+
+    options          =
+      url            : "http://localhost:#{KONFIG.kloud.port}/kite"
```
cihangir: instead of connecting directly, could we connect via nginx at `http://localhost/kloud/kite`?

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
...
+      autoConnect    : yes
+      autoReconnect  : yes
+      transportClass : SockJs
```
cihangir: could we restrict our transport set to the same ones we support on client side, i am not sure if we are testing/supporting all the transports of sockjs ( only xhrlongpolling is widely used atm)

workers/social/lib/social/models/computeproviders/kloud.coffee
```diff
@@ -0,0 +1,69 @@
...
+      auth           :
+        type         : 'kloudctl'
+        key          : KONFIG.kloud.secretKey
+
+    Kloud.transport = new Kite options
```
cihangir: <3

workers/social/lib/social/main.coffee
```diff
@@ -138,7 +138,7 @@ do ->
 
   options = { rateLimitOptions : KONFIG.nodejsRateLimiter }
 
-  app.post '/remote.api/:model/:id?', (require './remoteapi') koding
+  app.post '/remote.api/:token?/:model/:id?', (require './remoteapi') koding
```
cihangir: in some places at our logging infra, we are outputting the urls, having token in the uri might create problems by leaking them. having them in the header would be safer but harder to use.

workers/social/lib/social/models/computeproviders/kloud.coffee
rjeczalik: We can use `kloudSecret` auth here. (https://github.com/koding/koding/pull/10389)

:trollface:

gokmen: #10389 not merged yet, I'll update it once it's in.

workers/social/lib/social/models/computeproviders/kloud.coffee
gokmen: 👍 we only need `xhr-polling` I believe

workers/social/lib/social/main.coffee
gokmen: yeah there is no other way around :| but we do support both in Header and in-Route at least. maybe we can pinpoint locations where we are logging and shadow them but I don't think it would be a big problem.

PR #10323, opened 2017-01-15
Source default supervisord environment variables

szkl:

### DIFF
--- a/deployment/supervisord_initd.sh
+++ b/deployment/supervisord_initd.sh
@@ -44,6 +44,12 @@ ulimit -n 96000
 
 RETVAL=0
 
+# source default env vars
+
+if [ -f /etc/default/supervisord ]; then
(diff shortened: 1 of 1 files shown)


gokmen (approved):

deployment/supervisord_initd.sh
```diff
@@ -44,6 +44,12 @@ ulimit -n 96000
 
 RETVAL=0
 
+# source default env vars
```
cihangir: i am not against this. 
but may be just using `[[ ":$PATH:" != */usr/local/bin* ]] && PATH="/usr/local/bin:${PATH}"` here would be sufficient, just an idea

szkl: I didn't want to involve host environment in the application deployment. We can define other sorts of variables which are only set during image build time and they will be available to all processes.

PR #10327, opened 2017-01-16
kloud/provider: add default.json stack templates per provider 

rjeczalik:
The templates are copied from client/app/lib/providers/templates/
with the one difference that the hardcoded values are replaced
with userInput_* variables, so they can be templated user-side
if needed (e.g. for kd).

/cc @gokmen


go/src/koding/kites/kloud/provider/aws/default.json
```diff
@@ -0,0 +1,20 @@
...
+    "aws": {
+      "access_key": "${var.aws_access_key}",
+      "secret_key": "${var.aws_secret_key}"
+    }
+  },
+  "resource": {
+    "aws_instance": {
+      "aws-instance": {
+        "ami": "",
+        "instance_type": "${var.userInput_instance_type}",
```
gokmen: this should remain as in client example, it shouldn't ask anything by default. let's replace these `userInput`'s to something fixed like in client folder.

gokmen (commented):
I believe we don't need this PR, we should distribute these example contents from single point. and as I said earlier we can do that via an endpoint on socialapi which will read it from single point (client folder for now)

even though we put these files here we shouldn't ask for any information like `userInput_instance_type` these example files should work out of the box I believe.

go/src/koding/kites/kloud/provider/aws/default.json
rjeczalik: There will be a backend for it in kloud, so by default kloud is going to fill it with defaults.

The backend will also have an switch for pass raw templates (with userInput) to kd, so kd can have the interactive `kd init`.

rjeczalik:
> I believe we don't need this PR, 

Then I need to have an endpoint with templates, that have defaults and userInput vars. Basing on current templates it's not possible for kd to guess which field can be inputed and which not.

go/src/koding/kites/kloud/provider/aws/default.json
rjeczalik: Related: https://gist.github.com/usirin/884b568d5566ba964215fa5ac32aaf2f#create

rjeczalik:
Closing in favor of #10329.

PR #10332, opened 2017-01-17
Machine mount sync package - initial version

ppknap:
Depends on: ~#10306~

**Code coverage**: 81.8% of statements

## Architecture
image

## How Has This Been Tested?
Unit tests.

none

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/index/cache.go
+++ b/go/src/koding/klient/machine/index/cache.go
@@ -47,7 +47,7 @@ func (c *Cached) GetCachedIndex(root string) (*Index, error) {
 			}
 		}
 
-		if err = c.cacheIndex(idx, path); err != nil {
+		if err = SaveIndex(idx, path); err != nil {
 			return nil, err
 		}
 	}
@@ -95,35 +95,6 @@ func (c *Cached) getCachedIndex(root string) (idx *Index, path string, err error
 	}
 
 	return idx, path, nil
-}
-
-// cacheIndex atomically saves the provided index under a given path. If the
-// path is empty, this function will create or use existing index temporary
-// directory.
-func (c *Cached) cacheIndex(idx *Index, path string) (err error) {
-	dir, name := filepath.Split(path)
-	f, err := ioutil.TempFile(dir, name+"_")
-	if err != nil {
-		return err
-	}
-
-	if err := json.NewEncoder(f).Encode(idx); err == nil {
-		err = f.Sync()
-	}
-
-	if cerr := f.Close(); cerr != nil && err == nil {
-		err = cerr
-	}
-
-	if err == nil {
-		err = os.Rename(f.Name(), path)
-	}
-
-	if err != nil {
-		os.Remove(f.Name())
-	}
-
-	return err
 }
 
 // createTempPath creates a valid path to index file.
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/index/cache.go
```diff
@@ -184,3 +155,30 @@ func hashSHA1(val string) string {
 	h := sha1.Sum([]byte(val))
 	return hex.EncodeToString(h[:])
 }
+
+// SaveIndex atomically saves the provided index under a given path.
+func SaveIndex(idx *Index, path string) (err error) {
+	dir, name := filepath.Split(path)
+	f, err := ioutil.TempFile(dir, name+"_")
```
rjeczalik: Why do you need `name+"_"`, the file is going to be suffixed either way?

```go
f, err := ioutil.TempFile(filepath.Split(path))
```

rjeczalik:
@ppknap Please rebase.

go/src/koding/klient/machine/mount/sync.go
```diff
@@ -0,0 +1,297 @@
...
+
+	s := &Sync{
+		opts:  opts,
+		syncs: make(map[ID]*synced),
+	}
+
+	if opts.Log != nil {
+		s.log = opts.Log.New("sync")
+	} else {
+		s.log = machine.DefaultLogger.New("sync")
```
rjeczalik: This could be a global var.

rjeczalik: And assigned at L96.

go/src/koding/klient/machine/mount/sync.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// Add starts synchronization between remote and local directories. It creates
+// all necessary cache files if they are not present.
+func (s *Sync) Add(id ID, m Mount) error {
+	s.mu.RLock()
+	sd, ok := s.syncs[id]
+	s.mu.RUnlock()
+
+	if ok {
```
rjeczalik: I'm not sure these fine-grained locks are correct. Two different threads can simultaneously pass this check and both can assign a new *synced value with the same id to the map. While if the whole body was protected with a single lock, the second thread would fail here.

ppknap: the funny fact is that we don't care.. even if IDs are the same (which is not possible in valid logic), the only inconsistency will be that we fetch index twice (or just load from cache) and swap the file atomically. 

This is a trade off I had to make if I want to add mounts concurrently (during kd initialization). `newSync` may be slow and I want to avoid synchronous initialization process. 

I will add:
```go
s.mu.Lock()
if _, ok := s.syncs[id]; ok {
    s.mu.Unlock()
    return //some fancy error here
}
s.syncs[id] = sd
s.mu.Unlock()
```

go/src/koding/klient/machine/mount/sync.go
ppknap: Sorry but I can't see any advantages of this approach. However, I see one big disadvantage which is an introduction of global variable used only in one place as a fallback to incomplete options(used in tests) inside a type that is meant to be used as singleton.

ppknap:
@rjeczalik done

go/src/koding/klient/machine/mount/sync.go
rjeczalik: I do not quite follow, since you're using a global variable either way (`machine.DefaultLogger`) - I just commented to do the same here - use a global variable instead of allocating new default logger each time.

ppknap: The root logger `machine.DefaultLogger` is a parent for other loggers and it uses `config.Konfig.Debug` variable during its initialization.

Actually, I'm avoiding unnecessary allocation by not defining global variable. Please note, that this branch is a fallback which normally will not be executed.

rjeczalik (approved):

go/src/koding/klient/machine/mount/sync.go
```diff
@@ -0,0 +1,301 @@
...
+		return nil, err
+	}
+
+	return s, nil
+}
+
+// mktree ensures that synced working directory is created.
+func (s *synced) mktree() error {
+	dataPath := filepath.Join(s.wd, "data")
+	info, err := os.Stat(dataPath)
```
cihangir: prob MkdirAll is enough, it already handles Stat as fast path https://golang.org/src/os/path.go?s=537:766

cihangir (approved):

go/src/koding/klient/machine/mount/sync.go
ppknap: Thanks - TIL

PR #10339, opened 2017-01-17
kloud/stack: add Variables field to ApplyRequest

rjeczalik:
When non-empty, variables sent with an apply request are
going to be injected into stack template.

/cc @gokmen


ppknap (approved):

rjeczalik:
@gokmen I've also added Variables to plan request - so tf interpolation does not fail due to missing variables.

ppknap (approved):

cihangir:
👍

Issue #10333, opened 2017-01-17
kloud: rework base provider implementation to use remote.api


rjeczalik:
Currently kloud duplicates remote.api functionalities:

- access authorization for JStackTemplate, JMachine and JCredential resources
- reading / updating the JComputeStack and JMachine resources

The implementation also suffers from low test coverage, due to insufficient abstraction around resources operations. Replacing them with remote.api is going to ease improving the test coverage by leveraging `apitest` and `stacktest` packages.

The base provider implementation should be reworked to:

- allow for tracing requests (kloud / socialworker / socialapi / terraformer)
- ensure relevant metrics are collected
- rework plan, apply, authenticate and bootstrap methods to use remote.api

This issue is just an placeholder for future epic.

PR #10340, opened 2017-01-17
stack/provider: ensure user is authorized to access JStackTemplate

rjeczalik:
Temporary solution before #10333 is solved.

/cc @gokmen

### DIFF
--- a/go/src/koding/db/mongodb/modelhelper/stacktemplate.go
+++ b/go/src/koding/db/mongodb/modelhelper/stacktemplate.go
@@ -1,7 +1,9 @@
 package modelhelper
 
 import (
+	"errors"
 	"fmt"
+
 	"koding/db/models"
 
 	"gopkg.in/mgo.v2"
@@ -50,6 +52,39 @@ func GetStackTemplateFieldsByIds(ids []bson.ObjectId, fields []string) ([]*model
 	}
 
 	return stackTmpls, Mongo.Run(StackTemplateColl, query)
+}
+
+// ErrNoAccess is returned:
+//
+//   - by HasTemplateAccess function, when user has no access
+//     to the stack template
+//
+var ErrNoAccess = errors.New("no access for the requested resource")
+
+// HasTemplateAccess return non-nil error if the user is not allowed to access
+// the given stack template. It return nil error otherwise.
+func HasTemplateAccess(tmpl *models.StackTemplate, username string) error {
+	switch tmpl.AccessLevel {
+	case models.AccessPublic:
+		return nil // everyone has access
+	case models.AccessPrivate, "": // if AccessLevel is missing, we assume it's private
+		originID, err := GetAccountID(username)
+		if err != nil {
+			return ErrNoAccess
+		}
+
+		if originID != tmpl.OriginID {
+			return ErrNoAccess
+		}
+	case models.AccessGroup:
+		if ok, err := HasAnyRole(username, tmpl.Group, DefaultRoles...); err != nil || !ok {
+			return ErrNoAccess
+		}
+	default:
+		return errors.New("unrecognized access level: " + tmpl.AccessLevel)
+	}
+
+	return nil
 }
 
 func CreateStackTemplate(tmpl *models.StackTemplate) error {
--- a/go/src/koding/kites/kloud/stack/provider/builder.go
+++ b/go/src/koding/kites/kloud/stack/provider/builder.go
@@ -217,6 +217,20 @@ func (b *Builder) BuildStackTemplate(templateID string) error {
 	return nil
 }
 
+// Authorize verifies whether user is allowed to access b.StackTemplate.
+//
+// Prior calling to this method it is required to build the stack
+// template first.
+//
+// If the stack template was not built, the method is a nop.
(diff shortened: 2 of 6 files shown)


go/src/koding/kites/kloud/stack/provider/builder.go
```diff
@@ -217,6 +217,20 @@ func (b *Builder) BuildStackTemplate(templateID string) error {
...
 }
 
+// Authorize verifies whether user is allowed to access b.StackTemplate.
+//
+// Prior calling to this method it is required to build the stack
+// template first.
+//
+// If the stack template was not built, the method is a nop.
+func (b *Builder) Authorize(username string) error {
+	if b.StackTemplate == nil {
```
ppknap: I'd rather we put this check to `modelhelper.HasTemplateAccess` function because it is always needed before we call it.

ppknap (approved):

go/src/koding/kites/kloud/stack/provider/builder.go
rjeczalik: Actually `modelhelper.HasTemplateAccess` should not be nop if argument is nil - it should panic instead (programmer error). I did not add there explicit check though to be consistent with rest of the methods.

gokmen (approved):

go/src/koding/db/mongodb/modelhelper/stacktemplate.go
```diff
@@ -52,6 +54,39 @@ func GetStackTemplateFieldsByIds(ids []bson.ObjectId, fields []string) ([]*model
...
+		originID, err := GetAccountID(username)
+		if err != nil {
+			return ErrNoAccess
+		}
+
+		if originID != tmpl.OriginID {
+			return ErrNoAccess
+		}
+	case models.AccessGroup:
+		if ok, err := HasAnyRole(username, tmpl.Group, DefaultRoles...); err != nil || !ok {
```
cihangir: `IsParticipant` is helper function for this case ^^

go/src/koding/db/mongodb/modelhelper/stacktemplate.go
```diff
@@ -52,6 +54,39 @@ func GetStackTemplateFieldsByIds(ids []bson.ObjectId, fields []string) ([]*model
...
+			return ErrNoAccess
+		}
+	default:
+		return errors.New("unrecognized access level: " + tmpl.AccessLevel)
+	}
+
+	return nil
```
cihangir: unreachable code

go/src/koding/kites/kloud/stack/provider/builder.go
cihangir: I agree with @ppknap here, it is not always a programmer error, the data(base) itself might be broken

rjeczalik: Even if the data is corrupted, the model itself is never nil, in the worst case it can be zero value.

Either way should we add checks to all functions in modelhelper package that take pointer arguments (models)?

go/src/koding/db/mongodb/modelhelper/stacktemplate.go
ppknap: it's reachable for branches:
`models.AccessPrivate, ""`(err == nil,  and originID == tmpl.OriginID)
`models.AccessGroup`(err != nil || !ok evaluates to false)

go/src/koding/kites/kloud/stack/provider/builder.go
cihangir: `Builder.StackTemplate` could be nil. It's zero value is nil. So HasTemplateAccess could get `tmpl` as nil.


> should we add checks to all functions

i dont think it is that important ATM, but we can fix places whenever we touch near them

go/src/koding/db/mongodb/modelhelper/stacktemplate.go
cihangir: yep you are right, i missed it miserably. tunnel vision, it must be.

go/src/koding/kites/kloud/stack/provider/builder.go
rjeczalik: > Builder.StackTemplate could be nil. It's zero value is nil. So HasTemplateAccess could get tmpl as nil.

In this particular use-case yes, Builder.StackTemplate can be nil. This is why Builder.Authorize checks for nil, not HasTemplateAccess - Builder (caller) is responsible for ensuring correct argument gets passed to HasTemplateAccess.

rjeczalik: My point is that it is not feasible to add in a blanket manner nil-checks to every method that takes pointer argument - the nil-checks should be instead in places where we expect to have nil pointer. We do not expect to have nil argument in HasTemplateAccess, however we expect to have nil argument inside Authorize.

go/src/koding/db/mongodb/modelhelper/stacktemplate.go
rjeczalik: Done in 401f98dbab.

PR #10343, opened 2017-01-18
StackTemplate improvements

### LINKED ISSUE #10329: remote.api: create endpoint for serving default stack templates
Replace hardcoded defaults with userInput variables for each template. Store the defaults separately.

E.g. the endpoint could be `JStackTemplate.default` and take the following request payload:

```json
{
    "provider": "aws",
    "userInput": true
}
```

Response:

```json
{
    "ok": true,
    "error": null,
    "data": {
        "template": "{\"provider\": \"aws\", ...}",
        "userInput": {
            "image": "ubuntu-lts-1404",
            "instance_type": "t2.nano"
        },
    }
}
```

For `"userInput": false` request the method will return a stack template content with the defaults already baked in.

Supersedes #10327.

gokmen:
This PR introduces a new method (`samples`) on `JStackTemplate` which will provide a stack template example for given provider. Supported providers list can be fetched from `ComputeProviders.fetchProviders` (which is also updated to include only stack supported providers). All provider example templates has been updated to include optional values as `userInput`s but for the current client requirements default user input values are also added and there is no change required for client side. `JStackTemplate.samples` provides `json` and `yaml` versions of templates (`yaml` version has comments) and also if `useDefaults` option is passed as `true` templates will be provided with default `user_input`'s replaced.

image

With in this PR I also introduced a way to define return examples in apidocs which can be seen on related commit https://github.com/koding/koding/commit/03f24cd95499a4b88b5d6b41bd0360db6f2eb98b 

### TODO:
- [x] fixed missing stack template tests
- [x] updated providers for stack support and sample templates
- [x] updated template examples to have userInputs optional
- [x] added api endpoint for stack template samples per provider
- [x] added custom return schema support to api-generator
- [x] keep `defaults` under it's name for `userInputs`

fixes: https://github.com/koding/koding/issues/10329


rjeczalik:
How about renaming `user_inputs` -> `defaults` in the response? Since the switch to enable that field is called `useDefaults` in the request.

gokmen:
@rjeczalik how about `defaults: { user_inputs: { ... } }` instead? in that way we can extend it in the future.

rjeczalik:
@gokmen Can we make it `{defaults: {userInput: ...}}` so it can be easily matched with `var.userInput`?

szkl (approved):

gokmen:
all done ptal, also added tests for `clone` operations for this one https://github.com/koding/koding/pull/10346 cc/ @rjeczalik @szkl

cihangir:
👍

Issue #10187, opened 2016-12-31
Other user cannot view the changes on a file in some cases which may be caused by the errors on the console.


diclee:
## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/21575269/b7e5a2d4-cebb-11e6-970c-d36692e7ce7c.png)

error log:
![image](https://cloud.githubusercontent.com/assets/8261892/21575277/fc8e698e-cebb-11e6-9a89-242cc5d0e6ad.png)

![image](https://cloud.githubusercontent.com/assets/8261892/21575283/2beae694-cebc-11e6-9891-383c3ec2eb5b.png)

diclee:
It's happening on sandbox. Don't know if reproducible on latest or not, because I cannot test on latest due to #10144

diclee:
This is also reproducible on latest and production.

diclee:
Also after some changes made on a file, the file becomes uneditable by throwing the above errors with every key stroke.

diclee:
Verify passed on sandbox.

PR #10363, opened 2017-01-20
Remove editor property from line widget object

szkl:
Fixes https://github.com/koding/koding/issues/10187


szkl:
Editor is [destroyed](https://github.com/ajaxorg/ace/blob/master/lib/ace/line_widgets.js#L247-L248) if it is set as a property of the line widget object.

gokmen (approved):

PR #10372, opened 2017-01-20
Check value existence to determine return value

szkl:


cihangir (approved):

node_modules_koding/koding-config-manager/config.coffee
```diff
@@ -28,7 +28,7 @@ traverse.forEach KONFIG, (node) ->
       else if val is parsedVal.toString()
         val = parsedVal
 
-  return val or node
+  return if val? then val else node
```
gokmen: `return val ? node`

PR #10393, opened 2017-01-23
Payment trial fixes

cihangir:


cihangir:
This PR looks better without whitespaces here

https://github.com/koding/koding/pull/10393/files?diff=unified&w=1

client/app/lib/redux/modules/payment/constants.coffee
```diff
@@ -18,9 +18,7 @@ module.exports =
 
   Plan:
     FREE: 'p_free'
-    FREE_FOREVER: 'p_free_forever'
```
cihangir: cc/ @usirin

usirin: 👍

go/src/socialapi/workers/payment/subscription.go
```diff
@@ -169,6 +169,27 @@ func EnsureSubscriptionForGroup(groupName string, params *stripe.SubParams) (*st
...
 }
 
+func normalizeTrialEnd(now time.Time, trialEnd int64) int64 {
+	if trialEnd == 0 {
+		return trialEnd
+	}
+
+	if trialEnd < 0 {
+		return 0
+	}
```
rjeczalik: The two above could be folded into one:

```go
if trialEnd <= 0 {
    return 0
}
```

go/src/socialapi/workers/payment/subscription.go
```diff
@@ -121,20 +121,20 @@ func EnsureSubscriptionForGroup(groupName string, params *stripe.SubParams) (*st
...
-		if params.TrialEnd < sevenDaysLater {
-			params.TrialEnd = sevenDaysLater
-		}
+	// if group is not created within last ten minutes, subtract its duration from the trial end date
+	groupCreatedAt := group.Id.Time().UTC()
 
-		if params.TrialEnd > sevenDaysLater {
-			params.TrialEnd = thirtyDaysLater
-		}
+	if params.TrialEnd != 0 && !groupCreatedAt.Add(time.Minute*5).After(now) {
```
rjeczalik: The 5 min is some grace period?

go/src/socialapi/workers/payment/plan.go
```diff
@@ -21,79 +18,78 @@ const (
...
 // CreateDefaultPlans creates predefined default plans. This is meant to be run
 // when the worker starts to be sure the plans are there.
 func CreateDefaultPlans() error {
-	for key := range Plans {
-		if err := EnsurePlan(Plans[key]); err != nil {
+	planMu.RLock()
+	defer planMu.RUnlock()
+
+	for key := range plans {
+		if err := EnsurePlan(plans[key]); err != nil {
```
rjeczalik: Could be:

```go
for _, plan := range plans {
    if err := EnsurePlan(plan); err != nil {
...
```

rjeczalik (approved):

go/src/socialapi/workers/payment/plan.go
```diff
@@ -21,79 +18,78 @@ const (
...
-	UpTo50Users = "p_up_to_50"
-	Over50Users = "p_over_50"
+	Free    = "p_free"
+	Solo    = "p_solo"
+	General = "p_general"
 )
 
-// Plans holds koding provided plans on stripe
-var Plans = map[string]*stripe.PlanParams{
+var planMu sync.RWMutex
```
ppknap: Do we need this mutex? I can't see any places where the map is modified, There are also Read Locks only.

cihangir: i was also under impression that mutexes were not required hence i didnt add at first place but in wercker i found some race detections

go/src/socialapi/workers/payment/subscription.go
cihangir: yep, even 5 mins is a big duration, but i dont have any scientific way to measure this, i just guestimated some big random duration

go/src/socialapi/workers/payment/plan.go
ppknap: do you have stack traces? If there are data races and we didn't add Write locks, this mutex will not help (however it may slow down go runtime by adding barriers, thus preventing race detector from detecting still existing races)

Just asking, maybe TIwillL :)

rjeczalik: Wouldn't the race detector report races, if a global map is access from init function?

cihangir: i dont have the stack traces ATM, i saw them while back ago, not while implementing this feature. if you guys want i can remove these and wait for that happen again?

rjeczalik: > if you guys want i can remove these and wait for that happen again?

sgtm

ppknap: > Wouldn't the race detector report races, if a global map is accessed from init function?

Why it should do so, inits are guaranteed to run synchronically. That's the reason running goroutnes from init is discouraged.

> if you guys want i can remove these and wait for that happen again?

:+1:

ppknap (approved):

cihangir:
it seems everyone approved, merging this PR.

PR #10422, opened 2017-01-25
Cache "has card" info in the group.

cihangir:
Checking if team has CC takes about 400ms in each page load, cache it in the group so we will have a faster way to lookup it. If group does not have the info, fetch it from the socialapi, it will return up-to-date info while updating the database.


usirin:
@cihangir Are we disabling the lookup? I mean is this gonna be backwards compatible? If so i will need to make changes before we ship this to production.

cihangir:
@usirin it is backward compatible, if it cant find in the group data, it will fetch from the API, and API will update group with that data, so consecutive lookups will have the data. it is automigrating itself.

ppknap (approved):

usirin (approved):

PR #10434, opened 2017-01-27
klient/machine: improve indexing performance with concurrent file scanning

ppknap:
First step when creating a mount is to create the index from remote FS. For entire Koding repository we have following numbers:

**Scanning time**:  43.349127583s
**Files**:  212972
**Files time**:  2.838µs
**Files Disk Size**:  3.6 GiB
**Files Disk Size time**:  16.514726ms

The most time consuming operation is SHA-1 sum of stored files (which is expected and cached after the first scan). This operation can be made concurrently and that's what this PR introduces. Time measurements after the change:

**Scanning time**:  19.260137022s (improvement: **225%**)
**Files**:  212972
**Files time**:  8.605µs
**Files Disk Size**:  3.6 GiB
**Files Disk Size time**:  16.457452ms

## Motivation and Context
Decrease mount initialization time (which currently times out after one minute - due to Kite connection deadline).

## How Has This Been Tested?
Existing unit tests passed.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/index/index.go
+++ b/go/src/koding/klient/machine/index/index.go
@@ -23,7 +23,7 @@ type Entry struct {
 	MTime int64       `json:"mtime"` // File data change time since EPOCH.
 	Mode  os.FileMode `json:"mode"`  // File mode and permission bits.
 	Size  int64       `json:"size"`  // Size of the file.
-	SHA1  []byte      `json:"sha1"`  // SHA-1 hash of file content.
+	Hash  []byte      `json:"hash"`  // Hash of file content.
 }
 
 // NewEntryFile creates new Entry from a file stored under path argument.
@@ -41,7 +41,7 @@ func NewEntryFile(root, path string, info os.FileInfo) (e *Entry, err error) {
 		return nil, err
 	}
 
-	// Compute file's SHA-1 sum.
+	// Compute file's hash sum.
 	var sum []byte
 	if !info.IsDir() {
 		if sum, err = readCRC32(path); err != nil {
@@ -55,7 +55,7 @@ func NewEntryFile(root, path string, info os.FileInfo) (e *Entry, err error) {
 		MTime: info.ModTime().UnixNano(),
 		Mode:  info.Mode(),
 		Size:  info.Size(),
-		SHA1:  sum,
+		Hash:  sum,
 	}, nil
 }
 
--- a/go/src/koding/klient/machine/index/index.go
+++ b/go/src/koding/klient/machine/index/index.go
@@ -1,8 +1,8 @@
 package index
 
 import (
-	"crypto/sha1"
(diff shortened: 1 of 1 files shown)


cihangir:
> most time consuming operation is SHA-1 sum

what about using another hashing algo while having 10x more performance 😛

rjeczalik:
And in case we need parity with git index we can force index recreation

Cihangir <notifications@github.com> schrieb am Fr. 27. Jan. 2017 um 10:01:

go/src/koding/klient/machine/index/index.go
```diff
@@ -135,22 +140,41 @@ func NewIndexFiles(root string) (*Index, error) {
...
+	wg.Wait()
 	return idx, nil
 }
 
+// addEntry asynchronously adds new entry to index. This function has limits
+// to a number of started go-routines. Errors are ignored.
+func (idx *Index) addEntry(wg *sync.WaitGroup, root, path string, info os.FileInfo) {
+	idx.limitC <- struct{}{}
+
+	go func() {
```
cihangir: could we change the structure here a bit with goroutine worker pool pattern? Instead of creating goroutines per file operation we can limit it by `2*runtime.NumCPU()` 

just to clarify what i mean by [worker pool](https://gobyexample.com/worker-pools)

ppknap:
> what about using another hashing algo while having 10x more performance :stuck_out_tongue:

> And in case we need parity with git index we can force index recreation

OK, lets use crc32 as @cihangir sugessted [here](https://github.com/koding/koding/pull/10134#discussion_r94128683). We version the index type so it should be easy to switch anyway.

ppknap:
> could you also send the code for measuring these?

@cihangir this is rather a dummy program than a real benchmark (I planning to add real bench when we have more time) but if you are interested, the code is [here](https://gist.github.com/ppknap/e4e9e68c46268555b5fb13d3e16b2f63)

go/src/koding/klient/machine/index/index.go
ppknap: @cihangir no problem at all - this was the simplest solution. I will left `chan Type` as a member of `index`in order to always have max `2*NumCPU` goroutins running simultaneously (eg when doing N concurrent calls to `Apply`).

ppknap:
@cihangir - after switching to CRC-32 i got following results:
**SHA-1**:  13.191289437s
**CRC-32**:  11.610115897s

This was surprising so I profiled the index and most of the time spent is I/O during hashing. After using byte pools: 

**CRC-32 + BytePool**: Scanning time:  7.604643191s

I'm stopping at this point - need to focus on serialized file size optimization. Moreover, rest of the time is IO:
image

cihangir (approved):
💯

rjeczalik (approved):

PR #10439, opened 2017-01-28
Split client side bundle into css & third party vendor module bundles

usirin:
This is the initial effort to create separate bundles for load optimizations. It introduces 2 main changes:

- Uses a util called `isExternal` to determine if a required module is an external module (meaning under `client/node_modules` folder).
- uses [`CommonsChunkPlugin`](https://webpack.js.org/plugins/commons-chunk-plugin/) to create 2 main bundles (`bundle.main.js` and `bundle.vendor.js`) by utilizing `isExternal` util.
- uses [`ExtractTextPlugin`] to create a separate bundle for style files.
  - Since we now have 2 different bundles (see above) we make sure that we have 2 different style bundles (`bundle.main.css` and `bundle.vendor.css`) as well, using `allChunks: true` option of `ExtractTextPlugin`

----

### Before

image

### After

image

----

`bundle.main.js` and `bundle.main.css` will be our main application code separate from the third party vendor code we are using, we can ship them without changing anything.

As long as we make sure that our `node_modules` are not being updated, both `bundle.vendor.js` and `bundle.vendor.css` can be cached regardless of the version of `koding` we are shipping, some brainstorm would be really good in this subject. /cc @cihangir @szkl 

There are 2 main way of doing this in my mind:
- Since we already cache our `node_modules`, we can cache vendor bundles as long as cache is still valid.
- We can make a checksum comparison of vendor bundles (both css, and js) as `after build step`, and then we can invalidate the vendor cache if comparison fails.

What do you guys think?

### DIFF
--- a/client/app/lib/shortcutscontroller.coffee
+++ b/client/app/lib/shortcutscontroller.coffee
@@ -1,4 +1,4 @@
-Shortcuts      = require 'shortcuts'
+Shortcuts      = require '@koding/shortcuts'
 _              = require 'lodash'
 kd             = require 'kd'
 globals        = require 'globals'
--- a/client/ide/lib/collaborationcontroller.coffee
+++ b/client/ide/lib/collaborationcontroller.coffee
@@ -1,10 +1,7 @@
 _                             = require 'lodash'
-sinkrow                       = require 'sinkrow'
 kd                            = require 'kd'
-KDModalView                   = kd.ModalView
 FSFile                        = require 'app/util/fs/fsfile'
 nick                          = require 'app/util/nick'
-getCollaborativeChannelPrefix = require 'app/util/getCollaborativeChannelPrefix'
 showError                     = require 'app/util/showError'
 whoami                        = require 'app/util/whoami'
 RealtimeManager               = require './realtimemanager'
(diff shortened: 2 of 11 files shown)


szkl:
> Uses a util called isExternal to determine if a required module is an external module (meaning under client/node_modules file.

If it is under client/node_modules directory then it should be referred as "dependency".

usirin:
> If it is under client/node_modules directory then it should be referred as "dependency".

I think "dependency" is really a more generic term than it needs to be, a local modal can be a dependency of another module as well. What do you think about `isExternalDependency`?

szkl:
I think "external" term is too vague. `isVendorDependency` or `isNpmDependency` are a better choice. If you're going to name output files as vendor then former suits better.

usirin:
👍

szkl:
What caching are you referring to? Browser or build time?

usirin:
Browser.

usirin:
But i am pretty sure build time caching can be enabled as well with the same strategy we would agree on for browser caching, but this time for webpack.

usirin:
if cloudflare has a smart way of caching content though, my thoughts are obviously irrelevant.

gokmen:
This is a great improvement @usirin 👍 I believe we can split out the vendor as well if it's a problem as I'm seeing it's causing most of the size issue here. We can pre-defined base bundles for big chunks like Ace, React etc. instead of doing it dynamically with all others. Then we can have another bundle for 3rd parties. I think this can be handled better since major requirements (Ace, React) will remain as is most of the time.

cihangir:
> As long as we make sure that our node_modules are not being updated, both bundle.vendor.js and bundle.vendor.css can be cached regardless of the version of koding we are shipping, some brainstorm would be really good in this subject. /cc @cihangir @szkl


we can md5 it and put under a/p/p directly, so that koding git version wont be in the url. if consecutive koding deploys doesnt change the contents of vendor bundles, md5 will be same so cache wont be invalidated. trying the guess if it changed or not from package.json or from node_modules folder would be cumbersome.

szkl:
I agree with that. Build output can change without modifying any dependency.

gokmen (approved):

PR #10445, opened 2017-01-29
Small fix to not ignore errors from tunnelproxy addClientService

d6o:
## Description
Fixing code because a multierror object was instanced but was never used and errors was being ignored in the loop.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/go/src/koding/kites/tunnelproxy/server.go
+++ b/go/src/koding/kites/tunnelproxy/server.go
@@ -338,7 +338,7 @@ func (s *Server) addClient(ident, name, vhost string, services map[string]*Tunne
 
 		e := s.tunnels.addClientService(ident, tun)
 		if e != nil {
-			e = multierror.Append(e, err)
+			err = multierror.Append(err, e)
 		}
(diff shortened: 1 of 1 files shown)


go/src/koding/kites/tunnelproxy/server.go
```diff
@@ -338,7 +338,7 @@ func (s *Server) addClient(ident, name, vhost string, services map[string]*Tunne
 
 		e := s.tunnels.addClientService(ident, tun)
 		if e != nil {
-			e = multierror.Append(e, err)
+			err = multierror.Append(err, e)
```
rjeczalik: Could you also change L333 to:

```go
var err error
```

And simplify L345 to:

```go
if err != nil {
```

DiSiqueira: Thank you for the suggestion, Pull Request updated

rjeczalik (approved):

PR #10446, opened 2017-01-29
Add tests to errors variables

d6o:
## Description
auth_test.go, controller.go, account.go and fusetest_file.go
Adding validation test to err var

user_test.go
underscoring unused user var

channel_test.go
Added test when same message is sent twice, in second time ch should be nil.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/go/src/socialapi/workers/api/modules/message/controller.go
+++ b/go/src/socialapi/workers/api/modules/message/controller.go
@@ -255,6 +255,10 @@ func Delete(u *url.URL, h http.Header, _ interface{}, c *models.Context) (int, h
 
 		// delete the message here
 		err = cm.DeleteMessageAndDependencies(false)
+		if err != nil {
+			return response.NewBadRequest(err)
+		}
+
 		// then invalidate the cache of the parent message
 		bongo.B.AddToCache(parent)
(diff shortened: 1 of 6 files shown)


go/src/socialapi/workers/api/modules/message/controller.go
```diff
@@ -255,6 +255,10 @@ func Delete(u *url.URL, h http.Header, _ interface{}, c *models.Context) (int, h
 
 		// delete the message here
 		err = cm.DeleteMessageAndDependencies(false)
+		if err != nil {
+			return response.NewBadRequest(err)
```
rjeczalik: I don't think it's correct - it changes the flow: previously if deleting message and deps failed, we still invalidated the cache. Right not we do not.

DiSiqueira:
You are right @rjeczalik!! Rereading the code and this test in Message Controller wasn't correct, i reverted this change

rjeczalik (approved):

rjeczalik (approved):

cihangir (approved):
👍

PR #10453, opened 2017-01-30
kloud: escape variables in comments / stub payload_ variables

rjeczalik:
This PR:

- escapes variables from commented-out lines in user\_data
- stubs ${var.payload_*} during plan request, so no need to send blanks explicitely

### DIFF
--- a/go/src/koding/kites/kloud/stack/provider/variable.go
+++ b/go/src/koding/kites/kloud/stack/provider/variable.go
@@ -92,6 +92,7 @@ var escape = func(v *Variable) string {
 func EscapeDeadVariables(userdata string) string {
 	var buf bytes.Buffer
 
+	eofNL := strings.HasSuffix(userdata, "\n")
 	scanner := bufio.NewScanner(strings.NewReader(userdata))
 
 	for scanner.Scan() {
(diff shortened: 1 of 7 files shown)


go/src/koding/kites/kloud/stack/provider/variable.go
```diff
@@ -67,3 +81,40 @@ func ReplaceVariables(s string, vars []Variable, blank string) string {
...
+		fmt.Fprintln(&buf, ReplaceVariablesFunc(s, ReadVariables(s), escape))
+	}
+
+	// Scanning from strings.Reader is not going to fail, even if, there's
+	// no recovery from the failure.
+	_ = scanner.Err()
+
+	if eofNL {
+		return buf.String()
+	}
```
cihangir: here is something wrong with the logic. if it ends with new line, which is also true for the following case, we dont care about it and return the whole buffer, but if it doesnt have, we apply "TrimRight". In any case "Fprintln" will add new lines. What about removing this logic and using "TrimSpace"? Or am i missing something?

rjeczalik: Not exactly:

If original value ended with a new line - we preserve it. If it doesn't - we trim it, because as you observed - we always write a new line for each of the scanned lines.

rjeczalik: If we did not trim it, the following case would need to have a new line at the end:

https://github.com/koding/koding/pull/10453/files#diff-837eac487945eef9374603db1a552ff6R111

Since the function is only about escaping, it could be unexpected if it also appended a new-line.

ppknap (approved):

go/src/koding/kites/kloud/stack/provider/variable.go
cihangir: > Since the function is only about escaping, it could be unexpected if it also appended a new-line.

you got me

gokmen:
So, should we take in?

rjeczalik:
If @cihangir is ok with it, then 🚢 .

rjeczalik:
Merging for now. If there's anything more, I'll address in next PR.

cihangir:
LGTM

PR #10454, opened 2017-01-31
Improve JSON-encoded index size.

ppknap:
This PR introduces a few non-breaking changes that are meant to reduce index size after serializing to JSON format.

- JSON tags were shortened to 1 character.
- File name is stored only as a map key - no longer as map value member.
- When JSON marshaling, index is gzipped and base64 encoded, then sent as a string.
- Index cache will not rescan file tree if `Rescan` time will be greater than cached file age.

These changes results in:

- JSON encoded index size for Koding repo, decreased from 81MB to 6.1MB (**1328%**).
- Marshaling time increased from 1.31s to 1.63 (**24%**)
- Retrieving index from cache decreased from: 4.55s to 1.42s

Also, I was trying to use gob encoding but it resulted with greater JSON object size (8MB).

## Motivation and Context
Reduce the amount of data that needs to be sent between local and remote machine during mount initialization.

## How Has This Been Tested?
Existing Unit tests + added the case for Rescan.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/index/cache.go
+++ b/go/src/koding/klient/machine/index/cache.go
@@ -9,12 +9,18 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
 const tempIndexDirPrefix = "koding_index_"
 
 // Cached allows to cache and reuse previously created index.
 type Cached struct {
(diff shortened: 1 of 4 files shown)


rjeczalik (approved):

cihangir (approved):
💯

PR #10470, opened 2017-02-01
Code Improvement for Api Access Component

hakankaradis:


client/home/lib/utilities/components/apiaccess/container.coffee
```diff
@@ -21,14 +21,22 @@ module.exports = class ApiAccessContainer extends React.Component
...
-    super props
 
   render: ->
 
     <View
       apiTokens={@state.apiTokens}
-      apiAccessState={@state.apiAccessState}/>
+      apiAccessState={@state.apiAccessState}
+      switchToggle={@bound 'switchToggle'}
+      addNewApiToken={@bound 'addNewApiToken'}  />
```
usirin: Let's prefix these with `on`, eg: `onSwitchToggle`, `onAddNewToken` to make them more obvious that they are event bindings.

client/home/lib/utilities/components/apiaccess/view.coffee
```diff
@@ -42,27 +35,14 @@ module.exports = class View extends React.Component
...
-
   render: ->
 
     toggleState = @props.apiAccessState
 
     <div className='HomeApp-ApiToken'>
 
-      <Toggle checked={toggleState} className='HomeApp-ApiToken--swicth-toggle' callback={@bound 'switchToggle'} />
-      <Header callback={@bound 'addNewApiToken'}/>
+      <Toggle checked={toggleState} className='HomeApp-ApiToken--swicth-toggle' callback={@props.switchToggle} />
```
usirin: Can you define a `toggleProps` object and use it like `<Toggle {...toggleProps}>` to make it more readable (80 chars per line rule)

client/home/lib/utilities/components/apiaccess/view.coffee
```diff
@@ -76,13 +56,12 @@ module.exports = class View extends React.Component
       <div className='HomeApp-ApiToken--footer'>
         <div className='HomeApp-ApiToken--footer--button-wrapper'>
           <GuideButton />
-          <AddNewApiTokenButton toggleState={toggleState} callback={@bound 'addNewApiToken'}/>
+          <AddNewApiTokenButton toggleState={toggleState} callback={@props.addNewApiToken} />
```
usirin: Same here, let's define `buttonProps` and change this line to `<AddNewApiTokenButton {...buttonProps}>`

client/app/lib/flux/teams/actions.coffee
```diff
@@ -315,21 +316,26 @@ deleteApiToken = (apiTokenId) ->
...
   { groupsController, reactor } = kd.singletons
   team = groupsController.getCurrentGroup()
-  new Promise (resolve, reject) ->
 
-    team.modify { isApiEnabled : state }, (err) ->
-      return reject { err }  if err
-      reactor.dispatch actions.SET_API_ACCESS_STATE, { state }
+  team.modify { isApiEnabled : state }, (err) ->
+
+    return showError err  if err
```
usirin: I am not sure if this is the correct place to show an error. returning a promise from this action, and putting `showError` calls into rejection handler looks better to me as this is a view problem.

usirin: If i remember correctly this should already be returning a promise, since this is a bongo call.

client/app/lib/flux/teams/actions.coffee
```diff
@@ -315,21 +316,26 @@ deleteApiToken = (apiTokenId) ->
...
 
 
-addApiToken = (apiToken) ->
+addApiToken = ->
 
   { reactor } = kd.singletons
-  reactor.dispatch actions.ADD_API_TOKEN_SUCCESS, { apiToken }
+  remote.api.JApiToken.create (err, apiToken) ->
+
+    return showError err  if err
```
usirin: same here, let's remove this from here, and put it into the promise's rejection handler.

usirin (changes requested):
Let's fix it for once and for good :)

client/app/lib/flux/teams/actions.coffee
hakankaradis: I think there is no need to use Promise here, because we are not returning anything to view component, if there is an error, showError will show it 🤔

usirin: We decided to do this later with all the modules which are using view related utils.

usirin (approved):

PR #10480, opened 2017-02-02
Revert "go/github.com: add shlex package for gometalinter"

cihangir:
Reverts koding/koding#10479


cihangir:
glide is required for 3rd party packages. i missed that part while merging

Issue #9997, opened 2016-12-12
Unify remote klient clients.


ppknap:
Currently, there are at least three clients that call klient's kite:

- `KlientPool` [link](https://github.com/koding/koding/blob/master/go/src/koding/kites/kloud/klient/klient.go#L24)- looks like it is not used anywhere.
- `vagrantapi.Klient` [link](https://github.com/koding/koding/blob/master/go/src/koding/kites/kloud/api/vagrantapi/vagrantapi.go#L95)
- `remote.KodingClient` [link](https://github.com/koding/koding/blob/master/go/src/koding/klient/remote/getkodingkites.go#L33) - this client will be deprecated once we implement all machine related commands.

It would be nice to have one client for Klient that implements all its [methods](https://github.com/koding/koding/blob/master/go/src/koding/klient/app/klient.go#L441-L527)

rjeczalik:
I'd also throw to the wish-list:

- the unified klient should support one transport interface ([example](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/endpoint/kloud/kloud.go#L32-L35)), [instead](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/klient/klient.go#L34) [of](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/klient/klient.go#L34) [mocking](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/unmount.go#L44) [kd](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/shortcut/machinename.go#L12) [with](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/repair/kontrolrepair.go#L16) [untyped](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/open/command.go#L35) [interfaces](https://github.com/koding/koding/blob/6fef6570e62e1d377b0282c5207fa374509137ae/go/src/koding/klientctl/repair/klientrunningrepair.go#L17)

[...]

PR #10499, opened 2017-02-04
klient/machine: add DiskBlocks method to dynamic Client

ppknap:
Filesystem block information is required by FUSE. This PR just extends machine client to this (already implemented) method.

## Motivation and Context
Required by mounts.

## How Has This Been Tested?
No test - it's just a connector to already implemented logic.

### DIFF
--- a/go/src/koding/kites/kloud/klient/klient.go
+++ b/go/src/koding/kites/kloud/klient/klient.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"koding/klient/fs"
 	"koding/klient/machine/index"
 	"koding/klient/sshkeys"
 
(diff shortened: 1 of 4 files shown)


rjeczalik (approved):
💯

cihangir (approved):

ppknap:
@cihangir get disk info is already tested [here](https://github.com/koding/koding/blob/9f8c0a80a262ee35a2652bde62de65f0a8a5e685/go/src/koding/klient/fs/fs_test.go#L712)

Rest of the changes are just mocks, test helpers, and kite call to already tested function.

However, I agree that the used [KlientPool](https://github.com/koding/koding/blob/9f8c0a80a262ee35a2652bde62de65f0a8a5e685/go/src/koding/kites/kloud/klient/klient.go) should be tested separately. There is an issue for that: #9997

PR #10502, opened 2017-02-06
klient/machine: move mount group to machine group

ppknap:
This PR doesn't add or remove any logic, it just improves the consistency of `machine` logic.

## Motivation and Context
It was hard to iterate over merged multi-mount syncers. Event queue will be used per mount which would be hard to implement in previous structure where `synced`(one mount synchronization logic) was private and inside package that manages multi mounts.

## How Has This Been Tested?
Logic didn't change. - Existing Unit tests passed.


ppknap:
@cihangir @rjeczalik don't hesitate to just briefly look through the code. There are no important changes in current logic there.

rjeczalik (approved):

cihangir (approved):

cihangir:
@ppknap if you push the rename/moves in one commit git shows it as proper operation, just fyi

ppknap:
@cihangir I know, there is one commit. There were some renaming/removing/splitting in these files and git didn't want to make better diffs even with explicit `git mv` at the beginning.

Issue #10495, opened 2017-02-04
The members/admins don't get the notification for role changes when their role is updated.


diclee:

diclee:
Verify passed on sandbox & latest.

PR #10503, opened 2017-02-06
The members/admins don't get the notification for role changes when their role is updated.

hakankaradis:
Updated the place where we were listening the `MembershipRoleChanged` event of groupscontroller.

Fixes: #10495

http://recordit.co/uk5PJJoRPo

### DIFF
--- a/client/app/lib/mainviewcontroller.coffee
+++ b/client/app/lib/mainviewcontroller.coffee
@@ -1,6 +1,8 @@
 globals = require 'globals'
 checkFlag = require './util/checkFlag'
 kd = require 'kd'
+whoami = require './util/whoami'
+MembershipRoleChangedModal =  require 'app/components/membershiprolechangedmodal'
 
 
 module.exports = class MainViewController extends kd.ViewController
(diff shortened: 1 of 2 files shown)


client/app/lib/maincontroller/groupscontroller.coffee
```diff
@@ -42,6 +44,19 @@ module.exports = class GroupsController extends kd.Controller
...
+        { contents: { role, id, adminNick } } = data
+        reactor.dispatch 'UPDATE_TEAM_MEMBER_WITH_ID', { id, role }
+
+        if id is whoami()._id
+          modal = new MembershipRoleChangedModal
+            success: ->
+              modal.destroy()
+              global.location.reload yes
+          , { role, adminNick }
+
```
usirin: `GroupsController` isn't the place where we should do such things. I'd prefer this to be in either `MainView` or `MainViewController`. wdyt @gokmen ?

gokmen (changes requested):
Let's move UI block to `MainViewController` like @usirin suggested https://github.com/koding/koding/pull/10503/files#r99530195

gokmen (approved):

PR #10507, opened 2017-02-06
Enable adding users to group via JGroup model

usirin:
Therefore allow it to be used by `remote.api`


workers/social/lib/social/models/group/index.coffee
```diff
@@ -385,6 +386,84 @@ module.exports = class JGroup extends Module
...
 
 
+  # joinUser
+  #
+  # Joins user with given options to group either by logging in or converting
+  # them.
+  #
+  # @return {ResponseWithToken}
+  #
+  @joinUser = (options, callback) ->
```
cihangir: if you just add secure here, you will get the client, importing server/bongo is not required

workers/social/lib/social/models/group/index.test.coffee
```diff
@@ -576,6 +576,74 @@ runTests = -> describe 'workers.social.group.index', ->
...
+
+        withConvertedUser options, ({ client, account, group }) ->
+
+          joinOptions =
+            username: generateRandomString 8
+            password: generateRandomString()
+            email: "#{generateRandomString 12}@#{generateRandomString 6}.com"
+            slug: groupData.slug
+
+          JGroup.joinUser joinOptions, (err, result) ->
```
cihangir: infact this should fail. because we require an "invitation"

workers/social/lib/social/models/group/index.test.coffee
```diff
@@ -576,6 +576,74 @@ runTests = -> describe 'workers.social.group.index', ->
...
+          withConvertedUser options, ({ group }) ->
+
+            joinOptions =
+              username: userFormData.username
+              password: userFormData.password
+              email: userFormData.email
+              slug: group.slug
+
+            # then join the initially created user to our new api enabled group
+            JGroup.joinUser joinOptions, (err, result) ->
```
cihangir: same here

workers/social/lib/social/models/group/index.coffee
cihangir: could you also clean up the server handler of jointeam.coffee?

cihangir (changes requested):
clean up & bongo secure

workers/social/lib/social/models/group/index.test.coffee
usirin: it's working because `allowedDomains` is set to `['*']`

workers/social/lib/social/models/group/index.test.coffee
usirin: same here :D

workers/social/lib/social/models/group/index.coffee
usirin: :+1:

scripts/api-generator.coffee
```diff
@@ -76,6 +76,23 @@ swagger =
...
+        error:
+          type: 'Error'
+          description: 'Error description'
+          example:
+            message: 'Something went wrong'
+            name: 'SomethingWentWrong'
+        data:
+          type: 'object'
+          description: 'Result of the operation'
+          example: { token: 'some-uuid-string' }
```
gokmen: we don't need to hard code a response if it's only used by one method, instead you can define your example response in place like https://github.com/koding/koding/blob/master/workers/social/lib/social/models/computeproviders/stacktemplate.coffee#L325-L351

workers/social/lib/social/models/group/index.coffee
```diff
@@ -385,6 +386,54 @@ module.exports = class JGroup extends Module
...
 
+  # joinUser
+  #
+  # Joins user with given options to group either by logging in or converting
+  # them.
+  #
+  # @return {ResponseWithToken}
+  #
+  @joinUser = ->
+  @joinUser = secure (client, options, callback) ->
```
gokmen: this will allow anyone to join anyone to any group.

gokmen (changes requested):
Currently it's open public for such important actions like adding someone to some group. Let's have a session about it first.

usirin:
@cihangir @gokmen i updated the tests to cover all the cases can `JGroup.joinUser` method, which is basically moving the logic from jointeam handler to this method. 

@gokmen do you think these tests answer your concerns as well?

workers/social/lib/social/models/group/index.coffee
usirin: Tests should show that it's working exactly the same as `jointeam` handler, so it's not open to everyone as long as `allowedDomains` property is not set to `['*']`

gokmen:
@usirin nope, none of the test can answer my concern while this thing stays as is https://github.com/koding/koding/pull/10507/files#diff-23aa32e384f44c80809d4a6cb4c27095R397 this means "anyone" logged in to "any" team can execute this method and can add "anyone (with password)" to "any" team.

gokmen (approved):

gokmen:
> @usirin nope, none of the test can answer my concern while this thing stays as is https://github.com/koding/koding/pull/10507/files#diff-23aa32e384f44c80809d4a6cb4c27095R397 this means "anyone" logged in to "any" team can execute this method and can add "anyone (with password)" to "any" team.

ignore my comment, couldn't see the check with invitation code in the deep group.join flow. it looks good to me now 👍

cihangir (approved):
awesome, thanks for the refactoring and making test better in shape!

PR #10508, opened 2017-02-06
socialapi/api: retry on channel participant add

cihangir:
there are race conditions where we try to add user before creating them in socialapi

### DIFF
--- a/go/src/socialapi/workers/api/modules/participant/participant.go
+++ b/go/src/socialapi/workers/api/modules/participant/participant.go
@@ -9,7 +9,9 @@ import (
 	"socialapi/models"
 	"socialapi/request"
 	"socialapi/workers/common/response"
+	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/koding/bongo"
 	"github.com/koding/runner"
 )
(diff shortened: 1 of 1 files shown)


szkl (approved):

go/src/socialapi/workers/api/modules/participant/participant.go
```diff
@@ -45,22 +47,51 @@ func List(u *url.URL, h http.Header, _ interface{}, context *models.Context) (in
 func AddMulti(u *url.URL, h http.Header, participants []*models.ChannelParticipant, context *models.Context) (int, http.Header, interface{}, error) {
 	query := context.OverrideQuery(request.GetQuery(u))
 
+	bo := backoff.NewExponentialBackOff()
+	bo.InitialInterval = time.Millisecond * 50
+	bo.MaxInterval = time.Millisecond * 100
+	bo.MaxElapsedTime = time.Second * 2
```
rjeczalik: After setting `bo` you'd need to call bo.Reset(), otherwise the InitialInterval [won't get set up](https://sourcegraph.com/github.com/koding/koding@80d10f322aa2d5ae903d6f65421824d253bf9263/-/blob/go/src/vendor/github.com/cenkalti/backoff/exponential.go#L112:1-113:1).

rjeczalik (approved):

go/src/socialapi/workers/api/modules/participant/participant.go
cihangir: NewTicker handles that https://github.com/cenkalti/backoff/blob/8edc80b07f38c27352fb186d971c628a6c32552b/ticker.go#L45

PR #10511, opened 2017-02-06
Fix data inconsistencies on the fly

cihangir:


szkl (approved):

go/src/socialapi/models/channel.go
```diff
@@ -176,7 +176,30 @@ func (c *Channel) Create() error {
...
+		GroupName: c.GroupName,
+		Type:      c.TypeConstant,
+		Name:      c.Name,
+	}
+	ch, er := c.ByName(q)
+	if er != nil {
+		// ignore error for fetch request
+		return err
+	}
+	c = &ch
```
rjeczalik: The c here is a copy of a pointer value, writing it has effect only within the scope (https://play.golang.org/p/7wch4Wj5Re).

You can either:

```go
*c = ch
```

or change the func sig to `func idempotentCreate(c *Channel) (*Channel, error) {`.

cihangir: fixed with https://github.com/koding/koding/pull/10524/commits/a6c7aac39eb48522441a4811e91ae2cbe41b75c8

PR #10518, opened 2017-02-06
Fix for trailing slash and set button back to 'Copy'

andrewjcasal:
## Types of changes

## Checklist:

### DIFF
--- a/docs/badges.html
+++ b/docs/badges.html
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: Badges
-permalink: /badges
+permalink: /badges/
 ---
 
 <!-- HEADER -->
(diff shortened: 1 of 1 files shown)


docs/badges.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: Badges
-permalink: /badges
+permalink: /badges/
```
sinan: since you have the file in jekyll root, you don't need this line at all I guess. pls check.

docs/badges.html
```diff
@@ -64,6 +64,10 @@
 
   clipboard.on('success', function(e) {
     e.trigger.innerHTML = "Copied!"
+
+    setTimeout(function() {
+      e.trigger.innerHTML = "Copy"
+    }, 1000);
```
sinan: you need to clear the timer in case of multiple clicks.

```js
var timer = null;
clipboard.on('success', function(e) {
  e.trigger.innerHTML = "Copied!"
  if (timer) {
    clearTimeout(timer);
  }
  timer = setTimeout(function() {
    e.trigger.innerHTML = "Copy"
  }, 1000);
});
```

sinan (changes requested):

sinan (approved):

PR #10525, opened 2017-02-07
Apply formatting to bash files

cihangir:
```
go get -u github.com/mvdan/sh/cmd/shfmt
shfmt -l -w ./scripts/
```


cihangir:
looks better with https://github.com/koding/koding/pull/10525/files?w=1

scripts/test-instance/run-test
```diff
@@ -37,3 +37,4 @@ wait $PID # Wait until child process is exited
 trap - TERM INT
 
 wait $PID # In case an interrupt signal is received
+EOF
```
szkl: This will cause a failure. Line 10 is already closing the here document. We can block comment lines between 4 and 8 instead of a here document.

szkl (changes requested):

gokmen (approved):

cihangir:
@szkl PTAL

szkl (approved):

PR #10522, opened 2017-02-07
Klient machine mount queue

ppknap:
This PR adds queue and queue event coalescing logic needed to handle FS events.

Depends on: ~#10509~

## How Has This Been Tested?
Unit tests

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #10530, opened 2017-02-08
Klient machine mount file synchronization logic.

ppknap:
This PR adds synchronization logic for file events produced by mount notifiers (like FUSE).

The logic is:

`K` - event producers (like FUSE FS call).
`M`(where `K > M`) - coalesced and queued events for a single mount (`Anteroom` and `Queue` logic in `mount/sync` package).
`N`(where `N == 2*runtime.NumCPU`) - common execution worker pool for all events from all mounts.

`K -sends-events-to-> M -> N`   

Depends on: ~#10522~

## How Has This Been Tested?
Unit tests

## Types of changes


ppknap:
@cihangir @rjeczalik some of the changes are just renaming (i moved one package) it seems that git didn't detect all moves. Thus:

`go/src/koding/klient/machine/machinegroup/supervisors/supervisors.go`
is a mv to
`go/src/koding/klient/machine/machinegroup/syncs/syncs.go`

cihangir (approved):

go/src/koding/klient/machine/machinegroup/syncs/syncs.go
```diff
@@ -0,0 +1,239 @@
+package syncs
```
rjeczalik: This file was moved.

rjeczalik (approved):

PR #10531, opened 2017-02-08
Klient machine mount file system notification logic (without FUSE) 

ppknap:
This PR adds FS notification logic for mount notifiers (like FUSE).

Depends on: ~#10530~

## How Has This Been Tested?
Unit tests

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/notify/notify.go
+++ b/go/src/koding/klient/machine/mount/notify/notify.go
@@ -1,0 +1,43 @@
+package notify
+
+import (
+	"context"
+
+	"koding/klient/machine/index"
+	"koding/klient/machine/mount"
+)
+
+// BuildOpts represents the context that can be used by external notifiers to
+// build their own type. Built notifier should only read from provided indexes
+// and, if changes occur, commit observed changes using Cache interface.
+type BuildOpts struct {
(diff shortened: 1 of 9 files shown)


go/src/koding/klient/machine/mount/notify/notify.go
```diff
@@ -0,0 +1,43 @@
...
+// in order to create their instances.
+type Builder interface {
+	// Build uses provided build options to create Notifier instance.
+	Build(opts *BuildOpts) (Notifier, error)
+}
+
+// Notifier is an interface which must be implemented by external notifiers.
+type Notifier interface {
+	// Close cleans up notifier resources, if any.
+	Close()
```
rjeczalik: Would it be possible to have `Close() error` here? If notifier, for whatever reason, needs to implement io.Closer interface, it would conflict with this one. Or rename it to `Stop()`.

ppknap: Yes, that's good point. I will make all `Close()` methods `io.Closer`able.

ppknap: changed all `Close()` methods to `Close() error` here #10536

cihangir (approved):

rjeczalik (approved):

PR #10532, opened 2017-02-08
klient/machine: add String methods to index Changes.

ppknap:
Pretty print for index changes needed for mount debugging.

## How Has This Been Tested?
Unit tests

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/index/change.go
+++ b/go/src/koding/klient/machine/index/change.go
@@ -199,6 +199,12 @@ func (c *Change) Coalesce(newer *Change) *Change {
 	}
 }
 
+// String implements fmt.Stringer interface. It pretty prints stored change.
+func (c *Change) String() string {
+	age := time.Now().UTC().Sub(time.Unix(0, atomic.LoadInt64(&c.createdAt)))
+	return c.meta.String() + " " + age.String() + " " + c.name
+}
+
 // ChangeSlice stores multiple changes.
 type ChangeSlice []*Change
 
--- a/go/src/koding/klient/machine/index/change.go
+++ b/go/src/koding/klient/machine/index/change.go
@@ -9,12 +9,12 @@ import (
 type ChangeMeta uint64
 
 const (
-	ChangeMetaLocal  ChangeMeta = 1 << iota // local->remote synchronization.
-	ChangeMetaRemote                        // remote->local synchronization.
-	ChangeMetaAdd                           // File was added.
-	ChangeMetaRemove                        // File was removed.
-	ChangeMetaUpdate                        // File was updated.
-	ChangeMetaLarge                         // File size is above 4GB.
+	ChangeMetaLocal  ChangeMeta = 1 << iota // L: local->remote synchronization.
+	ChangeMetaRemote                        // R: remote->local synchronization.
+	ChangeMetaAdd                           // a: File was added.
+	ChangeMetaRemove                        // d: File was removed.
+	ChangeMetaUpdate                        // u: File was updated.
+	ChangeMetaHuge                          // H: File size is above 4GB.
 )
 
 // Followed constants are helpers for ChangeMeta.Coalesce method.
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/index/change.go
```diff
@@ -103,6 +103,25 @@ func (cm *ChangeMeta) Coalesce(newer ChangeMeta) ChangeMeta {
 	}
 }
 
+// String implements fmt.Stringer interface and pretty prints stored change.
+func (cm *ChangeMeta) String() string {
+	cpy := atomic.LoadUint64((*uint64)(cm))
+
+	const cmstr = "LRaduH"
```
cihangir: this is not future proof, if someone adds a new lines betwee iota consts, this would break. Please either use a map/struct with their integer/char representation or if you have any better idea.

go/src/koding/klient/machine/index/change.go
```diff
@@ -180,6 +199,12 @@ func (c *Change) Coalesce(newer *Change) *Change {
 	}
 }
 
+// String implements fmt.Stringer interface. It pretty prints stored change.
+func (c *Change) String() string {
+	age := time.Now().UTC().Sub(time.Unix(0, atomic.LoadInt64(&c.createdAt)))
```
cihangir: just curious why do you need to be atomic here? also could you use `c.CreatedAtUnixNano()`

cihangir (changes requested):
logic duplication/iota usage

go/src/koding/klient/machine/index/change.go
ppknap: this is a solution used in [std's os.FileMode.String()](https://golang.org/src/os/types.go?s=2562:2595#L49)

Adding new lines between iota const will break the tests. I can make it a map(however tried to avoid this). wdyt

go/src/koding/klient/machine/index/change.go
ppknap: Because, there is [logic](https://github.com/koding/koding/blob/10b9a090e637a38cb9ba526df4862bf0f34cc910/go/src/koding/klient/machine/index/change.go#L196) which may replace this field.

It is used by anteroom [logic](https://github.com/koding/koding/blob/3d9e5d90ad60a0a80566ef4fa4a36098dc984430/go/src/koding/klient/machine/mount/sync/anteroom.go#L82)(see comment) which may invalidate change that has been already sent to execution(sync) [workers](https://github.com/koding/koding/blob/3d9e5d90ad60a0a80566ef4fa4a36098dc984430/go/src/koding/klient/machine/machinegroup/syncs/syncs.go#L126).

go/src/koding/klient/machine/index/change.go
cihangir: what made you avoid that? what problems do you see?

ppknap: using heap-allocated map just to print a small value. This is not a big overhead but this solution is as simple as map and much faster. This is the only reason and I'm not opposed to using maps(or I just read std lib too much:) )

rjeczalik (approved):

go/src/koding/klient/machine/index/change.go
ppknap: I tried to use map approach but it has some drawbacks:

First I need to create `map[ChangeMeta]byte` but to preserve the order of iteration I will need to create another variable: slice `[]ChangeMeta` which I will be iterating over. This doesn't look very nice to me:

```go
var cm2byte := map[ChangeMeta]byte{
	ChangeMetaLocal: byte('L'),
	ChangeMetaRemote: byte('R'),
	ChangeMetaAdd: byte('a'),
	ChangeMetaRemove: byte('r'),
	ChangeMetaUpdate: byte('u'),
	ChangeMetaHuge: byte('H'),
}

var cmOrder = []ChangeMeta{
	ChangeMetaLocal,
	ChangeMetaRemote,
	ChangeMetaAdd,
	ChangeMetaRemove,
	ChangeMetaUpdate,
	ChangeMetaHuge,
}

// ...
for _, cm := range cmOrder {
   if cmOriginal&cm != 0 {
      // add cm2byte to return string, else "-" 
   }
}
```

Does this approach look better to you? Or am I missing something?

cihangir: yea this looks a bit long. I was thinking the other way around while creating the map

```
var cmMapping = map[byte]ChangeMeta{
	'L': ChangeMetaLocal,
	'R': ChangeMetaRemote,
	'a': ChangeMetaAdd,
	'd': ChangeMetaRemove,
	'u': ChangeMetaUpdate,
	'H': ChangeMetaLarge,
}

// ...
for c, i := range cmMapping {
	w := getPowerof2(uint64(i))
	if cpy&uint64(i) != 0 {
		buf[w] = c
	} else {
		buf[w] = '-'
	}
}

return string(buf[:len(cmMapping)])
// ...
```

full working example https://play.golang.org/p/cqLp2F8j-O


ps: we can omit `uint64(i)` conversion by putting it into construction of the "cmMapping"

ppknap: Thanks @cihangir that's clever:) I will use this approach

cihangir (approved):
👍

PR #10477, opened 2017-02-02
Gometalinter script

mehmetalisavas:
Gometalinter does not check all lints
Enabled linters are;
- deadcode
- ineffassign
- structcheck
- gosimple
- misspell
- unconvert

PR #10547, opened 2017-02-10
Gometalinter config

mehmetalisavas:
update of gometalinter.json file;
metalinter flags are move under the config file


mehmetalisavas:
once before merged, https://github.com/koding/koding/pull/10477 needs to be merged.

go/src/.gometalinter.json
```diff
@@ -7,5 +7,9 @@
     "gosimple",
     "misspell",
     "unconvert"
-  ]
+  ],
```
szkl: Let's keep `Enable` property at the bottom.

go/src/.gometalinter.json
```diff
@@ -7,5 +7,9 @@
...
+  "Concurrency": 16,
```
szkl: This is the default value.

szkl (changes requested):

go/src/.gometalinter.json
mehmetalisavas: yeah ur right, actually i have written as comment for the specification like this because commenting is not allowed by json.

mehmetalisavas:
@cihangir PTAL

mehmetalisavas:
cc/ @szkl

cihangir (approved):

szkl (approved):

PR #10548, opened 2017-02-11
New Stack Editor: Save Flow

gokmen:
### Save Flow

ref: http://recordit.co/4S0eusMbGE

### Credential Warnings on Toolbar Banner
http://recordit.co/AzDNq50kEZ


client/app/lib/util/stacks/updatecustomvariable.coffee
```diff
@@ -59,8 +59,7 @@ module.exports = updateCustomVariable = (options, callback) ->
         createAndUpdate { provider, title, meta, stackTemplate }, callback
       else
         credential.update { meta, title }, (err) ->
-          shareWithGroup credential, ->
-            callback err, stackTemplate
+          shareWithGroup credential, callback
```
usirin: Do we really want to share it with group, if there is an error?

client/new-stack-editor/lib/controllers/logs.coffee
```diff
@@ -21,7 +23,13 @@ module.exports = class LogsController extends BaseController
 
     content = stringify content
     content = "[#{dateFormat Date.now(), 'HH:MM:ss'}] #{content}"
+
     @editor.addContent content
+    debug 'log', content
+
+    if content[content.length - 3..] is '...'
```
usirin: A comment on what this is doing would be really helpful here.

client/app/lib/util/stacks/updatecustomvariable.coffee
gokmen: right, missed that one. updating.

sinan (approved):
👍

client/app/lib/remote-extensions/stacktemplate.coffee
```diff
@@ -3,9 +3,19 @@ remote = require('../remote')
 module.exports = class JStackTemplate extends remote.api.JStackTemplate
 
 
-  getCredentialIdentifiers: ->
+  getCredentialIdentifiers: (exclude = ['custom']) ->
 
     ids = []
-    ids = ids.concat id  for provider, id of (@credentials ? {})
+    for provider, id of (@credentials ? {}) when provider not in exclude
+      ids = ids.concat id
```
gokmen: one of them getting the `value` the other one is getting the `key` excluded

PR #10558, opened 2017-02-13
Fix data inconsistencies

cihangir:


workers/social/lib/social/models/group/validators.coffee
```diff
@@ -3,7 +3,7 @@
 getRoles = (permission, permissionSet) ->
   return []  unless permissionSet
   roles = ( perm.role for perm in permissionSet.permissions \
-                      when permission in perm.permissions   )
+                      when permission in perm?.permissions   )
```
usirin:
```coffee
when in (perm?.permissions or [])
```

might be safer, just in case. (not sure if it throws or not when `foo in undefined` happens)

gokmen:
@cihangir should we update this one or is it deprecated for now?

cihangir:
i continue working on this @gokmen

cihangir:
@rjeczalik @gokmen this is finally ready at your disposal

go/src/socialapi/models/account.go
```diff
@@ -70,26 +70,23 @@ func (a *Account) FetchOrCreate() error {
 	}
 
 	err := a.One(bongo.NewQS(selector))
-	// if we dont get any error
-	// it means we found the record in our db
-	if err == nil {
-		return nil
+	// first check if the err is not found err
+	if err != bongo.RecordNotFound {
```
rjeczalik:
```
if err != nil && err != bongo.RecordNotFound {
```

rjeczalik: Oh I got it on a second read - it should be nop if err is nil. Ignore me.

rjeczalik (approved):

gokmen (approved):

PR #10565, opened 2017-02-15
Channel group cleaner

mehmetalisavas:


go/src/socialapi/models/channel.go
```diff
@@ -1162,3 +1165,64 @@ func (c *Channel) ShowUnreadCount() bool {
...
+				continue
+			}
+			// if error is not nil and equal to record not found
+			// then remove the channel and its participants in postgre
+			if err != nil && err == mgo.ErrNotFound {
+				if err = channel.DeleteWithParticipantsForce(); err != nil {
+					errs = multierror.Append(errs, err)
+				}
+			} else {
+				errs = multierror.Append(errs, err)
```
cihangir: you are going to append even if error is nil

go/src/socialapi/models/channel.go
```diff
@@ -1162,3 +1165,64 @@ func (c *Channel) ShowUnreadCount() bool {
...
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+
+		for _, channel := range channels {
+			_, err := modelhelper.GetGroup(channel.GroupName)
+			// if group already exists in mongo, then we don't need to fecth the same data
+			// while fetching channels
+			if err == nil {
+				offset++
```
cihangir: offset needs to be increased even if there is an error

go/src/socialapi/models/channel.go
```diff
@@ -1162,3 +1165,64 @@ func (c *Channel) ShowUnreadCount() bool {
...
+				}
+			} else {
+				errs = multierror.Append(errs, err)
+			}
+		}
+
+		// This check provide us to break the loop if there is no data left that need
+		// to be processed fetch tolerance is limit/2, if fetched channels count is
+		// less than limited number then break the loop.
+		if len(channels) < (limit / 2) {
```
cihangir: i think "limit / 2" is unnecessary, if it is smaller than len(channels) we can assume safe to break out of the loop

go/src/socialapi/models/channel.go
mehmetalisavas: actually, we'r appending only the error except `mgo.ErrNotFound` errors. there is early return on above line while checking `if err==nil{continue}`

go/src/socialapi/models/channel.go
mehmetalisavas: i thought it would be nice to try to fetch Group again. is it guarenteed to get the same error ?

cihangir: most probably yes

go/src/socialapi/models/channel.go
cihangir: 🤦‍♂️ 👍

mehmetalisavas:
@cihangir ping

go/src/socialapi/scripts/channelcleaner/channelcleaner_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// In this case we need to try to create same named group in mongo.
+	// then it gives key duplication error, so err checking is disabled here.
+	models.CreateGroupInMongo(groupName, groupChannel.Id)
+	// So(err, ShouldBeNil)
+
+	return groupChannel, account
+}
+
+//
+// func accountCreatorWithCount(count int) error {
```
cihangir: why this commented out code still here?

cihangir (changes requested):
commented out code.

cihangir:
please rebase you branch

cihangir (approved):

PR #10579, opened 2017-02-16
Transfer Ownership and Delete Account feature

hakankaradis:
- [x] Add Delete Account button and modal
- [x] Add Transfer Ownership button with front-end and back-end functionalities
- [x] Notify New owner and team for role change for new owner and redirect user to the team to delete
- [x] Add test for transferownership func
- [x] Add destroy account func
- [x] Add test for destroy account
- [x] Add checkUserPassword util with test
- [x] Add delete account and team verify modal
- [x] Papertrail hook created for this query `[ACCOUNT_DESTROY_FAILED]`

#### Screencast
`Transfer Ownership: ` http://recordit.co/BgUKDppBBz
`Delete Account:`
http://recordit.co/8TyGiG7pKA
 http://recordit.co/FBjwTfCy76


hakankaradis:
@cihangir @gokmen I will update error messages, copy in modals and styling beside that it is ready for review

workers/social/lib/social/models/account.coffee
```diff
@@ -1239,3 +1244,94 @@ module.exports = class JAccount extends jraphical.Module
...
+          group.destroy client, (err) ->
+            kallback 'GroupDestroy', next, err
+
+        (next) ->
+          JUser = require './user'
+          JUser.unregister client, username, (err) ->
+            kallback 'JUser', next, err
+
+        (next) ->
+          JSession = require './session'
```
cihangir: i think it is better to remove JSession after deleting everything

workers/social/lib/social/models/group/index.coffee
```diff
@@ -1466,32 +1467,62 @@ module.exports = class JGroup extends Module
...
+        JAccount.one { _id: accountId }, (err, account) =>
 
-        JAccount.one { _id:accountId }, (err, account) =>
           return callback err if err
 
           @fetchRolesByAccount account, (err, newOwnersRoles) =>
+
             return callback err if err
 
+            if 'blockedAccount' in newOwnersRoles
```
cihangir: 💯

workers/social/lib/social/models/group/index.coffee
```diff
@@ -1466,32 +1467,62 @@ module.exports = class JGroup extends Module
...
+
+    { clientAccountId, accountId, slug } = options
+
+    return callback new KodingError 'Account is not provided'  unless accountId
+
+    @fetchOwner (err, owner) =>
+
+      return callback err if err
+
+      if clientAccountId and JSON.stringify(clientAccountId) isnt JSON.stringify(owner.getId())
```
cihangir: i am not sure if this check is required, we already check for permission in the $ function

workers/social/lib/social/models/group/index.coffee
```diff
@@ -1501,8 +1532,22 @@ module.exports = class JGroup extends Module
...
+                role: 'owner'
+                id: accountId
+                group: slug
+                adminNick: owner.profile.nickname
+
+              JGroup.one { slug }, (err, group) ->
+                if group
+                  group.sendNotification 'MembershipRoleChanged', contents
+
+                ownershipRelation.update { $set: { targetId: account.getId() } }, kallback
```
cihangir: it is better to remove the first relationship & add new one with the relevant data, 
1) we might cache the relationships in the future and this might cause problems
2) created at field in the _id property would have the wrong timestamp
3) i dont remember any other place where we update the relationship

workers/social/lib/social/models/group/index.coffee
```diff
@@ -1501,8 +1532,22 @@ module.exports = class JGroup extends Module
...
+              # notify group for new owner
+              contents =
+                role: 'owner'
+                id: accountId
+                group: slug
+                adminNick: owner.profile.nickname
+
+              JGroup.one { slug }, (err, group) ->
+                if group
+                  group.sendNotification 'MembershipRoleChanged', contents
```
cihangir: lets send the notification after having the new role inplace

workers/social/lib/social/models/group/index.test.coffee
```diff
@@ -159,6 +159,94 @@ runTests = -> describe 'workers.social.group.index', ->
...
+          { account: member2 } = data
+          done()
+
+      it 'should not be able to transferOwnership to kicked member', (done) ->
+
+        accountId = member2._id
+
+        queue = [
+          (next) ->
+            group.kickMember adminClient, accountId, (err) ->
```
cihangir: 👍

workers/social/lib/social/models/group/index.test.coffee
```diff
@@ -159,6 +159,94 @@ runTests = -> describe 'workers.social.group.index', ->
...
+
+        async.series queue, done
+
+      it 'should throw error for not provided account', (done) ->
+
+        group.transferOwnership {}, (err) ->
+          expect(err.message).to.be.equal 'Account is not provided'
+          done()
+
+      it 'should be able to transfer the ownership to the member', (done) ->
```
cihangir: 👍

workers/social/lib/social/models/utils.test.coffee
```diff
@@ -1,44 +1,89 @@
...
+        expect(err).to.exist
+        expect(err.type).to.be.equal 'NOT_REACHABLE'
+        done()
+
+    it 'should support query strings on addresses', (done) ->
+      isAddressValid 'http://yahoo.com?test=foo&bar', (err) ->
+        expect(err).to.not.exist
+        done()
+
+  describe 'utils::checkUserPassword', ->
```
cihangir: 💯

cihangir:
@hakankaradis a little bird told me that you want to switch to backend ;)

cihangir (changes requested):
some small things, overall 💯

workers/social/lib/social/models/group/index.coffee
gokmen: and side note, if you need to compare those Id's you can use `owner.getId().equals clientAccountId` which is the correct way for comparison here.

gokmen: and side note, if you need to compare those Id's you can use `owner.getId().equals clientAccountId` which is the correct way for comparison here.

client/app/lib/util/fetchMyRelativeGroups.coffee
```diff
@@ -0,0 +1,27 @@
...
+
+  whoami().fetchRelativeGroups (err, groups) ->
+
+    return callback err  if err
+
+    { groupsController } = kd.singletons
+    currentGroup = _.find groups, (group) ->
+      group.slug is groupsController.getCurrentGroup().slug
+
+    rejectedSlugs = [ 'koding', currentGroup.slug ]
```
gokmen: let's handle this case in client version of this call in backend, I mean lets `fetchRelativeGroups$` returns the groups excluded `koding` and current `group` but `fetchRelativeGroups` method will keep return groups as is.

client/app/lib/util/fetchMyRelativeGroups.coffee
```diff
@@ -0,0 +1,27 @@
...
+    groups = _.reject groups, (group) ->
+      group.slug in rejectedSlugs
+
+    groups = _.sortBy groups, 'slug'
```
gokmen: you can also include these parts to that method as well.

client/app/lib/util/verifyPassword.coffee
```diff
@@ -0,0 +1,13 @@
+whoami = require 'app/util/whoami'
+remote = require 'app/remote'
+
+module.exports = (password, callback) ->
+
+  whoami().fetchEmail (err, email) ->
+    options = { password, email }
+    remote.api.JUser.verifyPassword options, (err, confirmed) ->
```
gokmen: we shouldn't require two explicit calls in here I believe, let's update or provide another method for verification password for the active session owner.

client/home/lib/account/deleteaccount/deleteaccountmodal.coffee
```diff
@@ -0,0 +1,60 @@
...
+    @addSubView footer = new kd.CustomHTMLView
+      cssClass: 'modal-footer'
+
+    footer.addSubView deleteAccount = new kd.ButtonView
+      title          : 'Delete Team And Account'
+      cssClass       : 'GenericButton delete-account-button'
+      callback       : ->
+        fetchMyRelativeGroups (err, groups) ->
+
+          groups = groups.filter (group) -> 'owner' in group.roles
```
gokmen: we can also pass this query to the backend so we won't need to fetch all the groups that I'm member of which is not required for this case. number of groups that you are member of will always be more number of groups that you own, all of the time. so better to not fetch all that data over the wire.

client/home/lib/myteam/components/transferownershipbutton/index.coffee
```diff
@@ -0,0 +1,88 @@
...
+            cancelBtn.hide()
+            ownershipBtn.destroy()
+            new kd.NotificationView
+              title: "You don't have any other team member, please delete team"
+              duration: 3000
+
+
+  verifyModal: (group, accountId) ->
+
+    username = $("[value='#{accountId}']").text()
```
gokmen: have no idea why do we need to this in this way with jquery.

workers/social/lib/social/models/account.coffee
```diff
@@ -140,6 +143,8 @@ module.exports = class JAccount extends jraphical.Module
           (signature Object, Function)
         pushNotification:
           (signature Object, Function)
+        destroyAccount:
```
gokmen: what about `destroy` only?

workers/social/lib/social/models/account.coffee
```diff
@@ -1239,3 +1244,94 @@ module.exports = class JAccount extends jraphical.Module
...
+    @fetchRelativeGroups (err, groups) =>
+
+      # remove koding team
+      groups = groups.filter (group) -> group.slug isnt 'koding'
+
+      ownerOfGroups = groups.filter (group) -> 'owner' in group.roles
+
+      if ownerOfGroups.length > 1
+        return callback new KodingError 'You cannot delete your account when you have ownership in other team', ownerOfGroups
+
```
gokmen: again in here if you provide `query` support in `fetchRelativeGroups` you can use it in here too.

workers/social/lib/social/models/account.coffee
```diff
@@ -1239,3 +1244,94 @@ module.exports = class JAccount extends jraphical.Module
...
+        errors.push { label, err }  if err
+        next()
+
+      group = groups[0]
+
+      errors = []
+
+      # delete the team and delete the account
+      username = @profile.nickname
```
gokmen: `username = @getAt 'profile.nickname'`

workers/social/lib/social/models/account.coffee
```diff
@@ -1239,3 +1244,94 @@ module.exports = class JAccount extends jraphical.Module
...
+
+      if ownerOfGroups.length > 1
+        return callback new KodingError 'You cannot delete your account when you have ownership in other team', ownerOfGroups
+
+      kallback = (label, next, err) ->
+
+        errors.push { label, err }  if err
+        next()
+
+      group = groups[0]
```
gokmen: `[ group ] = groups`

gokmen (changes requested):
looks cool just some small changes required 👍

client/home/lib/account/deleteaccount/singlegroupinfo.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor: (options, data) ->
+
+    options = _.assign {}, options,
+      cssClass : 'single-group-info'
+
+    super options, data
+
+    group = data
+    roles = data.roles
```
gokmen: `data -> @getData()`

client/home/lib/account/deleteaccount/singlegroupinfo.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+    teamlogowrapper.addSubView teamname = new kd.CustomHTMLView
+      cssClass: 'team-name'
+      partial: capitalizeFirstLetter group.slug
+
+    @addSubView buttonwrapper = new kd.CustomHTMLView
+      cssClass: 'action-wrapper'
+
+    buttonwrapper.addSubView ownershipBtn = new TransferOwnershipButton {}, group
+    ownershipBtn.once 'destroy', =>
```
gokmen: `KDObjectWillBeDestroyed`

client/home/lib/myteam/components/transferownershipbutton/index.coffee
```diff
@@ -0,0 +1,88 @@
...
+        <strong>CAUTION! </strong>You are going to transfer the ownership
+        of your team with <strong>#{username}</strong>
+      </p> <br>
+      <p>Please enter <strong>current password</strong> into the field below to continue: </p>"
+
+    new VerifyPasswordModal 'Confirm', partial, (currentPassword) =>
+      verifyPassword currentPassword, (err) =>
+        return  if showError err
+
+        group.transferOwnership { accountId, currentPassword, slug: group.slug }, (res) =>
```
gokmen: error handling?

client/home/lib/myteam/components/transferownershipbutton/index.coffee
```diff
@@ -0,0 +1,88 @@
...
+verifyPassword = require 'app/util/verifyPassword'
+require('./styl/transferownershipbutton.styl')
+
+
+module.exports = class TransferOwnershipButton extends kd.CustomHTMLView
+
+  constructor: (options, data) ->
+
+    options = _.assign {}, options,
+      cssClass : 'transferownershipbutton'
```
gokmen: curry might be better here.

workers/social/lib/social/models/account.coffee
```diff
@@ -1239,3 +1257,92 @@ module.exports = class JAccount extends jraphical.Module
...
+
+      # delete the team and delete the account
+      username = @getAt 'profile.nickname'
+
+      queue = [
+
+        (next) ->
+          group.destroy client, (err) ->
+            kallback 'GroupDestroy', next, err
+
```
gokmen: we need to destroy other resources on other groups owned by this user, at this point.

gokmen (approved):
💯

cihangir (approved):

PR #10581, opened 2017-02-16
machine/index: add EntryPromiseUnlink meta

rjeczalik:
This PR:

- adds EntryPromiseUnlink meta
- adds Entry.Ref field
- adds utility helpers for atomic read of Entry's fields

The fuse.Unlink method decrements Entry.Ref, however when it reaches 0 the file cannot be deleted right away, as it may still be referenced by process owning a file handle. Once ForgetInode is called, if inode is marked with EntryPromiseUnlink, the file is effectively deleted from cache.

### DIFF
--- a/go/src/koding/klient/machine/index/index.go
+++ b/go/src/koding/klient/machine/index/index.go
@@ -51,13 +51,15 @@ func (e *Entry) Has(meta int32) bool   { return atomic.LoadInt32(&e.Meta)&meta =
 
 // SwapMeta flips the value of a Meta field, setting the set
 // bits and unsetting the unset ones.
-//
-// As it uses multiple atomic ops, it is safe to call only
-// when protected by (*Index).mu.
 func (e *Entry) SwapMeta(set, unset int32) int32 {
-	meta := (atomic.LoadInt32(&e.Meta) | set) &^ unset
-	atomic.StoreInt32(&e.Meta, meta)
-	return meta
+	for {
+		older := atomic.LoadInt32(&e.Meta)
+		updated := (older | set) &^ unset
+
+		if atomic.CompareAndSwapInt32(&e.Meta, older, updated) {
+			return updated
+		}
+	}
 }
 
 // NewEntryFile creates new Entry from a file stored under path argument.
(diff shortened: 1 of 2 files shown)


go/src/koding/klient/machine/index/index.go
```diff
@@ -13,20 +13,20 @@ import (
...
+	"sync/atomic"
 
 	"github.com/djherbis/times"
 )
 
 // Version stores current version of index.
 const Version = 1
 
-// EntryMeta describes entry metadata.
-type EntryMeta int32
```
ppknap: Why deleting the type? You can still do atomic things on meta with `(*int32)(&e.Meta)`

go/src/koding/klient/machine/index/index.go
```diff
@@ -35,9 +35,29 @@ type Entry struct {
...
+func (e *Entry) IncRef() int32         { return atomic.AddInt32(&e.Ref, 1) }
+func (e *Entry) DecRef() int32         { return atomic.AddInt32(&e.Ref, -1) }
+func (e *Entry) Has(meta int32) bool   { return atomic.LoadInt32(&e.Meta)&meta == meta }
+
+// SwapMeta flips the value of a Meta field, setting the set
+// bits and unsetting the unset ones.
+//
+// As it uses multiple atomic ops, it is safe to call only
+// when protected by (*Index).mu.
+func (e *Entry) SwapMeta(set, unset int32) int32 {
```
ppknap: > As it uses multiple atomic ops, it is safe to call only when protected by (*Index).mu.

This will make this function lock-free and thread safe.
```go
for {
    older := atomic.LoadInt32(&e.Meta)

    updated := (older | set) &^ unset
    if atomic.CompareAndSwapInt32(&e.Meta, older, updated) {
        return updated
    }
}
```

ppknap (approved):

go/src/koding/klient/machine/index/index.go
rjeczalik: I found no use for a separate type, except it requires a lot of casting. Thus I simplified the api.

go/src/koding/klient/machine/index/index.go
rjeczalik: Nice one, done.

cihangir (approved):

PR #10611, opened 2017-02-20
Replace git-rev with git-rev-sync

szkl:

### DIFF
--- a/configure
+++ b/configure
@@ -4,7 +4,6 @@ log = console.log
 fs = require 'fs'
 os = require 'os'
 path = require 'path'
-git = require 'git-rev'
 request = require 'request'
 { extend, defaults, clone } = require 'underscore'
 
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -150,22 +149,22 @@ updateVault = ->
...
-  else
-    version = try (fs.readFileSync path.join(__dirname, './VERSION'), 'utf-8').trim()
-
-  if argv.hostname
-    options.hostname = argv.hostname
-    updateVault()
-  else if os.type() is "Darwin"
-    options.hostname = "dev.koding.com"
+try
+  version = require('git-rev-sync').short()
```
cihangir: `version = str.substring 0, 8` is required - there are other places where we apply string expansions

cihangir: or we need to fix other places as well :)

package.json
```diff
@@ -121,6 +120,7 @@
     "chai": "3.5.0",
     "coffeelint": "1.16.0",
     "crypto": "0.0.3",
+    "git-rev-sync": "1.8.0",
```
cihangir: 👍

configure
```diff
@@ -150,22 +149,22 @@ updateVault = ->
...
+
+if argv.hostname
+  options.hostname = argv.hostname
+  updateVault()
+else if os.type() is "Darwin"
+  options.hostname = "dev.koding.com"
+  updateVault()
+else
+  log "lemme find your IP"
+  request "https://p.koding.com/-/ip", (err, res, body) ->
```
cihangir: also there is https://www.npmjs.com/package/sync-request :)

cihangir (changes requested):
substring

configure
szkl: I've found no other reference to git-rev module.

cihangir (approved):

PR #10588, opened 2017-02-17
machine/index: improve promise ops handling

rjeczalik:

PR #10619, opened 2017-02-21
Machine index improvements

ppknap:
This PR adds support for Entry promises in index - this means that two indexes for synchronization are no longer needed. Additionally it:

 - Drops Entry checksum(CRC32 Hash) - it was not used at this stage of mount implementation.
 - Adds EntryPromiseVirtual - which means that the file can be shown by FUSE but it's not physically available.
 - Since now, Index will store root Entry during initialization. Its name/path will be empty string `""` and not `"."` since we do not use dots in other entries.
 - `RemoteIdx` and `LocalIdx` were dropped, since now there is only one `Index`.

Depends on: ~#10588~

## Motivation and Context
Simplify sync logic.

## How Has This Been Tested?
Unit tests.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/index/index.go
+++ b/go/src/koding/klient/machine/index/index.go
@@ -6,8 +6,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"hash/crc32"
-	"io"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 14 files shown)


go/src/koding/klient/machine/index/node.go
```diff
@@ -240,24 +254,31 @@ func (nd *Node) DiskSize(maxsize int64) (size int64) {
 //
 // It ignored nodes marked as deleted.
 func (nd *Node) ForEach(fn func(string, *Entry)) {
+	nd.forEach(fn, false)
+}
+
+// ForEachAll traverses the tree and calls fn on every node's entry.
+func (nd *Node) ForEachAll(fn func(string, *Entry)) {
```
rjeczalik: Is it needed?

go/src/koding/klient/machine/index/node_test.go
```diff
@@ -211,19 +145,19 @@ func TestNodeLookup(t *testing.T) {
...
+				t.Fatalf("got %d, want %d", size, nd.Entry.Size())
 			}
 		})
 	}
 }
 
 func TestNodeCount(t *testing.T) {
 	cases := map[int64]int{
-		-1:   32,
+		-1:   33,
```
rjeczalik: Why the difference?

go/src/koding/klient/machine/index/node.go
```diff
@@ -278,6 +299,11 @@ func (nd *Node) Lookup(path string) (*Node, bool) {
 	return nd.lookup(path, false)
 }
 
+// LookupAll looks up a node given by the path.
+func (nd *Node) LookupAll(path string) (*Node, bool) {
+	return nd.lookup(path, true)
```
rjeczalik: Is it needed?

go/src/koding/klient/machine/index/node.go
ppknap: yes, sometimes I need to include deleted/virtual nodes and sometimes not. Since there is only one index we need to know which files are available locally(have no promises or promise add).

Also `index.Merge` needs to traverse all nodes even these marked as deleted.

go/src/koding/klient/machine/index/node_test.go
ppknap: > Since now, Index will store root Entry during initialization. Its name/path will be empty string "" and not "." since we do not use dots in other entries.

go/src/koding/klient/machine/index/node.go
ppknap: see my comment above

rjeczalik (approved):

go/src/koding/klient/machine/index/index.go
```diff
@@ -267,63 +182,108 @@ func (idx *Index) Lookup(name string) (*Node, bool) {
...
 	idx.mu.RUnlock()
 
+	rootBranch := filepath.Join(root, branch)
+	visited := map[string]struct{}{
+		rootBranch: struct{}{}, // Skip root.
+	}
+
 	if !ok {
-		rt = newNode()
+		goto skipBranch
```
cihangir: a believe we can use function call here

cihangir (approved):
apart from "goto" statement LGTM

PR #10629, opened 2017-02-22
Add existence check for payment props of teams

usirin:


client/app/lib/util/getGroupStatus.coffee
```diff
@@ -12,6 +12,9 @@ module.exports = getGroupStatus = (group) ->
     # This state is only possible for members.
     when not group.payment then Status.NEEDS_UPGRADE
 
+    # This state should never be here.
+    when not group.payment.customer then Status.UNKNOWN
```
cihangir: this is not unknown every non-complete data should be treated as NEEDS_UPGRADE - also this is valid state since we create the subscription on the fly for the teams that are created before our pricing release

usirin:
@cihangir updated

cihangir (approved):
Awesome, thanks @usirin 💯

PR #10634, opened 2017-02-23
Intercom Minor Fix

hakankaradis:
Remove already attached js (intercomLauncher.js) from html files
Remove redundant variables and click event listener from trail.js


sinan (approved):
👍

PR #10635, opened 2017-02-23
Use new Notification View in showError util

hakankaradis:
- [x] update react-flip-move package
- [x] fix styling for notification list
- [x] update showError util to render new notification view

### Default Parameters in ShowError util
- type is `default`
- content is `Something went wrong!`
- dismissible is `yes`
- duration is `3000`
- primaryButtonTitle is `null`
- onPrimaryButtonClick is `null`
- secondaryButtonTitle is `null`
- onSecondaryButtonClick is `null`

### Screencast
http://recordit.co/yxkeWQGqru

### DIFF
--- a/client/app/lib/providers/computecontroller.coffee
+++ b/client/app/lib/providers/computecontroller.coffee
@@ -1148,7 +1148,7 @@ module.exports = class ComputeController extends KDController
         .then (shared) ->
           new kd.NotificationView { title: 'Permissions fixed' }
         .catch (err) ->
-          showError err, 'Failed to fix permissions'
+          showError err
 
 
   checkMachinePermissions: ->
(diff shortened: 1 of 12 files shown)


client/app/lib/providers/computecontroller.coffee
```diff
@@ -1148,7 +1148,7 @@ module.exports = class ComputeController extends KDController
         .then (shared) ->
           new kd.NotificationView { title: 'Permissions fixed' }
         .catch (err) ->
-          showError err, 'Failed to fix permissions'
+          showError err
```
gokmen: maybe we can just pass `showError` to `catch` handler instead like;

```
.catch showError
```

gokmen (changes requested):
👍

gokmen (approved):

Issue #10681, opened 2017-03-02
kd: cp does not work for large files


rjeczalik:
Trying to copy ~400MiB file using `kd cp`:

```
$ kd cp apple:/home/rjeczalik/koding-client.zip /Volumes/rjk.io/
Copying file or directory...
 100 / 100 [===========================================================================] 100.00% 12s
Copy complete.
```

After the command finishes, the file does not exist.

Bonus points if the progress bar actually displays progress - for small amount of files that vary in size it's pretty unusable - the progress granularity is a single file, not the amount of bytes copied.

PR #10637, opened 2017-02-23
Rsync as machine mount syncer

ppknap:
This PR adds rsync(1) based synchronization medium to mounts.

Depends on: ~#10695~ ~#10693~ 

## Motivation and Context
Make mounts work.

## How Has This Been Tested?
Unit tests.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #10687, opened 2017-03-03
klient/machine: add rsync executable support - progress parser

ppknap:
This PR adds rsync parser/executor as separate package. It covers only progress parsing logic.

- needed for rsync-based prefetch.
- rsync is used in `kd cp`
- will remove rsync logic from: #10637 

## Motivation and Context
`remote/rsync` package lacks support for `index.Change`(s) and it doesn't work too well #10681 

## How Has This Been Tested?
Unit tests

## Types of changes


go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Run starts new rsync process. And waits for it to complete.
+func (c *Command) Run(ctx context.Context) error {
+	if err := c.valid(); err != nil {
+		return err
+	}
+
+	if c.Cmd == nil {
+		c.Cmd = exec.CommandContext(ctx, "rsync")
+		c.Cmd.Env = os.Environ()
```
rjeczalik: Not needed, moreover Environ() call makes deep copy of env slice.

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Add default arguments.
+	c.Cmd.Args = append(c.Cmd.Args, "-zlptgoDd")
```
rjeczalik: Do we add a healthcheck whether all the options are available? Or do we embedded rsync with klient (`/opt/kite/klient/embedded/bin/rsync`)?

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Use remote shell if SSH private key path is set.
+	if c.PrivateKeyPath != "" {
+		// TODO(ppknap): check if RC4 cipher will work on every machine without
+		// altering sshd_config on destination.
+		rsh := []string{
+			"ssh", "-T", "-x", "-i", c.PrivateKeyPath,
```
rjeczalik: Could we use `crypto/ssh` in future instead? What about dropping a `TODO`?

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+	var (
+		now           time.Time
+		n, size, part int64
+	)
+
+	scanner := bufio.NewScanner(r)
+	scanner.Split(scanNonControl)
+	for scanner.Scan() {
+		if now.IsZero() {
+			now = time.Now()
```
rjeczalik: What about moving this to L145 and removing the if branch?

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+			part = 0
+			continue
+		}
+
+		ms := sizeRe.FindStringSubmatch(line)
+		if len(ms) < 2 {
+			continue
+		}
+
+		p, err := strconv.Atoi(strings.Replace(ms[1], ",", "", -1))
```
rjeczalik: Wouldn't `strings.NewReplacer(",", "")` be faster here? It'd use `byteReplacer` instead of rune decoding. I know, micro-optimalizations 😱 - just curious.

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			continue
+		}
+
+		part = int64(p)
+
+		speed := int64(float64(part+size) / float64(time.Since(now)/time.Second))
```
rjeczalik: Wouldn't you want to take ceiling here, to have `speed := 3` when `speed := int64(2.9)`?

```
speed := int64(float64(part+size) / float64(time.Since(now)/time.Second) + 0.5)
```

go/src/koding/klient/machine/transport/rsync/rsync.go
rjeczalik: Also not sure if and how `rsync` output progress can change its format between different versions.

rjeczalik (approved):
Minor comments.

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: It would, defined as a global var. Just thought that replacing 1-4 commas in very short string don't need special optimizations:) (yes, micro-optimizations)

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: This is intended to set/start the time once any of the exec output is available.

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: If you manage to inject `crypto/ssh` to rsync executable, then it will be possible:). What is wrong with `TODO`? It is about injecting `-c` `arcfour` parameters.

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: I had such check in #10637. I must have removed it because of tests, but going to add some later(note that this is not the entire impl)

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: Thought that losing ~0.5B/sec will not harm anyone. It turns out that I was wrong :smile:

rjeczalik: I'm not saying it will, all I'm saying the rounding computation is wrong.

go/src/koding/klient/machine/transport/rsync/rsync.go
rjeczalik: First line of exec output, ok now I see it.

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: >Thought that losing ~0.5B/sec will not harm anyone.

I meant here that rounding computation was not intended... Anyway, I will change it.

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -165,15 +164,15 @@ func (c *Command) scan(r io.Reader) {
 		if len(ms) < 2 {
 			continue
 		}
-
-		p, err := strconv.Atoi(strings.Replace(ms[1], ",", "", -1))
+		rmComma.Replace(ms[1])
```
rjeczalik: rm

cihangir (approved):

go/src/koding/klient/machine/transport/rsync/rsync_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"koding/klient/machine/transport/rsync"
+)
+
+var update = flag.Bool("update", false, "update golden files")
```
cihangir: 👍

PR #10695, opened 2017-03-06
klient/machine: create cached ssh address builder

ppknap:
This PR splits `kd machine ssh` logic into reusable methods. These methods and additional cache are needed for proper work of rsync syncer.

## How Has This Been Tested?
Existing unit tests.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/sync/sync.go
+++ b/go/src/koding/klient/machine/mount/sync/sync.go
@@ -22,8 +22,7 @@ import (
 // IndexFileName is a file name of managed directory index.
 const IndexFileName = "index"
 
-// DynamicSSHFunc is an adapter that allows to retrieve information needed to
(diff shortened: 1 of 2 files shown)


go/src/koding/klient/machine/mount/sync/sync.go
```diff
@@ -22,6 +22,10 @@ import (
 // IndexFileName is a file name of managed directory index.
 const IndexFileName = "index"
 
+// DynamicSSHFunc is an adapter that allows to retrieve information needed to
```
rjeczalik: What about:

```
DynamicSSHFunc locates the remote host which ssh should connect to.
```

Host and port are not the all information needed to make an SSH connection, so the desc may be misleading.

rjeczalik (approved):

cihangir (approved):

PR #10704, opened 2017-03-06
Use revision as fallback property for instance name tag

szkl:


gokmen:
how is this change affecting code coverage 🤔

gokmen (approved):

PR #9057, opened 2016-09-19
Enable OAuth integration per team/per provider

gokmen:
Each team can allow members to use oauth integration from different providers. This is the main requirement for GitLab integration on multiple teams on one on-premise install of Koding. But with the changes included in this PR it will allow us to keep oauth data per team/per user/per provider.
## Description

Makes oauth integration multiple across teams, providers, users. Updates GitLab oauth integration to use this new oauth mechanism and deprecates existing old way implementations for other oauth providers.
## Motivation and Context

Existing implementation was designed to use oauth integration per user/per provider but with multiple teams on different subdomains we needed to create different applications on providers for each team which can be handled by team admins.
## How Has This Been Tested?

Tested manually for GitLab provider, other tests are wip.
## Screenshots (if appropriate):

Dashboard App Anchor Support
## Types of changes
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
## TODO
- [x] Move `foreignAuth` data into it's own collection from `JUser`
- [x] Keep sensitive information about team oauth integration in another collection
- [x] Provide UI for team admins to set up/disable oauth integrations (GitLab as OAuth provider)
- [x] Check and verify provided oauth endpoint and provided credentials
- [x] Update existing GitLab integration to work with new foreignAuth mechanism
- [x] Disable deprecated oauth handlers
- [x] Make sure only one `foreignId, provider, username, group` on `JForeignAuth`
- [x] Remove existing oauth data on oauth setup change
- [x] Remove sensitive information of oauth integration from `JGroupData` collection on integration disable
- [x] Bongo improvements for authenticateUser flow and not-handled callbacks
- [x] Anchor support for Dashboard sections
- [x] Show onboarding notifications to team members once oauth integration is set-up by team admin

PR #10721, opened 2017-03-07
Delete team: Make sure slack integration and JForeignAuth data is cleaned

### LINKED ISSUE #10707: Slack integration is kept when a team is deleted and a new team is created with the same team url.

hakankaradis:
Fixes #10707

if there is any slack authentication made by any user before deleting the team, we are removing now

### DIFF
--- a/workers/social/lib/social/models/group/index.coffee
+++ b/workers/social/lib/social/models/group/index.coffee
@@ -1705,6 +1705,12 @@ module.exports = class JGroup extends Module
         JName.one { name: slug }, (err, name) ->
           removeHelper name, err, next, 'JName'
 
+      (next) ->
+        # unlink slack if anyone in this team auth slack previously
+        JUser = require '../user'
+        JUser.update { "foreignAuth.slack.#{slug}": { $exists: true } }, { $unset: { "foreignAuth.slack.#{slug}": 1 } }, (err) ->
+          kallback 'JForeignAuth', next, err
+
       (next) =>
         Relationship.remove { sourceId : @getId() }, next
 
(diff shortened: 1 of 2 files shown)


workers/social/lib/social/models/group/index.coffee
```diff
@@ -1705,6 +1705,12 @@ module.exports = class JGroup extends Module
         JName.one { name: slug }, (err, name) ->
           removeHelper name, err, next, 'JName'
 
+      (next) ->
+        # unlink slack if anyone in this team auth slack previously
+        JUser = require '../user'
+        JUser.update { "foreignAuth.slack.#{slug}": { $exists: true } }, { $unset: { "foreignAuth.slack.#{slug}": 1 } }, (err) ->
+          kallback 'JForeignAuth', next, err
```
gokmen: we no longer have `JUser.foreignAuth.*`[1] in the schema. It seems the slack integration is still using that, it's ok to cleanup them but it would be nice if we can do it over `socialApi` since it's added by that. cc/ @cihangir 

[1] https://github.com/koding/koding/pull/9057

gokmen (approved):
LGTM except the issue I commented, which is not part of this for now.

gokmen (approved):

workers/social/lib/social/models/group/index.coffee
cihangir: we had slack integration before migration of JForeignAuth. it seems we forgot to migrate that as well, best thing to do would be migrating slack to JForeignAuth

cihangir (commented):
👍  LGTM

PR #10723, opened 2017-03-08
Machine mount prefetch

ppknap:
This PR adds mount file prefetching. In order to reduce amount of downloaded files, prefetch can detect `.git` directory and reduce received files to `.git` folder content.

Depends on: ~#10637~

## Motivation and Context
Speed up mounting.

## How Has This Been Tested?
Manually.

https://asciinema.org/a/22bb04z4astfdoa4ju8r0d73u

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinegroup/mount.go
+++ b/go/src/koding/klient/machine/machinegroup/mount.go
@@ -11,7 +11,6 @@ import (
 	"koding/klient/machine/machinegroup/syncs"
 	"koding/klient/machine/mount"
 	msync "koding/klient/machine/mount/sync"
-	"koding/klient/machine/transport/rsync"
 	"koding/klientctl/ssh"
 )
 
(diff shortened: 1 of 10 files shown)


go/src/koding/klient/machine/machinegroup/mount.go
```diff
@@ -184,20 +206,74 @@ func (g *Group) AddMount(req *AddMountRequest) (res *AddMountResponse, err error
...
-	if err = g.sync.Add(mountID, req.Mount, g.nb, g.sb, dynSSH, dynClient); err != nil {
+	if err = g.sync.Add(addReq); err != nil {
 		g.log.Error("Synchronization of %s mount failed: %s", mountID, err)
 		return nil, err
 	}
 
+	sc, err := g.sync.Sync(mountID)
+	if err != nil {
+		// Panic here since missing mount here should be impossible.
+		panic("mount " + req.Mount.String() + " doesn't exist")
```
rjeczalik: How does it differ from UpdateIndex, where we return an error instead?

go/src/koding/klient/machine/mount/sync/sync.go
```diff
@@ -240,6 +240,58 @@ func (s *Sync) Info() *Info {
...
+		Download:        true,
+		SourcePath:      s.m.RemotePath + "/",
+		DestinationPath: s.CacheDir() + "/",
+		Username:        username,
+		Host:            host,
+		SSHPort:         port,
+	}
+
+	// Look for git VCS.
+	if n, ok := s.idx.LookupAll(".git"); ok {
```
rjeczalik: ok && n.IsDir()

go/src/koding/klient/machine/machinegroup/mount.go
ppknap: Five lines above we created a mount and new syncer(which we return here). The difference is that the only possible error returned from `g.sync` is `MountNotFound`, but it must be found since we just added it. Thus, we panic instead of returning error == the panic is for the sake of error handling.

UpdateIndex may receive mountID which doesn't exist, so the error is possible and logically valid.

rjeczalik (approved):

PR #10726, opened 2017-03-08
klient/os: add Environ wrapper

rjeczalik:
Used by os.exec handler to merge environmental variables
for a user process.

PR #10728, opened 2017-03-08
klient/os: add exec/kill methods

rjeczalik:
deps: https://github.com/koding/koding/pull/10726

This PR adds the following kite methods:

  - os.exec
  - os.kill

The os.exec methods is a reworked version of existing
exec method with a new API that is needed for KD. Since
the exec method is still used on client side, it's being
marked as deprecated and new version rolled out.

The os.kill method addresses shortcoming in current "kd run"
implementation - if user aborts command execution with CTRL+C,
the command won't get killed on a remote machine.

New implementation is going to call machine.kill upon exit,
which effectively will kill the user process on the remote side.

The os.kill method is limited to only those processes that
were started with os.exec.

### DIFF
--- a/go/src/koding/klient/os/util_test.go
+++ b/go/src/koding/klient/os/util_test.go
@@ -1,0 +1,254 @@
+package os_test
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	kos "koding/klient/os"
+
+	"github.com/koding/kite"
+	"github.com/koding/kite/dnode"
+)
+
+const timeout = 5 * time.Second
+
+func die(v ...interface{}) {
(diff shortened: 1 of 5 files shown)


go/src/koding/klient/os/util_test.go
```diff
@@ -0,0 +1,254 @@
...
+const timeout = 5 * time.Second
+
+func die(v ...interface{}) {
+	fmt.Fprintln(os.Stderr, v...)
+	os.Exit(1)
+}
+
+// TestCommandHelper is used by TestExec / TestKill tests to
+// execute test code as an external process.
+func TestCommandHelper(t *testing.T) {
```
ppknap: :+1:

ppknap (approved):

rjeczalik:
@cihangir Ptal, I'm going to address your comments if you have any in a separate PR.

cihangir (commented):
👍

PR #10734, opened 2017-03-08
klient/machine: FUSE notfier doesn't start after restart returning `file exists` error

ppknap:
The error is reported by `os.MkdirAll` function when mounted directory already exists. 

Additionally, `Umount` function call was moved at the top of `NewFilesystem` function in order to prevent another `os.MkdirAll` error: `transport endpoint is not connected` caused by FUSE itself. 

## Motivation and Context
Make mount work after kd restart.

## How Has This Been Tested?
Manually

## Types of changes


rjeczalik (approved):

rjeczalik:
> The error is reported by os.MkdirAll function when mounted directory already exists.

Unexpected, MkdirAll should return nil if the dir already exists 😱

ppknap:
> Unexpected, MkdirAll should return nil if the dir already exists :scream:

I was surprised as well.

PR #10736, opened 2017-03-09
klient/machine/mount: use lazy umount on linux when normal one fails

### LINKED ISSUE #10735: FUSE umount doesn't work when mounted directory is accessed during the process.
KD machine umount returns success when called on mount which is in use. However, the folder is not umounted on linux. This is like a difference between `umount` and `umount -l`

Problem: https://asciinema.org/a/64mlyhfcyakxwwaqqwjudqc9r

ppknap:
This PR adds fallback to lazy umount on linux.

Unfortunatelly, `jacobsa/fuse` umount function [cannot be parameterized](https://github.com/jacobsa/fuse/blob/0b9db0a706bc0248f8a37cba1dd5372b4c9a8b72/unmount_linux.go#L11).

Fixes: #10735

## Motivation and Context
Destination directory is not umounted when it's used by another process.

## How Has This Been Tested?
Manually.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/notify/fuse/fuse.go
+++ b/go/src/koding/klient/machine/mount/notify/fuse/fuse.go
@@ -477,17 +477,24 @@ func (fs *Filesystem) Destroy() {
 	fs.mu.Unlock()
 }
 
+// Umount unmounts FUSE filesystem.
 func Umount(dir string) error {
+	umountCmd := func(name string, args ...string) error {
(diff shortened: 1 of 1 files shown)


rjeczalik (approved):

rjeczalik:
@cihangir There are a number of mount-related PRs opened - merging them for now to continue with QA, if you have any comments we will address them in a separate PR.

cihangir:
👍

PR #10745, opened 2017-03-09
remoteapi: regenerate swagger client

rjeczalik:


go/src/koding/remoteapi/models/j_group_role.go
```diff
@@ -16,7 +16,7 @@ type JGroupRole struct {
 	// id
 	ID string `json:"_id,omitempty"`
 
-	// is configurable
+	// is configureable
```
ppknap:
```
$ scripts/wercker/run-command scripts/gometalinter.sh
/pipeline/build/go/src/koding/remoteapi/models/j_group_role.go:19:7:warning: "configureable" is a misspelling of "configurable" (misspell)
```

rjeczalik: Since it's a field of jRole I'm not sure I can't just fix it and be backward compatible.

```
workers/social/lib/social/models/group/role.coffee
16:      isConfigureable :
26:    { title : 'moderator',  isDefault : yes,  isConfigureable: yes }
```

@gokmen wdyt?

ppknap: @rjeczalik whe have three options:

1) Fix it in jRole (which is most complicated)
2) Alter generated file by hand (which is dummy)
3) Remove spelling check from gometallinter (which is sad)

Personally, I don't think it's important, just pointing that tests failed because of this. If `1)` is too complicated we may end up with `2)`

gokmen: I believe we are not using that information anymore, it can be removed from `JRole` schema.

rjeczalik: I'll remove it in a separate PR. For now I fixed it manually.

What would be better instead of spellcheck linter is some bot that would send PR with fixes automagically. 😎

ppknap: and this bot must not change `klient` to `client`... :slightly_smiling_face:

ppknap (approved):

go/src/koding/remoteapi/models/j_group_role.go
gokmen: lol

PR #10769, opened 2017-03-14
socialapi/tests: unify test server and stub data creation

cihangir:


cihangir:
failed big time with this refactoring attempt. will revisit this sometime later with https://github.com/koding/koding/issues/10771

PR #10775, opened 2017-03-15
Countly Exporter

cihangir:


cihangir:
circle build has passed, we can ignore wercker ( it is already removed, but this PR has it )

go/src/socialapi/workers/countly/api/api.go
```diff
@@ -0,0 +1,123 @@
...
+			AppKey: group.Countly.AppKey,
+			APIKey: group.Countly.APIKey,
+		}, nil
+	}
+
+	app, err := c.client.CreateApp(&client.App{
+		Name:  slug,
+		Owner: c.globalOwner,
+	})
+	if err != nil {
```
rjeczalik: There might be a race between group.HasCountly and CreateApp. If CreateApp errors out with AlreadyExists (making this up, did not find any errors code on http://resources.count.ly/v1.0/reference#iappscreate), we could return this group instead.

go/src/socialapi/workers/countly/api/api.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"koding/db/mongodb/modelhelper"
+	"koding/tools/utils"
+	"net/http"
+	"net/url"
+	"socialapi/config"
+	"socialapi/models"
+	"socialapi/workers/common/response"
+	"socialapi/workers/countly/client"
```
rjeczalik: nit: please group import paths

https://github.com/golang/go/wiki/CodeReviewComments#imports

go/src/socialapi/workers/countly/exporter/exporter.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return err
+		}
+		// we persist in API call but here we have the old group, so update it
+		group.Countly.AppKey = res.AppKey
+		group.Countly.AppID = res.AppID
+		group.Countly.APIKey = res.APIKey
+	}
+
+	events := []client.Event{client.Event{
```
rjeczalik: nit: `gofmt -s` can simplify this declaration

go/src/socialapi/workers/countly/exporter/exporter.go
```diff
@@ -0,0 +1,105 @@
...
+func (c *CountlyExporter) Close() error {
+	return nil
+}
+
+func getGroupName(event *eventexporter.Event) string {
+	if len(event.Properties) == 0 {
+		return ""
+	}
+
+	if name, ok := event.Properties["groupName"]; ok {
```
rjeczalik: The following won't panic if name is not a string.

```go
if name, ok := event.Properties["groupName"].(string); ok {
    return name, nil
}
```

go/src/socialapi/workers/countly/api/api.go
cihangir: i dont do the imports, that is what goimport does, and i dont want to manage them by hand that is machine's job ;)

we may need to fork and customize it a little bit. wdyt?

go/src/socialapi/workers/countly/api/api.go
cihangir: App creation does not give error on same app name. only app key is unique ( has an index in that field ) Name is just a convenience. But you are right there will be race conditions ( if there is a chance it will happen )

We could utilize a global lock but thought it would be overkill ATM to implement ( not the implementation but having it here )

go/src/socialapi/workers/countly/exporter/exporter.go
cihangir: 👍  done

go/src/socialapi/workers/countly/exporter/exporter.go
cihangir: 👍

go/src/socialapi/workers/countly/api/api.go
rjeczalik: Not sure if tunning heuristics could help or changing the layout of Go code within koding.

I assume there's something that groups imports that do not begin with a known remote repo domain together in the first group. Since we're setting GOPATH inside koding repo, our import paths for local packages are not go-gettable.

E.g. if you instruct goimports explicitely what prefix do your local packages have, then the imports are grouped _almost_ correctly: 

```diff
$ goimports -local socialapi -w api.go
$ git diff
diff --git a/go/src/socialapi/workers/countly/api/api.go b/go/src/socialapi/workers/countly/api/api.go
index ed03ffd110..10ca911a78 100644
--- a/go/src/socialapi/workers/countly/api/api.go
+++ b/go/src/socialapi/workers/countly/api/api.go
@@ -1,10 +1,12 @@
 package api

 import (
-       "koding/db/mongodb/modelhelper"
-       "koding/tools/utils"
        "net/http"
        "net/url"
+
+       "koding/db/mongodb/modelhelper"
+       "koding/tools/utils"
+
        "socialapi/config"
        "socialapi/models"
        "socialapi/workers/common/response"
```

rjeczalik (approved):

cihangir:
tests are completed both wercker & circle ( wercker shows as in progress but in fact it is done )

cihangir:
merging this & will deploy to sandbox

go/src/socialapi/workers/countly/api/api.go
cihangir: I was just going to suggest "group packages together if they start with koding or socialapi"

somthing like `-group "socialapi,koding"` and output would be like 

```diff
$ goimports -local socialapi -local koding -group "socialapi,koding" -w api.go
$ git diff
diff --git a/go/src/socialapi/workers/countly/api/api.go b/go/src/socialapi/workers/countly/api/api.go
index ed03ffd110..10ca911a78 100644
--- a/go/src/socialapi/workers/countly/api/api.go
+++ b/go/src/socialapi/workers/countly/api/api.go
@@ -1,10 +1,12 @@
 package api

 import (
-       "koding/db/mongodb/modelhelper"
-       "koding/tools/utils"
        "net/http"
        "net/url"
+
+       "koding/db/mongodb/modelhelper"
+       "koding/tools/utils"
        "socialapi/config"
        "socialapi/models"
        "socialapi/workers/common/response"
``` 

may be we could open a PR too, i now some people uses a modified version of goimports just for this reason

rjeczalik: Yes, that's good idea.

I _think_ it'd be even possible to improve the heuristics, to mark a package as local and group it accordingly if it exists in the same GOPATH as the source file that is being processed. I haven't read the source code of the goimports thought.

PR #10778, opened 2017-03-15
klient/machine: do not try to create working dir until new mount is requested

ppknap:
This PR prevents klient from stopping on initialization phase when `.cache` directory is not accessible.

## Description
Each mount uses its own MkdirAll during initialization so the removed call is not needed anyway.

## Motivation and Context
Make klient start.

## How Has This Been Tested?
Manually

### DIFF
--- a/go/src/koding/kites/config/cache.go
+++ b/go/src/koding/kites/config/cache.go
@@ -94,20 +94,16 @@ func KodingHome() string {
 	return home
 }
 
-// KodingCacheHome gives the path of the koding cache directory.
+// KodingMounts gives the path of the koding cache directory.
 //
-// The default value is overwritten with KODING_CACHE_HOME env,
+// The default value is overwritten with KODING_MOUNTS env,
 // if it points to a valid directory.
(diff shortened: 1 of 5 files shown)


go/src/koding/kites/config/cache.go
```diff
@@ -105,6 +105,10 @@ func KodingCacheHome() string {
 		cache = filepath.Join(CurrentUser.HomeDir, ".cache", "koding")
 	}
 
+	// Best-effort attempt, ignore errors.
+	_ = os.MkdirAll(cache, 0755)
+	_ = util.Chown(cache, CurrentUser.User)
```
rjeczalik: I'm not convinced this is the right place.

I'd rather add it [here](https://github.com/koding/koding/blob/master/go/src/koding/klientctl/install.go#L286) and [here](https://github.com/koding/koding/blob/master/go/src/koding/klientctl/update.go#L211-L212).

rjeczalik (approved):

cihangir (approved):

PR #10785, opened 2017-03-16
klient/machine: get klient from pool on each machine client call

### LINKED ISSUE #10754: Keep klient open when using klientpool
Prevent errors like: sendError: `can't send, client is closed `

ppknap:
This PR fixes: #10754

## Description
When klient produced by pool loses connection, it [Closes itself and removes from the pool](https://github.com/koding/koding/blob/354b39c8a7b28cf879a3e97cd9059241e331b70e/go/src/koding/kites/kloud/klient/klient.go#L96). However, machine client still reuses taken and already closed klient. Thus, producing `can't send, client is closed` error.

This PR creates a new Client wrapper that will always get klient from the pool before connection.

## Motivation and Context
Bug fix.

## How Has This Been Tested?
Manually.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/client/kite.go
+++ b/go/src/koding/klient/machine/client/kite.go
@@ -5,7 +5,10 @@ import (
 	"time"
 
 	"koding/kites/kloud/klient"
+	"koding/klient/fs"
 	"koding/klient/machine"
+	"koding/klient/machine/index"
+	"koding/klient/os"
 
 	"github.com/koding/kite"
 )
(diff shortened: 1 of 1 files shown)


rjeczalik (approved):

go/src/koding/klient/machine/client/kite.go
```diff
@@ -46,10 +46,6 @@ func (kb *KiteBuilder) Ping(dynAddr DynamicAddrFunc) (machine.Status, machine.Ad
 // Build builds new kite client that will connect to machine using provided
 // address.
 func (kb *KiteBuilder) Build(ctx context.Context, addr machine.Addr) Client {
-	if _, err := kb.pool.Get(addr.Value); err != nil {
-		return NewDisconnected(ctx)
-	}
-
```
rjeczalik: I thought it's actually a nice optimisation, so if client is disconnected, you don't lookup the pool on each call.

ppknap: Indeed it is, but it's also duplicated by `Ping` [method](https://github.com/koding/koding/blob/7ff210291a6f0a7358c8470d709446119a13876b/go/src/koding/klient/machine/client/kite.go#L36). With this one it would be harder to recover. Also, `Build` is done only once per valid klient.

cihangir (approved):

PR #10791, opened 2017-03-17
Read Datadog API key from KONFIG object environment variable

szkl:

### DIFF
--- a/.ebextensions/099_datadog.config
+++ b/.ebextensions/099_datadog.config
@@ -23,6 +23,8 @@ container_commands:
         command: "cp .ebextensions/datadog/datadog.repo /etc/yum.repos.d/datadog.repo; yum -y makecache; yum -y install datadog-agent; /etc/init.d/datadog-agent stop"
     91setup_datadog:
         command: "cp .ebextensions/datadog/datadog.conf /etc/dd-agent/datadog.conf"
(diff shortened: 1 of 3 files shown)


.ebextensions/099_datadog.config
```diff
@@ -23,6 +23,8 @@ container_commands:
         command: "cp .ebextensions/datadog/datadog.repo /etc/yum.repos.d/datadog.repo; yum -y makecache; yum -y install datadog-agent; /etc/init.d/datadog-agent stop"
     91setup_datadog:
         command: "cp .ebextensions/datadog/datadog.conf /etc/dd-agent/datadog.conf"
+    91setup_datadog_apikey:
+        command: "bash -c 'source .env.sh && echo api_key: $KONFIG_DATADOG_API_KEY >>/etc/dd-agent/datadog.conf'"
```
cihangir: 1) this will end up appending datadog key in every deployment, better to have it once
2) i believe it would be safer to append the key first, then move it to its location

szkl: Previous command will be overwriting `/etc/dd-agent/datadog.conf` file, won't it?

cihangir: yes you are right, sorry for the confusion.

cihangir (approved):

PR #10808, opened 2017-03-20
klient/machine: unsync all poped events, even these deprecated

ppknap:
Without this change `kd machine mount list` could display misleading information about
SYNCed files.

## Description
Some file synchronization events may become invalid after they are poped from the queue(changing sync directon etc). We mark these changes deprecated so they will not be run(if possible). However when we pop form the queue we increase Sync counter. Which was decreased only for non deprecated events but should be always. 

## Motivation and Context
Incorrect information about mounts.

## How Has This Been Tested?
Manually.

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/sync/anteroom.go
+++ b/go/src/koding/klient/machine/mount/sync/anteroom.go
@@ -201,7 +201,12 @@ func (a *Anteroom) detach(path string, id uint64) {
 	defer a.mu.Unlock()
 
 	if ev, ok := a.evs[path]; ok && ev.ID() == id {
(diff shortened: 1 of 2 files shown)


rjeczalik (approved):

cihangir (approved):

cihangir:
failed at wercker with 

```
--- FAIL: TestCreateBalance (1.02s)
	create_test.go:125: want err = nil; got timed out after 1s
```

never seen before

ppknap:
@cihangir this fail is not related, will look at this and try to make fix in a separate PR

PR #10829, opened 2017-03-22
Introduce deploy job to Circle CI configuration

szkl:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,18 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: deploy
+          command: |
+            case "$CIRCLE_BRANCH" in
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -23,6 +23,18 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: deploy
```
cihangir: could we change the name here to smth like "trigger deploy", had really hard time figuring out what is the difference between this and the bottom change :)

cihangir (approved):

cihangir:
you can send the change in a diff PR

.circleci/config.yml
szkl: Sure. I just noticed this comment.

PR #10858, opened 2017-03-24
go/klientctl: add basic support for possible metrics

cihangir:
This is a demonstration PR for how we are going to wrap commands with metrics.

### DIFF
--- a/go/src/koding/klientctl/ctlcli/ctlcli.go
+++ b/go/src/koding/klientctl/ctlcli/ctlcli.go
@@ -116,9 +116,10 @@ func ExitErrAction(f ExitingErrCommand, log logging.Logger, cmdName string) cli.
 		}
 
 		Close()
-		ExitFunc(exit)
-
-		return nil
+		if exit == 0 {
+			return nil
+		}
+		return fmt.Errorf("exit code: %d", exit)
 	}
 }
 
--- a/go/src/koding/klientctl/main.go
+++ b/go/src/koding/klientctl/main.go
@@ -758,9 +758,30 @@ func run(args []string) {
 		)
 	}
 
+	app.Commands = wrapActions(app.Commands)
 	app.Run(args)
 }
 
+func wrapActions(commands []cli.Command) []cli.Command {
+	for i, command := range commands {
+		if command.Action != nil {
+			commands[i].Action = createActionFunc(command.Action.(cli.ActionFunc))
+		}
+		if len(command.Subcommands) > 0 {
+			commands[i].Subcommands = wrapActions(command.Subcommands)
+		}
+	}
+	return commands
+}
+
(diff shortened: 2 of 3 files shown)


go/src/koding/klientctl/ctlcli/ctlcli.go
```diff
@@ -116,9 +116,10 @@ func ExitErrAction(f ExitingErrCommand, log logging.Logger, cmdName string) cli.
 		}
 
 		Close()
-		ExitFunc(exit)
```
cihangir: i am not sure why we added os.exit here. It prevents running AfterFunc's of any command.

rjeczalik: This is to ensure kd terminate with a proper exit code. This is kind of workaround for the cli package that is used in kd. Once moved to cobra, will be removed.

cihangir: Hm, i need the timing per command, could we generalize and exit with 1 if there is an error till we migrate to cobra?

rjeczalik: The way I'm workarounding this design is to register a function that will get executed on exit - ctlcli.OnExit. I'm not sure if making the exit code 1 won't break any behaviour somewhere.

ppknap: Yeah, actually this was added when the codegangsta/cli returned method had a different signature(`func(c *cli.Context)` instead of `func(c *cli.Context) error`) and at that times this was quite valid.

ppknap: > Hm, i need the timing per command, could we generalize and exit with 1 if there is an error till we migrate to cobra?

I think we always use `1` as the exit code for new commands.

go/src/koding/klientctl/main.go
```diff
@@ -758,9 +758,30 @@ func run(args []string) {
...
+	app.Commands = wrapActions(app.Commands)
 	app.Run(args)
 }
 
+func wrapActions(commands []cli.Command) []cli.Command {
+	for i, command := range commands {
+		if command.Action != nil {
+			commands[i].Action = createActionFunc(command.Action.(cli.ActionFunc))
+		}
+		if len(command.Subcommands) > 0 {
```
ppknap: just a nitpick - we don't need this condition.

ppknap (approved):

go/src/koding/klientctl/ctlcli/ctlcli.go
cihangir: This is the only place where we return non 0-1  return code in klientctl https://github.com/koding/koding/blob/master/go/src/koding/klientctl/logcmd/command.go#L90

rjeczalik (approved):

PR #10860, opened 2017-03-26
kloud/metadata: add cloud-init simple merging support

rjeczalik:
The Merge function is going to be used to merge kloud's
cloud-init, which will be responsible for deploying
klient service with a customized provision.sh, into
user's cloud-init content.

After metadata package is ready, kloud is going
to be refactored to not depend on S3 bucket lookup
to install a klient, thus S3 credential configuration
can be removed from kloud, making it effectively
do not depend directly on S3 credentials anymore.

After the above is in place, kloud/userdata package
is going to be removed.


go/src/koding/kites/kloud/metadata/cloudinit_test.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		"testdata/conflict_write_files.yml": {
+			conflict: []string{"write_files", "1", "path"},
+		},
+	}
+
+	for yml, cas := range cases {
+		t.Run(filepath.Base(yml), func(t *testing.T) {
+			if *update {
+				ci, err := parseCloudInit(yml)
```
cihangir: just a personal preference, i would write the code for updating in a diff function since it is not directly related with the tests + it could decrease the function body line count

cihangir (approved):

go/src/koding/kites/kloud/metadata/cloudinit_test.go
rjeczalik: Done.

ppknap (approved):

Issue #10825, opened 2017-03-22
Warning message is not displayed correctly on new notifications.


diclee:
## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/24176930/74f57720-0e5c-11e7-9693-8220202b83bf.png)

- [ ] Also, the warning icon is not aligned correctly.

diclee:
![image](https://cloud.githubusercontent.com/assets/8261892/24311996/d9180180-1093-11e7-91a4-6c55b06ced10.png)

diclee:
Verify passed on sandbox.

PR #10863, opened 2017-03-27
NotificationView: decode message and styl update

hakankaradis:
- Decode the error message
- Vertically align: icon and text content
- Button on hover:  remove text-decoration

Fixes: #10825


sinan (approved):
@hakankaradis the css 👑

PR #10881, opened 2017-03-28
config: add a way to run countly in local env

cihangir:
Running steps

./configure --config dev --countlyPath <full path to countly server>
./run buildservices force
./run

key point here is to have “--countlyPath” in the configuration params, otherwise it will work in docker.


cihangir:
files under  "deployment/countly/config/" copied from https://github.com/koding/tf_countly/tree/master/config

gokmen (approved):

sinan (approved):
👍

szkl (approved):

Issue #10885, opened 2017-03-28
Managed VM is added to and removed from sidebar after reloading the page.


diclee:
## Current Behavior
Adding a managed VM:
![gif](http://g.recordit.co/QPt1XAIfcl.gif)
http://recordit.co/QPt1XAIfcl

Removing a managed VM:
![gif](http://g.recordit.co/wUPlLYKfgi.gif)
http://recordit.co/wUPlLYKfgi

diclee:
Verify passed on latest.

Issue #10884, opened 2017-03-28
Deleted stack template is listed on dashboard until the page is reloaded.


diclee:
## Current Behavior
![gif](http://g.recordit.co/ZqXdr3J2HC.gif)
http://recordit.co/ZqXdr3J2HC

And it's also displayed on sidebar if you open dashboard.

diclee:
It's still displayed on dashboard for the member until reloading the page:
![gif](http://g.recordit.co/wxQIxbasMs.gif)
http://recordit.co/wxQIxbasMs

diclee:
Verify passed on sandbox.

Issue #10883, opened 2017-03-28
User is not switched to the machine when it's initialized from sidebar.


diclee:
## Current Behavior
![gif](http://g.recordit.co/gcjFttp3Jo.gif)
http://recordit.co/gcjFttp3Jo

diclee:
Verify passed on latest.

diclee:
This issue is reproducible again:
![gif](http://g.recordit.co/6rgwQBavMd.gif)
http://recordit.co/6rgwQBavMd

diclee:
Verify passed on sandbox.

Issue #10879, opened 2017-03-28
'Preview' link for readme is misaligned.


diclee:
## Current Behavior
![image](https://cloud.githubusercontent.com/assets/8261892/24383357/4f28a116-1311-11e7-82aa-3ec65a2f9c4d.png)

diclee:
Verify passed on latest.

PR #10890, opened 2017-03-29
Load improvements fixes

gokmen:
ref: http://recordit.co/0sLLdky3qy

This PR fixes the issues identified below which were introduced mostly the switch to `ComputeStorage` on `Load Improvements` PR https://github.com/koding/koding/pull/10671 

- [x] fixes https://github.com/koding/koding/issues/10885
- [x] fixes https://github.com/koding/koding/issues/10884
- [x] fixes https://github.com/koding/koding/issues/10883
- [x] fixes https://github.com/koding/koding/issues/10879
- [x] fixes https://github.com/koding/koding/issues/10709

### DIFF
--- a/client/app/lib/components/sidebarstacksection/index.coffee
+++ b/client/app/lib/components/sidebarstacksection/index.coffee
@@ -104,8 +104,7 @@ module.exports = class SidebarStackSection extends React.Component
       when 'Clone'
         computeController.fetchStackTemplate templateId, (err, template) =>
           return @onMenuItemClickError 'cloning'  if err
-
-          EnvironmentFlux.actions.cloneStackTemplate template
+          EnvironmentFlux.actions.cloneStackTemplate template  if template
       when 'Destroy VMs' then deleteStack { stack }
       when 'VMs'
         firstMachineId = stack.get('machines').first.get '_id'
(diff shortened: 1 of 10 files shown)


client/app/lib/components/sidebarstacksection/index.coffee
```diff
@@ -117,17 +116,18 @@ module.exports = class SidebarStackSection extends React.Component
...
-          computeController.setStackTemplateAccessLevel template, 'group'
+          if template
+            computeController.setStackTemplateAccessLevel template, 'group'
       when 'Make Private'
         computeController.fetchStackTemplate templateId, (err, template) =>
           return @onMenuItemClickError 'cloning'  if err
 
-          computeController.setStackTemplateAccessLevel template, 'private'
+          if template
+            computeController.setStackTemplateAccessLevel template, 'private'
```
usirin: I think not having a `template` can be considered as `error` to give a proper feedback to user. WDYT?

usirin (approved):
Only 1 minor issue (which can be considered as a non-issue as well), rest looks good.

client/app/lib/components/sidebarstacksection/index.coffee
gokmen: These should handled correctly that I've seen so many places which never takes into account the existence of these objects. For now I left them as is with existence checks and later on we can handle all these cases separately.

PR #10926, opened 2017-03-31
go/models: implement GroupData into koding/models

cihangir:


go/src/koding/db/models/data.go
```diff
@@ -0,0 +1,59 @@
...
+
+	dt, ok := findPath(strings.Split(key, "."), *d)
+	if !ok {
+		return "", ErrDataKeyNotExists
+	}
+
+	return dt, nil
+}
+
+func findPath(path []string, data Data) (interface{}, bool) {
```
rjeczalik: It could take ... string instead (`findPath(data Data, path ...string)`), so the API could be simpler:

```go
d, ok = findPath(testDataMap, nonExistingKey)
```

Instead of:

```go
d, ok = findPath([]string{nonExistingKey}, testDataMap)
```

go/src/koding/db/models/groupdata.go
```diff
@@ -0,0 +1,8 @@
+package models
+
+// GroupData holds a group's private info.
+type GroupData struct {
+	ID   string `json:"_id"`
+	Slug string `json:"slug"`
+	Data *Data  `bson:"data,omitempty" json:"data,omitempty"`
```
rjeczalik: I think it could be just `Data Data`, no need for pointer.

```go
type GroupData struct {
 	ID   string `json:"_id"`
 	Slug string `json:"slug"`
 	Data Data  `bson:"data,omitempty" json:"data,omitempty"`
```

rjeczalik: Btw aren't there tags for bson missing? (e.g. `bson:"_id"`)

cihangir: 1) What advantages we are going to have when we dont have a pointer?
2) if bson tags are not defined, lowercased version of the fields are used, so we dont have to define. but i've realized i used string in id field, need to change to bson.ObjectId

rjeczalik: 1. Less indirections, no need for nil-checks. This part from Get() can be removed

```go
+	if d == nil {
 +		return "", ErrDataKeyNotExists
 +	}
```

And all Get, GetString etc. methods could be defined on the value receiver (`func (d Data) Get(...`).

Also:

```
If the receiver is a map, func or chan, don't use a pointer to them. If the receiver is a slice and the method doesn't reslice or reallocate the slice, don't use a pointer to it.
```
https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

2. 👍

cihangir: I was under the impression that [Bar.Data](https://play.golang.org/p/S2GAI9Xp2w) wouldnt be nil and that was my main driving force to use pointer.
Also:
`when in doubt, use a pointer receiver.`

cihangir: having `json:"data,omitempty"` does not change the behaviour as well.

cihangir:
@rjeczalik build has passed, ptal

rjeczalik (approved):

rjeczalik:
@cihangir 👍

PR #10927, opened 2017-03-31
Use JGroupData for countly credentials

cihangir:
This is not a backward compatible change, we need to run migrations during switch operation.


cihangir:
@diclee this PR will prevent gitlab integration from working properly until we run the migrations. Mostly prob while you are testing on latest it will fail. fyi.

.circleci/config.yml
```diff
@@ -18,6 +18,12 @@ jobs:
 
       - image: redis
 
+      # countly will listen at port 80 till we can give -p "$KONFIG_COUNTLY_APIPORT:80" 
+      - image: koding/countly-server:latest
```
cihangir: cc// @szkl

client/analytics/lib/analyticsappview.js
```diff
@@ -13,6 +13,12 @@ export default class AnalyticsAppView extends JView {
 
   pistachio () {
     // return 'sinan\'s analytics'
-    return '<iframe src="http://dev.koding.com:8090/countly" frameborder="0" width="100%" height="100%" border="0"></iframe>'
+    
+    // TODO: replace username with group.slug and replace apiKey with group's api key
+    var username =  'admin'
```
cihangir: @sinan we need to get the real info with `remote.api.JGroupData.fetchByKey 'countly', callback`

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
         type      : String
         validate  : require('../name').validateName
         set       : (value) -> value.toLowerCase()
+      payload     : Object
```
cihangir: @gokmen i had to rename this field, bongo was overriding it. i tried to refactor bongo but that required a lot of effort and reverted my changes.

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+    availableKeys = [ 'countly' ]
+
+    if availableKeys.indexOf(path) is -1
+      return callback new Error 'path is forbidden'
+
+    slug = client.context.group
+    JGroupData.fetchDataAt slug, path, callback
+
+
+  @fetchDataAt: (slug, path, callback) ->
```
cihangir: moved the functionality that we have in group/index to here

workers/social/lib/social/models/proxy/index.coffee
```diff
@@ -1,86 +0,0 @@
-jraphical   = require 'jraphical'
```
cihangir: not sure how these survived my clean up

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+      JGroupData.create slug, callback
+
+  # fetchByKey fetches given path from GroupData if only they are available
+  # within the group.
+  #
+  # @param {String} path
+  #   Provide the path you would like to fetch.
+  #
+  # @return {Object} persisted value.
+  @fetchByKey = secure (client, path, callback) ->
```
gokmen: we need to use `permit` here otherwise anyone can call this over any `JGroup` instances.

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+    # for more granular control, specify each sub key as well.
+    availableKeys = [ 'countly' ]
```
gokmen: we can define these keys in config maybe.

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+
+    if availableKeys.indexOf(path) is -1
+      return callback new Error 'path is forbidden'
```
gokmen: `unless path in availableKeys`

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+
+    slug = client.context.group
```
gokmen: better to check `client` existence, if you use `permit` it'll be there probably.

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -15,5 +19,83 @@ module.exports = class JGroupData extends Module
...
+
+      payload = if data then getAt data.payload, path else null
+      return callback null, payload
+
+  # see docs on JGroup::modifyData
+  modifyData : (slug, data, callback) ->
+    return callback new Error 'slug is required' unless slug
+
+    # it's only allowed to change followings
+    whitelist  = ['github.organizationToken', 'test_key__' ]
```
gokmen: maybe we can move these too to config.

gokmen (changes requested):
This will break GitHub flow with organization setup and also login via GitHub. We may need to check that out. And on the other hand we also need to define a way to set accessLevel on group data; JGroupData was designed to keep secret data of group which only admins/owners of the group can read/write but with this change it will allow any member (right now anyone but it can be fixed see https://github.com/koding/koding/pull/10927/files#r110040692) to read them. In GitHub case it's dangerous since any member of the team can access the organization token and so on.

workers/social/lib/social/models/group/groupdata.coffee
cihangir: permitted version is in JGroup as you wrote it, nothing changed on that flow [see here](https://github.com/koding/koding/blob/master/workers/social/lib/social/models/group/index.coffee#L580-L585)

"fetchDataAt" is used with permit and only allows admins
"fetchByKey" is used without permit and only allows fetching "predefined keys" by anyone in the team

workers/social/lib/social/models/group/groupdata.coffee
cihangir: i didnt move them to config since we dont expect them to change by diff envs. do we?

workers/social/lib/social/models/group/groupdata.coffee
cihangir: 👍

workers/social/lib/social/models/group/groupdata.coffee
cihangir: 👍

szkl (approved):

workers/social/lib/social/models/group/groupdata.coffee
gokmen: currently `fetchByKey` can be called by **anyone** whether they are in team or not. we need to use `permit` here to check group membership.

workers/social/lib/social/models/group/groupdata.coffee
gokmen: you're right 👍

workers/social/lib/social/models/group/groupdata.coffee
cihangir: Yes that is right, but please see line 64. It will fetch the current group's data, not from the ones you don't belong to.

gokmen: Ah I see, but better to do it in correct way with `permit` which we can prevent `guest` sessions or we can change the permission type if needed.

go/src/koding/db/models/countly.go
```diff
@@ -7,3 +7,16 @@ type Countly struct {
...
+func (c *Countly) Valid() bool {
+	if c == nil {
+		return false
+	}
+
+	if c.APIKey == "" {
+		return false
+	}
+
+	return true
```
rjeczalik:
```
return c != nil && c.APIKey != ""
```

go/src/koding/db/mongodb/modelhelper/groupdata.go
```diff
@@ -14,38 +14,36 @@ const GroupDataColl = "jGroupDatas"
...
 
-	return gd, Mongo.Run(GroupDataColl, query)
+	return Mongo.Run(GroupDataColl, query)
 }
 
 // HasGroupDataPath checks if the given path has data or not.
 func HasGroupDataPath(slug, path string) (bool, error) {
-	gdp, err := GetGroupDataPath(slug, path)
+	gdp := &models.GroupData{}
+	err := GetGroupDataPath(slug, path, &gdp)
```
rjeczalik: gdp is already a pointer

```
err := GetGroupDataPath(slug, path, gdp)
```

rjeczalik (approved):

go/src/koding/db/models/countly.go
```diff
@@ -10,13 +10,5 @@ type Countly struct {
 
 // Valid checks if countly instance is valid.
 func (c *Countly) Valid() bool {
-	if c == nil {
```
cihangir: i find separate version more readable

gokmen (approved):

gokmen:
just rebased and force pushed to make sure tests passes, fyi.

PR #10948, opened 2017-04-04
Add hidden mount inspect command which tracks synchronization events.

ppknap:
This PR adds `kd machine mount inspect` command. The command is not visible for users and allows to inspect recent synchronization events.

There are up to 100 events stored inside mount's ring buffer. 

## Motivation and Context
Debug mount synchronization issues.

## How Has This Been Tested?
Unit tests, Manually.

https://asciinema.org/a/1qpr8xc3hlp41x2un22t0dlwe

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machinegroup/mount.go
+++ b/go/src/koding/klient/machine/machinegroup/mount.go
@@ -429,29 +429,32 @@ func (g *Group) InspectMount(req *InspectMountRequest) (*InspectMountResponse, e
 	}
 
 	res := &InspectMountResponse{}
-	if req.Sync {
-		if sc, err := g.sync.Sync(mountID); err == nil {
-			if res.History, err = sc.History(); err != nil {
-				g.log.Error("Cannot get mount %s syncing history: %s", mountID, err)
-			}
-		} else {
-			g.log.Warning("Mount %s is not synchronized: %s", mountID, err)
+	if !req.Sync {
+		return res, nil
+	}
+
+	if sc, err := g.sync.Sync(mountID); err == nil {
+		if res.History, err = sc.History(); err != nil {
+			g.log.Error("Cannot get mount %s syncing history: %s", mountID, err)
 		}
+	} else {
+		g.log.Warning("Mount %s is not synchronized: %s", mountID, err)
 	}
 
 	return res, nil
 }
 
 // getMountID looks up and converts provided identifier to mount ID.
-func (g *Group) getMountID(identifier string) (mount.ID, error) {
-	mountID, err := mount.IDFromString(identifier)
-	if err != nil {
-		absPath, e := filepath.Abs(identifier)
-		if mountID, err = g.mount.Path(absPath); e != nil || err != nil {
-			g.log.Error("Cannot found mount with identifier: %s", identifier)
-			return "", fmt.Errorf("unknown mount: %q", identifier)
-		}
+func (g *Group) getMountID(identifier string) (mountID mount.ID, err error) {
+	if mountID, err = mount.IDFromString(identifier); err == nil {
+		return
 	}
 
-	return mountID, err
+	absPath, e := filepath.Abs(identifier)
+	if mountID, err = g.mount.Path(absPath); e != nil || err != nil {
+		g.log.Error("Cannot found mount with identifier: %s", identifier)
+		return "", fmt.Errorf("unknown mount: %q", identifier)
+	}
+
+	return
 }
--- a/go/src/koding/klient/machine/machinegroup/mount.go
+++ b/go/src/koding/klient/machine/machinegroup/mount.go
@@ -10,6 +10,7 @@ import (
 	"koding/klient/machine/machinegroup/syncs"
 	"koding/klient/machine/mount"
 	"koding/klient/machine/mount/prefetch"
+	"koding/klient/machine/mount/sync/history"
 )
 
 // MountRequest defines machine group mount request.
@@ -361,13 +362,9 @@ func (g *Group) Umount(req *UmountRequest) (res *UmountResponse, err error) {
 	}
 
 	// Get mount ID from identifier.
-	mountID, err := mount.IDFromString(req.Identifier)
+	mountID, err := g.getMountID(req.Identifier)
 	if err != nil {
-		absPath, e := filepath.Abs(req.Identifier)
-		if mountID, err = g.mount.Path(absPath); e != nil || err != nil {
-			g.log.Error("Cannot found mount with identifier: %s", req.Identifier)
-			return nil, fmt.Errorf("unknown mount: %q", req.Identifier)
-		}
+		return nil, err
 	}
 
 	// Get mount machine.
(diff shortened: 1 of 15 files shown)


go/src/koding/klient/machine/machinegroup/mount.go
```diff
@@ -402,3 +399,59 @@ func (g *Group) Umount(req *UmountRequest) (res *UmountResponse, err error) {
...
+	}
+
+	// Get mount ID from identifier.
+	mountID, err := g.getMountID(req.Identifier)
+	if err != nil {
+		return nil, err
+	}
+
+	res := &InspectMountResponse{}
+	if req.Sync {
```
cihangir: you could simplify the indentations by early return statements

go/src/koding/klient/machine/machinegroup/mount.go
```diff
@@ -402,3 +399,59 @@ func (g *Group) Umount(req *UmountRequest) (res *UmountResponse, err error) {
...
+	}
+
+	return res, nil
+}
+
+// getMountID looks up and converts provided identifier to mount ID.
+func (g *Group) getMountID(identifier string) (mount.ID, error) {
+	mountID, err := mount.IDFromString(identifier)
+	if err != nil {
+		absPath, e := filepath.Abs(identifier)
```
cihangir: same here with ^^

go/src/koding/klient/machine/mount/sync.go
```diff
@@ -225,6 +229,15 @@ func (s *Sync) Info() *Info {
 	}
 }
 
+// History gets recent history of synchronized files.
+func (s *Sync) History() ([]*history.Record, error) {
+	if h, ok := s.s.(*history.History); ok {
+		return h.Get(), nil
+	}
+
+	return nil, errors.New("synchronization history in unavailable")
```
cihangir: s/in/is

go/src/koding/klient/machine/mount/sync/history/history.go
```diff
@@ -0,0 +1,161 @@
...
+		exC := h.s.ExecStream(evC)
+		for {
+			select {
+			case ex, ok := <-exC:
+				if !ok {
+					return
+				}
+
+				h.add(&Record{
+					CreatedAt: time.Now(),
```
cihangir: prob better to use UTC, time might change on day light saving ( or similar ) cases

go/src/koding/klient/machine/mount/sync/history/history_test.go
```diff
@@ -0,0 +1,74 @@
...
+				t.Fatalf("want err = nil; got %v (i:%d)", err, i)
+			}
+		}
+
+		recs := h.Get()
+		if len(recs) != change.RecLen {
+			t.Fatalf("want recs len = %d; got %d (i:%d)", change.RecLen, len(recs), i)
+		}
+
+		for j := len(recs) - 1; j >= 0 && (len(recs)-1-j) < 3; j-- {
```
cihangir: sorry but i didnt understand the logic here, why we care about the last 3 records? if 3 is important and a constant, we could simplify the loop

go/src/koding/klient/machine/transport/rsync/rsync.go
```diff
@@ -309,3 +317,20 @@ func dropCR(data []byte) []byte {
...
+func noNilMultiWriter(ws ...io.Writer) io.Writer {
+	var writers []io.Writer
+	for _, w := range ws {
+		if w != nil {
+			writers = append(writers, w)
+		}
+	}
+
+	if len(writers) == 0 {
+		return nil
```
cihangir: what about ioutil.Discard?

cihangir (approved):

go/src/koding/klient/machine/transport/rsync/rsync.go
ppknap: `os.DevNull` will be the default writer when nil is set to `exec.Command`.

go/src/koding/klient/machine/mount/sync/history/history_test.go
ppknap: I added a comment to the code explaining what's going on in this loop.

go/src/koding/klient/machine/machinegroup/mount.go
```diff
@@ -402,3 +399,62 @@ func (g *Group) Umount(req *UmountRequest) (res *UmountResponse, err error) {
...
+	}
+
+	res := &InspectMountResponse{}
+	if !req.Sync {
+		return res, nil
+	}
+
+	if sc, err := g.sync.Sync(mountID); err == nil {
+		if res.History, err = sc.History(); err != nil {
+			g.log.Error("Cannot get mount %s syncing history: %s", mountID, err)
```
rjeczalik: Why the err is not returned back to the caller?

go/src/koding/klient/machine/mount/sync/history/history.go
```diff
@@ -0,0 +1,161 @@
...
+		close(h.stopC)
+	})
+
+	return nil
+}
+
+// Get gets recorded history. Returned slice length will not exceed history
+// maximum size.
+func (h *History) Get() []*Record {
+	recs := make([]*Record, 0)
```
rjeczalik: var recs []*Record

go/src/koding/klientctl/machine.go
```diff
@@ -256,6 +256,42 @@ func MachineCpCommand(c *cli.Context, log logging.Logger, _ string) (int, error)
...
+
+	// Enable all options when none of them are set.
+	isSync := c.Bool("sync")
+	if !isSync {
+		isSync = true
+	}
+
+	opts := &machine.InspectMountOptions{
+		Identifier: idents[0],
+		Sync:       isSync,
```
rjeczalik: isSync is always true

rjeczalik (approved):

go/src/koding/klientctl/machine.go
ppknap: for now - yes, I'm planning to add other options so isSync won't be always true.

go/src/koding/klient/machine/mount/sync/history/history.go
ppknap: if we do so and user calls `Get` on empty history he will get `null` as an output instead of `[]`

go/src/koding/klient/machine/machinegroup/mount.go
ppknap: because there are plans to have other options than `Sync` and I want this function to be greedy and get as much data as it can.

go/src/koding/klient/machine/mount/sync/history/history.go
rjeczalik: And why it matters?

rjeczalik: Oh, you mean we rely on recs being non-nil to format correctly output for `kd machine mount inspect` command? Tricky, but let it be.

ppknap: What's tricky here? If we have not records we return empty array here. Nothing tricky:)

rjeczalik: > What's tricky here?

Formatting of `kd machine mount inspect` output should not rely on internal api (empty slice vs nil slice), especially when the behaviour is undocumented. The command handler itself should inspect if there's no history, and e.g. display `No history available.` or alike.

In some future some other developer can spot opportunity for refactoring, changing the line according to my suggestion, not knowing that it could break some other code elsewhere, especially when there's no test for this.

rjeczalik: I think changing the inline doc could make us both happy:

```go
// Get gets recorded history. Returned slice length will not exceed history
// maximum size.
//
// If there's no history available, the method returns empty, non-nil slice.
func (h *History) Get() []*Record {
```

PR #10957, opened 2017-04-07
Api improvements

gokmen:
Updates swagger definition for better code generation, see https://github.com/koding/koding-api-codegen also defines a way to login over api `JUser.login` becomes a public method which doesn't require an `APITOKEN` to work.

```js
var KodingApi = require("koding-api");

new KodingApi.JUserApi()
  .login({ username: "someuser", password: "somepassword", groupName: "someteam" })
  .then(data => console.log("result:", data)); 
     // data.replacementToken can be used as APITOKEN then.

```


szkl (approved):

workers/social/lib/social/models/user/index.coffee
```diff
@@ -643,6 +637,13 @@ module.exports = class JUser extends jraphical.Module
       return callback null, { isEligible: yes }
 
 
+  logout = (reason, clientId, callback) ->
```
cihangir: could we change the name to a more specific one like "terminateSessoin" when i first read logout, i thought you forget error handling in the function and outputting errors but turned out it is only called on error cases.

workers/social/lib/social/models/user/index.coffee
```diff
@@ -692,15 +693,17 @@ module.exports = class JUser extends jraphical.Module
...
+          if err
 
             logout 'error fetching account', clientId, callback
 
           else
-
             # A valid session, a valid user attached to
             # it voila, scenario #2
-            callback null, { session, account }
+            session.lastAccess = lastAccess = new Date
```
cihangir: we dont need to update session each time we get a client for it.
either lets remove this code or lets add an in memory cache where we will update this once a while say 1 time per hour.

workers/social/lib/social/remoteapi.coffee
```diff
@@ -112,7 +110,19 @@ updateSessionTimestamp = (session, callback) ->
     callback null, session
 
 
-fetchSession = (Models, token, callback) ->
+fetchSession = (Models, options, callback) ->
```
cihangir: "Models" looks a little bit strange :)

cihangir (dismissed):
after we fix/optimize session update, we can merge.

workers/social/lib/social/models/user/index.coffee
gokmen: Right it's too much for now, I'll disable it as you suggested but eventually we need to move session to Redis or some other in memory first db, we can implement the last activity updater later.

workers/social/lib/social/remoteapi.coffee
gokmen: "remoteapi" ?

Issue #10968, opened 2017-04-11
Terminal is getting broken in some cases.


diclee:
## Current Behavior
![gif](http://g.recordit.co/KeUr9it2B3.gif)
http://recordit.co/KeUr9it2B3

In above example it's getting highlighted after running 'ls /' command.

gokmen:
@rjeczalik this might be related with latest `cloud-init` updates that we need to dig into it. cc/ @szkl

rjeczalik:
@diclee Is it on sandbox or production? What's the kite ID and is username `veronicamars`?

diclee:
Verify passed on latest.

PR #10971, opened 2017-04-11
kloud/metadata: use embedded terminfo

rjeczalik:
I incorrectly assumed that `xterm-color` terminfo can be use by embedded screen - it made the produced output glitchy, as described by #10968.

This PR uses `screen-256color` terminfo from embedded distribution, which fixes the problem for newly deployed instances. I'm going to create another PR, which downloads and sets up embedded screen for existing klient deployments upon start, so on update klient uses working screen and continues to work after a update.

Also this PR fixes invalid hostname issue:

```                                                                                                     
sudo: unable to resolve host rjeczalik 
```

/cc @gokmen


gokmen (approved):

ppknap:
LGTM

ppknap (commented):
LGTM

PR #11007, opened 2017-04-18
JGroup: fix fetchDataAt over new JGroupData model

gokmen:
This fixes Login with GitHub issues as well as the token issues

### DIFF
--- a/workers/social/lib/social/models/group/groupdata.coffee
+++ b/workers/social/lib/social/models/group/groupdata.coffee
@@ -23,9 +23,8 @@ module.exports = class JGroupData extends Model
 
     sharedMethods :
       static      :
-        fetchByKey: [
-          (signature String, Function)
-        ]
+        fetchByKey: (signature String, Function)
+
 
   @create = (slug, callback) ->
     data = new JGroupData { slug }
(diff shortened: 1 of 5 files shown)


szkl (approved):

workers/social/lib/social/models/group/groupdata.coffee
```diff
@@ -77,22 +77,25 @@ module.exports = class JGroupData extends Model
...
-      key = "payload.#{item}"
-      val = getAt data, item
-      if val
-      then operation.$set[key]   = val
-      else operation.$unset[key] = ''
+    operation = {}
+
+    allowedPaths.forEach (path) ->
+      key = "payload.#{path}"
+      if val = data[path]
```
szkl: Line 92 had a call to `getAt` function to assign return value to `val` variable. Is it unnecessary?

szkl:
```
  1) workers.social.group.index #destroy() create a team should add resources, invitations, apiToken, groupData to team and add slack integration to admin account:
     Uncaught AssertionError: expected null to exist
      at workers/social/lib/social/models/group/index.test.coffee:303:29
      at workers/social/lib/social/models/group/groupdata.coffee:104:9
      at readableAddChunk (_stream_readable.js:176:18)
      at Socket.Readable.push (_stream_readable.js:134:10)
      at TCP.onread (net.js:548:20)
```

workers/social/lib/social/models/group/groupdata.coffee
gokmen: yep, it was using `getAt` for old implementation, but for now it's checking for existing keys on a given object which doesn't require `getAt` at all.

PR #11008, opened 2017-04-18
kd/template: print defaults for user input

rjeczalik:
This PR improve printing of `userInput_*` variables, so the following:

```
$ kd template init
Provider type []: marathon
Set "image" to []: ubuntu:16.04
Set "cpus" to []: 2
Set "mem" to []: 1024

Template successfully written to kd.yaml.
```

becomes:

```
$ kd template init
Do you want to overwrite "kd.yaml" file? [y/N]: y

Provider type []: marathon
Set "image" to [ubuntu:14.04]:
Set "cpus" to [1]:
Set "mem" to [256]:

Template successfully written to kd.yaml.
```

### DIFF
--- a/go/src/koding/klientctl/template.go
+++ b/go/src/koding/klientctl/template.go
@@ -151,6 +151,22 @@ func TemplateDelete(c *cli.Context, log logging.Logger, _ string) (int, error) {
 }
 
 func TemplateInit(c *cli.Context, log logging.Logger, _ string) (int, error) {
+	output := c.String("output")
+
+	if _, err := os.Stat(output); err == nil {
+		yn, err := helper.Ask("Do you want to overwrite %q file? [y/N]: ", output)
+		if err != nil {
+			return 1, err
+		}
+
+		switch strings.ToLower(yn) {
+		case "yes", "y":
+			fmt.Fprintln(os.Stderr)
+		default:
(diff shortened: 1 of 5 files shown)


go/src/koding/klientctl/template.go
```diff
@@ -151,6 +151,22 @@ func TemplateDelete(c *cli.Context, log logging.Logger, _ string) (int, error) {
...
+
+	if _, err := os.Stat(output); err == nil {
+		yn, err := helper.Ask("Do you want to overwrite %q file? [y/N]: ", output)
+		if err != nil {
+			return 1, err
+		}
+
+		switch strings.ToLower(yn) {
+		case "yes", "y":
+			fmt.Fprintln(os.Stderr)
```
ppknap: Is there any reason why we print empty line to stderr and not stdout?

ppknap (approved):

go/src/koding/klientctl/template.go
rjeczalik: No reason, why?

ppknap: because `helper.Ask` prints to stdout and then we print to stderr even when we have no errors. This may be confusing when someone eg. redirects only stderr.

PR #11014, opened 2017-04-20
kd: add init command

rjeczalik:
This PR adds `kd init` command, which is responsible for creating and building a stack interactively.

Example usage:

```
$ kd init
Initializing with new kd.yaml template file...

Provider type []: aws   
Set "instance_type" to [t2.nano]:

Template successfully written to kd.yaml.
Creating new credential for "Aws" provider...

Title [rjeczalik Thu Apr 20 10:21:23 2017]: new-aws-cred
Access Key ID [***]:
Secret Access Key [***]:
Region [us-east-1]: eu-west-1
Creating credential...
Created "new-aws-cred" credential with 58f86f95877dfc7cd3e4f8f1 identifier.
Create new stack...

Stack name [Gastly Aws Stack]: New AWS Stack

Creatad "New AWS Stack" stack with 58f86fb0bccfca387d85cc81 ID.
Waiting for the stack to finish building...

[30%] Fetching and validating credentials
[55%] Building stack resources
[70%] Checking VM connections
[100%] apply finished
```
```
$ kd init
error executing "init" command: Project already initialized with "New AWS Stack" stack (58f86fb0bccfca387d85cc81)
```

PR #11015, opened 2017-04-20
kd/machine: add start / stop commands

rjeczalik:
deps: ~#11014~


go/src/koding/klientctl/main.go
```diff
@@ -548,9 +548,30 @@ func run(args []string) {
 			}, {
 				Name:            "cp",
 				Description:     cmdDescriptions["cp"],
+				Usage:           "Copies a local file to a remote filesystem.",
```
ppknap: Actually, communication can be in both directions.

ppknap (approved):

PR #11024, opened 2017-04-21
Custom Permissions support

gokmen:
This PR introduces custom permission support into existing Bongo models over JPermissionSet with custom validators. Existing implementation of "customization" feature was only checking specific flags defined in client-side on client-side. There were custom flags defined under `JGroup.customize.*` which was not taken into account for the backend calls. This was causing issues with `memberCanCreateStack` customization which was only checking in the client-side and were allowing api calls.

Along with that, this PR also introduces read-only stack support to NSE and clone improvements on existing Stack Editor.

### DIFF
--- a/client/new-stack-editor/lib/views/index.coffee
+++ b/client/new-stack-editor/lib/views/index.coffee
@@ -206,6 +206,20 @@ module.exports = class StackEditor extends kd.View
       @toolbar.actionButton.hideLoader()
 
 
+  setReadOnly: (readonly = true) ->
+
+    debug 'setReadOnly', readonly
+
+    if @readonly = readonly
+      @setClass 'readonly'
+      @sideView.hide()
+    else
+      @unsetClass 'readonly'
+
+    for view in EDITORS
+      @[view].setReadOnly @readonly
+
+
   _loadSnapshot: (id) ->
 
     return no  unless id
--- a/workers/social/lib/social/models/group/permissionset.test.coffee
+++ b/workers/social/lib/social/models/group/permissionset.test.coffee
@@ -1,0 +1,119 @@
+{
+  expect
+  withDummyClient
+  withConvertedUser
+  checkBongoConnectivity
+} = require '../../../../testhelper'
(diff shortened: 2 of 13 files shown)


workers/social/lib/social/models/group/permissionset.test.coffee
```diff
@@ -0,0 +1,119 @@
...
+
+        it 'should allow for guests when validator returned yes', (done) ->
+          withConvertedUser ({ client }) ->
+            (JPermissionSet.permit {
+              advanced : [
+                { permission: 'open group', validateWith: (..., cb) -> cb null, yes }
+              ]
+              success  : -> done()
+            }) client
+
```
usirin: We might need to call `done` for the cases where we think it's gonna fail for each test, because, as far as i can see, it's gonna hang if the test isn't passing and gonna cause timeout. Not sure if it's intended but i guess if we call done with `err` it should fail early without waiting a timeout.

client/new-stack-editor/lib/views/index.coffee
```diff
@@ -206,6 +206,20 @@ module.exports = class StackEditor extends kd.View
...
+    debug 'setReadOnly', readonly
+
+    if @readonly = readonly
+      @setClass 'readonly'
+      @sideView.hide()
+    else
+      @unsetClass 'readonly'
+
+    for view in EDITORS
+      @[view].setReadOnly @readonly
```
usirin: [minor] `this[view].setReadOnly @readonly`

usirin (commented):
Only minor stuff.

usirin (approved):

workers/social/lib/social/models/group/permissionset.test.coffee
gokmen: right, I'll add done with exceptions to the failure states as well.

gokmen: merging it for now, will update later on.

PR #11063, opened 2017-04-26
Fixes around Updated Stack Template widget on sidebar

usirin:
I created a new event called `StackRevisionUpdated` on `computeController` to make it happen **ONLY IF** `StackRevisionChecked` event occurs **after** receiving a `StackTemplateChanged` event. 

Since `StackTemplateChanged` event only happens after initial load on realtime, `sidebarController` can hook into this event, and set its state to make `UpdatedStackWidget`  visible.

All other approaches i tried resulted in writing a state transition code inside a component, which was pretty ugly.

### DIFF
--- a/client/app/lib/providers/computecontroller.coffee
+++ b/client/app/lib/providers/computecontroller.coffee
@@ -70,7 +70,15 @@ module.exports = class ComputeController extends KDController
 
       @on 'StackAdminMessageDeleted', @bound 'handleStackAdminMessageDeleted'
 
-      groupsController.on 'StackTemplateChanged', (event) => @checkGroupStacks event?.contents
+      groupsController.on 'StackTemplateChanged', (event) =>
+        return  unless templateId = event?.contents
+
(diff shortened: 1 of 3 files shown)


gokmen (approved):
👍

gokmen:
I just realized that this is not the correct approach since the `StackTemplateChanged` event doesn't mean a change happened in the template itself but in the group instead. I didn't get it in the first place when I read over the code but then I realized while testing this out, the `StackTemplateChanged` event is coming from `GroupsController` and it's ok with naming from that perspective. On the other what we are trying to achieve with this PR is completely different; it should show the popup "when the content of a stack template changed" right now it pops up every time we set a different stack template to team which is not correct.

So, I'm going to revert this one and we can have a session about it. Sorry for that I didn't get it in the first place before merging it. We should listen the instance changed event on `JStackTemplate` instance to show/hide the popup if there is a generated stack from that stack template.

PR #11084, opened 2017-04-27
machine/mount/fuse: filesystem improvements

ppknap:
This PR reworks readdir fuse method to seek directory position when requested and stored offsets mismatch. It also adds synchronisation filters for some git files that should not be sent to remote.

### DIFF
--- a/go/src/koding/klient/machine/index/index.go
+++ b/go/src/koding/klient/machine/index/index.go
@@ -64,6 +64,11 @@ func NewIndexFiles(root string) (*Index, error) {
 			return nil
 		}
 
+		// Don't include OSX trashes to scanned files.
+		if info.Name() == ".Trash" && info.Mode().IsDir() && runtime.GOOS == "darwin" {
+			return filepath.SkipDir
+		}
+
 		fC <- &fileDesc{path: path, info: info}
 		return nil
 	}
@@ -197,6 +202,11 @@ func (idx *Index) MergeBranch(root, branch string) (cs ChangeSlice) {
 
 		if _, ok := visited[path]; ok {
 			return nil
+		}
+
+		// Don't include OSX trashes to scanned files.
+		if info.Name() == ".Trash" && info.Mode().IsDir() && runtime.GOOS == "darwin" {
+			return filepath.SkipDir
 		}
 
 		// File exists in local but not in remote.
--- a/go/src/koding/klient/machine/mount/notify/fuse/handle.go
+++ b/go/src/koding/klient/machine/mount/notify/fuse/handle.go
@@ -37,7 +37,7 @@ type FileHandle struct {
 	File    *os.File
 
 	size  int64
-	write int64
+	write bool
 }
 
 // Close closes stored file handle.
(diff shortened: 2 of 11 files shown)


go/src/koding/klient/machine/index/index.go
```diff
@@ -64,6 +64,11 @@ func NewIndexFiles(root string) (*Index, error) {
 			return nil
 		}
 
+		// Don't include OSX trashes to scanned files.
+		if info.Name() == ".Trash" && info.Mode().IsDir() && runtime.GOOS == "darwin" {
```
rjeczalik: Would it make sense to turn it into `isIgnored(fi os.FileInfo) bool` and extend for more ignored files, like vim's swap and temporary ones etc.

go/src/koding/klient/machine/mount/notify/fuse/handle.go
```diff
@@ -35,7 +36,8 @@ type FileHandle struct {
 	InodeID fuseops.InodeID
 	File    *os.File
 
-	size *int64
+	size  int64
+	write int64
```
rjeczalik: Looks like only 0/1 is ever assigned to write. What about bool instead?

go/src/koding/klient/machine/mount/notify/fuse/handle.go
```diff
@@ -173,44 +187,173 @@ func (fhg *FileHandleGroup) Close() error {
...
+	dh.mu.Unlock()
+
+	return
+}
+
+// ReadDir writes dirents do provided destination slice. It returns the number
+// of read bytes.
+func (dh *DirHandle) ReadDir(offset fuseops.DirOffset, dst []byte) (n int, err error) {
+	dh.mu.Lock()
+	if offset > fuseops.DirOffset(len(dh.stream)) {
```
rjeczalik: `if int(offset) > len(dh.stream) {` - a bit shorter

go/src/koding/klient/machine/mount/notify/fuse/handle.go
```diff
@@ -173,44 +187,173 @@ func (fhg *FileHandleGroup) Close() error {
...
+	ds = make([]fuseutil.Dirent, 0, n.ChildN()+2)
+
+	var offset fuseops.DirOffset = 1
+	n.Children(0, func(child *node.Node) {
+		if !child.Exist() {
+			return
+		}
+
+		ds = append(ds, fuseutil.Dirent{
+			Offset: offset,
```
rjeczalik: Can we use `fuseops.DirOffset(len(ds)) + 1` here instead?

go/src/koding/klient/machine/index/index.go
rjeczalik: Or e.g. read the list of ignored files from .gitignore

ppknap: @rjeczalik I added this as a workaround for failing tests. Will move mount/skipper.go logic to index which will be more generic solution. But didn't want to do this in this PR.

go/src/koding/klient/machine/mount/notify/fuse/handle.go
ppknap: It's leftover from the idea to have this atomic and without a mutex. Will change this to bool.

rjeczalik (approved):

rjeczalik:
Rebased & resolved conflict.

cihangir (approved):
i am mostly unfamiliar with the logic but as code quality LGTM

cihangir:
has merge conflicts

PR #11096, opened 2017-04-28
klientctl: allow to umount all machines at once

ppknap:
## Description
https://asciinema.org/a/47oocpuezxc2ax4euznfaqmqw

### DIFF
--- a/go/src/koding/klientctl/endpoint/machine/mount.go
+++ b/go/src/koding/klientctl/endpoint/machine/mount.go
@@ -6,14 +6,17 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"koding/klient/machine"
 	"koding/klient/machine/machinegroup"
 	"koding/klient/machine/mount"
 	"koding/klient/machine/mount/prefetch"
 	"koding/klient/machine/mount/sync/history"
+	"koding/klientctl/helper"
 
-	"github.com/dustin/go-humanize"
+	humanize "github.com/dustin/go-humanize"
 	"github.com/koding/logging"
 )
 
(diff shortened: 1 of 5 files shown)


go/src/koding/klientctl/endpoint/machine/mount.go
```diff
@@ -201,25 +206,80 @@ func (c *Client) Umount(options *UmountOptions) (err error) {
...
+	}
+
+	// Remove mounts.
+	for _, identifier := range identifiers {
+		umountReq := &machinegroup.UmountRequest{
+			Identifier: identifier,
+		}
+		var umountRes machinegroup.UmountResponse
+
+		if err := c.klient().Call("machine.umount", umountReq, &umountRes); err != nil {
```
rjeczalik: What about making it best-effort attempt at unmounting all mounts? So if single unmount fails we keep going unmounting test of them.

rjeczalik (approved):

go/src/koding/klientctl/endpoint/machine/mount.go
```diff
@@ -201,25 +206,80 @@ func (c *Client) Umount(options *UmountOptions) (err error) {
...
+		identifiers = mids
+	}
 
-	// Remove mount.
-	umountReq := &machinegroup.UmountRequest{
-		Identifier: options.Identifier,
+	// If there are no mounts we return nil.
+	if len(identifiers) == 0 {
+		fmt.Fprintf(os.Stdout, "There is nothing to unmount\n")
+		return nil
```
rjeczalik: I think it should return non-nil error when len(options.Identifiers) == 1, just to be backward-compatible with previous behaviour.

rjeczalik:
Rebased & resolved conflicts.

go/src/koding/klientctl/endpoint/machine/mount.go
ppknap: It is backward-compatibile, `len(options.Identifiers) == 1` is when user provides: "kd umount <mount>". The case above (`len(options.Identifiers) == 0`) covers the command "kd umount -a" when there is no mounts. Eg.  "kd umount --force" will result in error [here](https://github.com/koding/koding/pull/11096/files#diff-6c91a42efe84e0738dea5008055828c2R151).

cihangir (approved):

PR #11097, opened 2017-04-29
NSE: Preview support

gokmen:
This PR provides dynamic Preview support for Stack editor in NSE and also preview support for Readme pane.

Bonus: `JStackTemplate.preview` endpoint will provide a preview of provided stack template content.

### DIFF
--- a/client/app/lib/util/stacks/generatepreview.coffee
+++ b/client/app/lib/util/stacks/generatepreview.coffee
@@ -22,9 +22,9 @@ module.exports = generatePreview = (options = {}, callback) ->
           continue
 
         if content = jspath.getAt availableData[type], field
-          search   = if type is 'custom'  \
-            then ///\${var.#{type}_#{field}}///g
-            else ///\${var.koding_#{type}_#{field}}///g
+          search   = if type is 'custom'
+          then ///\${var\.#{type}_#{field}}///g
+          else ///\${var\.koding_#{type}_#{field}}///g
           template = template.replace search, content.replace /\n/g, '\\n'
         else
           errors[type] ?= []
--- a/client/app/lib/util/stacks/generatepreview.coffee
+++ b/client/app/lib/util/stacks/generatepreview.coffee
@@ -1,0 +1,41 @@
+jspath = require 'jspath'
+requirementsParser = require './requirementsparser'
+
+module.exports = generatePreview = (options = {}, callback) ->
+
+  { account, group, template, custom = {} } = options
+
+  availableData = { group, account, custom }
+  requiredData  = requirementsParser template
+  errors        = {}
+  warnings      = {}
+
+  processTemplate = ->
+
+    for type, data of requiredData
+
+      for field in data
+
+        if type is 'userInput'
+          warnings.userInput ?= []
+          warnings.userInput.push field
+          continue
+
+        if content = jspath.getAt availableData[type], field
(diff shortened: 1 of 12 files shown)


client/app/lib/util/stacks/generatepreview.coffee
```diff
@@ -0,0 +1,41 @@
...
+
+      for field in data
+
+        if type is 'userInput'
+          warnings.userInput ?= []
+          warnings.userInput.push field
+          continue
+
+        if content = jspath.getAt availableData[type], field
+          search   = if type is 'custom'  \
```
sinan: no need for `\` and the indentation below

client/app/lib/util/stacks/generatepreview.coffee
```diff
@@ -0,0 +1,41 @@
...
+            then ///\${var.#{type}_#{field}}///g
```
sinan: `///\${var\.#{type}_#{field}}///g` escape the `.`

client/app/lib/util/stacks/generatepreview.coffee
```diff
@@ -0,0 +1,41 @@
...
+            else ///\${var.koding_#{type}_#{field}}///g
```
sinan: again the `\.`

client/new-stack-editor/lib/views/editor.coffee
```diff
@@ -15,23 +17,41 @@ module.exports = class Editor extends BaseView
...
     @aceView = new AceView {
       cssClass: 'editor'
       useStorage: no
       createBottomBar: no
     }, FSHelper.createFileInstance { path: "localfile:/#{filename}" }
 
+    @getPreview = preview  if preview
+
+
+  getContent: (as = 'is') ->
```
sinan: minor: i dont think the default param and the early return is necessary

website/swagger.json
```diff
@@ -5460,6 +5460,77 @@
...
+        "description": "returns preview stack template for given template and custom variables",
+        "responses": {
+          "200": {
+            "description": "stacktemplate preview in provided format",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "template": {
+                  "type": "string",
+                  "default": "Hello gokmen -- bar"
```
sinan: intentional? 😏

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
```diff
@@ -400,6 +402,55 @@ module.exports = class JStackTemplate extends Module
...
+  #   }
+  #
+  # @example return
+  #
+  #   {
+  #     "template": "Hello gokmen -- bar",
+  #     "errors": {},
+  #     "warnings": {}
+  #   }
+  #
```
sinan: block comment would be better for usage description, unless we do this on other places.

sinan (approved):
I have some nice to haves, only important one is the regex comment.

website/swagger.json
gokmen: :) yes, it's going to show an output with the logged in user, so I've tried with my own and pasted the output.

workers/social/lib/social/models/computeproviders/stacktemplate.coffee
gokmen: 👍

PR #11098, opened 2017-04-29
Fix swagger definition for custom return types

gokmen:


rjeczalik:
Looks like we haven't update generated Go client in a while. Tried to do this now, looks like it no longer generates client methods for `/JStackTemplate.create`, so build fails with:

```
# koding/kites/kloud/stack
src/koding/kites/kloud/stack/import.go:248: sb.api.JStackTemplate.JStackTemplateCreate undefined (type *j_stack_template.Client has no field or method JStackTemplateCreate)
src/koding/kites/kloud/stack/import.go:286: sb.api.JStackTemplate.JStackTemplateUpdate undefined (type *j_stack_template.Client has no field or method JStackTemplateUpdate)
src/koding/kites/kloud/stack/import.go:306: sb.api.JStackTemplate.JStackTemplateGenerateStack undefined (type *j_stack_template.Client has no field or method JStackTemplateGenerateStack)
```

I tried to update swagger from 0.8 to 0.10, and the new one is parsing method names starting *after* the dot and then it complains about duplicates afterwards:

```
$ go generate .
The swagger spec at "../../../../website/swagger.json" is invalid against swagger specification 2.0. see errors :
- "modify" is defined 3 times
- "clone" is defined 2 times
- "shareWith" is defined 2 times
- "build" is defined 2 times
- "search" is defined 2 times
- "byCode" is defined 2 times
- "bootstrap" is defined 2 times
- "fetch" is defined 2 times
- "leave" is defined 2 times
- "destroy" is defined 4 times
- "kick" is defined 2 times
- "add" is defined 2 times
- "someWithRelationship" is defined 2 times
- "byId" is defined 2 times
- "update" is defined 7 times
- "delete" is defined 6 times
- "ping" is defined 2 times
- "create" is defined 13 times
- "some" is defined 11 times
- "byRelevance" is defined 2 times
- "one" is defined 11 times
- "stop" is defined 2 times
- "setAccess" is defined 2 times
- "count" is defined 2 times
- "remove" is defined 8 times

remoteapi.go:52: running "swagger": exit status 1
```

For now I'm not going to update our Go client, and instead we should replace to use the same codegen that is generating the spec (koding-api-codegen).

gokmen:
@rjeczalik we can merge this as is, I'm aware about the duplicate definition issue on swagger.json which is also wrong but there is no way to do that in cleaner way. We can use the one generated by coding-api-codegen but we need testing on that as well since I just generated that for Go but never used or tried it. It works well for js client though.

rjeczalik (approved):

Issue #11099, opened 2017-04-29
kd: add bug command for extended troubleshooting


rjeczalik:
Make it easier to gather feedback / details from a remote vm.

Add internal `klient upload` command for uploading log files and add `kd bug` as a convenience command that wraps `klient upload` + `kd mount inspect` + `kd log` and creates a github issue body.

Ensure all relevant mount information (syncing / indexing) is written to a log file.

rjeczalik:
TODO

- [ ] klient: ad d internal upload command (temporarily, until kd + klient is merged)
- [ ] kd: add log upload subcommand
  - [ ] kd/daemon: ensure logs have proper permissions
  - [ ] kd/log: write kd bug metadata to ~/.config/bugs and upload it
- [ ] scripts: add script for fetching bug metadata, downloading parts and merging them

PR #11116, opened 2017-05-02
kd/log: add upload command

rjeczalik:
This PR adds `kd log upload` command which is a prerequisite for #11099.

It also makes Upload not fail if content was uploaded previously and there's no new content to upload. In such cases Upload returns previous URL instead.

It also fixes a bug when gzipped content was erroneously gzipped again.


go/src/koding/klientctl/endpoint/machine/upload.go
```diff
@@ -0,0 +1,62 @@
...
+		} else {
+			req.File = f.File
+		}
+
+		e := c.klient().Call("log.upload", req, &resp)
+		if e != nil {
+			f.Error = e.Error()
+			if !force {
+				return e
+			} else {
```
ppknap: Linters tend to warn about not needed `else` here.

ppknap (approved):

PR #11122, opened 2017-05-03
NSE: Readonly stacks flow implemented

gokmen:
http://recordit.co/jPbsqbxdW6

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -230,12 +230,16 @@ module.exports = utils = {
     kd.team              ?= {}
     { team }              = kd
     team[formName]        = formData
-    localStorage.teamData = JSON.stringify team
+
+    try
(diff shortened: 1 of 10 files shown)


client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -230,12 +230,16 @@ module.exports = utils = {
     kd.team              ?= {}
     { team }              = kd
     team[formName]        = formData
-    localStorage.teamData = JSON.stringify team
+
+    try
+      localStorage.teamData = JSON.stringify team
```
sinan: you shouldn't need this while doing `stringify`

sinan (approved):

client/landing/site.landing/coffee/core/utils.coffee
gokmen: it's for `localStorage` access.

PR #11106, opened 2017-05-01
klientctl/metrics: add basic infra for storing metrics into bolt db

cihangir:

PR #11127, opened 2017-05-03
kloud/metrics: add handler for collecting metrics into kloud

cihangir:
* Add handler into kloud for getting messages from klients
* Add a new type for handling metrics properly in an optimized way for publishing to our system
* [][]byte comes from https://github.com/koding/koding/pull/11106/files#diff-8bf8a92c9f63f8aeeef76411693ae2c9R69


ppknap (approved):
👍

Issue #11125, opened 2017-05-03
machine/mount: prefetch fails if remote git repo is in use or in dirty state


rjeczalik:
```
$ kd mount aqua_squash:~/koding 
Mounting to /Users/rjeczalik/koding/mnt/development directory.
Checking remote path...
Mounted remote directory /home/rjeczalik/koding has 124854 file(s) of total size 2.4 GiB
Initializing mount /home/rjeczalik/koding -> /Users/rjeczalik/koding/mnt/development...
Using git files strategy to prefetch initial mount data.
Copying files: 100.0% (1535/1535), 324 MiB/324 MiB | 1.5 MiB/s
Cannot prefetch mount files: git command exited with error: exit status 128 ("fatal: Unable to create '/Users/rjeczalik/.cache/koding/mount-eda57448-332f-4bbc-8110-3539d49e477b/data/.git/index.lock': File exists.\n\nAnother git process seems to be running in this repository, e.g.\nan editor opened by 'git commit'. Please make sure all processes\nare terminated then try again. If it still fails, a git process\nmay have crashed in this repository earlier:\nremove the file manually to continue.\n")
Created mount with ID: eda57448-332f-4bbc-8110-3539d49e477b
```

Refuse to mount if remote repo is locked. Related #11124.

### Signature
```
2f726a65637a616c696b2f31336430303064382d663136612d343330312d613936332d3334626134636364336636342f6275672f646238386662323034656363323136646237313836343631643461323332316630643035313364622e6a736f6e2e677a2e30
```

ppknap:
Fixed by #11134 :

https://asciinema.org/a/1gri8ns3xpl3gzur2d6is7zmt

Issue #11124, opened 2017-05-03
machine/mount: prefetch failure renders mount unusable


rjeczalik:
```
$ kd mount green_coconut:~/koding
Mounting to /Users/rjeczalik/koding/mnt/development directory.
Checking remote path...
Mounted remote directory /home/rjeczalik/koding has 124853 file(s) of total size 2.4 GiB
Initializing mount /home/rjeczalik/koding -> /Users/rjeczalik/koding/mnt/development...

Using git files strategy to prefetch initial mount data.
Copying files: 100.0% (1534/1534), 324 MiB/324 MiB | 1.7 MiB/s
Cannot prefetch mount files: git command exited with error: exit status 128 ("error: packfile .git/objects/pack/pack-3291f55478f854d5e5117a4758a35c83978c26c1.pack does not match index\nerror: packfile .git/objects/pack/pack-3291f55478f854d5e5117a4758a35c83978c26c1.pack do
es not match index\nfatal: unable to read tree e90ddb32d5464c024514b08725f62407f7f0698d\n")
Created mount with ID: c455d936-7470-4f21-92a2-8965b132e293
```

```
$ cd ~/koding/mnt/development/
error: packfile .git/objects/pack/pack-3291f55478f854d5e5117a4758a35c83978c26c1.pack does not match index
error: packfile .git/objects/pack/pack-3291f55478f854d5e5117a4758a35c83978c26c1.pack does not match index
fatal: unable to read tree e90ddb32d5464c024514b08725f62407f7f0698d
$
```

To solve that make prefetch failure critical causing mount to fail. Cleanup the cache and mount directory afterwards. After the mount is proven to be able to recover from prefetch problems, make the prefetch failure a soft error again.

### Signature
```
2f726a65637a616c696b2f31336430303064382d663136612d343330312d613936332d3334626134636364336636342f6275672f346339376433623833383135646335303338336538323338353434656366326235346230663462342e6a736f6e2e677a2e30
```

rjeczalik:
I think this could have happened because the files / repo was in use on remote (koding service was started while the rsync copied the files over). Did not confirm yet, just a hint.

ppknap:
Interesting:

```
Copying files: 100.0% (1534/1534), 324 MiB/324 MiB | 1.7 MiB/s
```

as far as I remember, koding bare repository has around ~800-900MB. Maybe we start mounting(rsync scanned list of files) when repository was not downloaded yet. 

[...]

PR #11134, opened 2017-05-04
klient/mount: filter unwanted files from remote and local index

### LINKED ISSUE #11128: kd/mount: skipped local files are uploaded to remote machine

ppknap:
This PR fixes: #11128, #11125, #11124

## Description
When index is being created on remote machine and unwanted files are found, filter will fail with non-nil and non-SkipPath error. This will fail mounting on its first building stage with proper error describing what happened.

When mount syncer updates its index or fuse produces synchronisation events, they will be filtered and not sent to the event queue.

## How Has This Been Tested?
Manually, Unit tests.

## Types of changes


go/src/koding/klient/machine/index/cache.go
```diff
@@ -10,12 +10,27 @@ import (
...
 
 // TempIndexDirPrefix defines the prefix of temporary index file created by
 // Cached structure.
 const TempIndexDirPrefix = "koding_index_"
 
+// DefaultFilter defines a default filter used to process index caching.
+var DefaultFilter filter.Filter = filter.MultiFilter{
+	filter.OsSkip(filter.DirectorySkip(".Trash"), "darwin"),                                                                   // OSX trash directory.
+	filter.OsSkip(filter.DirectorySkip(".Trashes"), "darwin"),                                                                 // OSX trash directory.
+	filter.NewWithError(filter.PathSuffixSkip(".git/index.lock"), "git repository on remote is in use"),                       // git index lock file.
```
cihangir: i am not sure if we should have these kind of filters - we can break a git index and cause devs pain
ie: 
* rebase on remote 
* rebase fails
* index.lock stays
* we should not be able to continue

@szkl might comment on these internals better

cihangir: on the other hand, ideally, we should not care what the user might have in their mount

ppknap: @cihangir `filter.NewWithError(` means that if these files are present on remote, the local user who tries to mount such repository will get an error. In your case:
- rebase on remote
- rebase fails
- index.lock stays
- user tries to mount rebase-failed remote
- first mount step is to create cache index(what this logic does)
- filter detects `.git/index.lock` and since it's "WithError", creating index returns error
- nothing on remote happen.
- user who tried to mount will receive following error:
```
git repository on remote is in use (path: .git/index.lock)
```
- mounts will fail

go/src/koding/klient/machine/mount/sync.go
```diff
@@ -23,6 +24,20 @@ import (
...
+var DefaultFilter filter.Filter = filter.MultiFilter{
+	filter.OsSkip(filter.DirectorySkip(".Trash"), "darwin"),   // OSX trash directory.
+	filter.OsSkip(filter.DirectorySkip(".Trashes"), "darwin"), // OSX trash directory.
+	filter.PathSuffixSkip(".git/index.lock"),                  // git index lock file.
+	filter.PathSuffixSkip(".git/refs/stash.lock"),             // git stash lock file.
+	filter.PathSuffixSkip(".git/HEAD.lock"),                   // git HEAD lock.
+	filter.PathSuffixSkip(".git/ORIG_HEAD.lock"),              // git ORIG_HEAD lock.
+	filter.NewRegexSkip(`\.git/refs/heads/[^\s]+\.lock$`),     // git branch lock.
+	filter.NewRegexSkip(`\.git/index\.stash\.\d+\.lock$`),     // git stash ref. lock.
+	filter.NewRegexSkip(`\.git/objects/pack/tmp_pack_[^/]+`),  // temporary git files.
```
rjeczalik: Can we add also `. fseventsd/.*`?

https://github.com/koding/koding/issues/11137

rjeczalik (approved):

Issue #11135, opened 2017-05-04
kd/mount: syncing does not start after a successful mount


rjeczalik:
kd/mount: syncing does not start after a successful mount

## Steps to reproduce
[![asciicast](https://asciinema.org/a/aihe97ig13vm5bsg6h696uiov.png)](https://asciinema.org/a/aihe97ig13vm5bsg6h696uiov)

### Signature
```
2f726a65637a616c696b2f31336430303064382d663136612d343330312d613936332d3334626134636364336636342f6275672f363838356366383439393366383033333137383766616162616637353363663461383731623861652e6a736f6e2e677a2e30
```

rjeczalik:
Afterwards the sync deleted files which were not in the working three (`node_modules/` etc.).

rjeczalik:
Using the `git` strategy all the files were queued for syncing (106111), after 1h there were still around 80k to sync.

Using the `all` strategy files were synced in 13min with no syncing needed - https://asciinema.org/a/737kjps13y3llmepvkvfgnfyk.

ppknap:
Done by #11141

PR #11136, opened 2017-05-04
prefetch: disable git optimization

rjeczalik:
Related to #11135 and rationale https://github.com/koding/koding/issues/11135#issuecomment-299124671.

### DIFF
--- a/go/src/koding/klient/machine/mount/prefetch/strategy.go
+++ b/go/src/koding/klient/machine/mount/prefetch/strategy.go
@@ -9,7 +9,8 @@ import (
 
 // DefaultStrategy defines a default set of prefetchers.
 var DefaultStrategy = Strategy{
-	"git": Git{},
+	// TODO(rjeczalik): disabled due to #11135
+	// "git": Git{},
 	"all": All{},
 }
 


ppknap:
I think this can be done in smarter way: 
- the number of stored files in git can be obtaining from `.git/index` size file. Lets call this `Ngidx`,
- the total files in .git repository: `Ngit`
- all files: `Nall`

if `(Nall - Ngit - Ngidx)/Nall << 1` then `.git` index optimisation won't produce many sync events. Elsewhere [11135 comment](https://github.com/koding/koding/issues/11135#issuecomment-299124671) applies.

Lets merge this anyway since the main goal is to make mounts stable.

ppknap (approved):

rjeczalik:
> (Nall - Ngit - Ngidx)/Nall << 1

Yes, if there're no untracked files then there should be virtually no sync events produced, that's clear - no need to complicate it. However basing purely on observation, I think that's not how it actually worked.

The numbers were:

- Nall=110k
- Ngidx+Ngit=33k

And after `sudo kd restart` it started to sync 100k files, not ~70k. Even if it started to sync only 70k files, then still it'd be slower than `all` strategy with reasonable speedy internet connection.

ppknap:
It could added more files because of "disappearing files" bug we were talking last day. There is/was also FUSE SetAttributes problem that triggered update events for each Ngit files. These two could easily add remaining 30k events.

> no need to complicate it

This is actually the simplest way to determinate the number of untracked files from provided data.

PR #11146, opened 2017-05-08
kites/config: add more configuration settings for mounts

rjeczalik:
This PR renames configuration from local to mount. The purpose
is to enable for mount configuration settings and not have
two places where mounts are configured.

So the following:

    $ kd config set local.mountHome ~/mount

Changes to:

    $ kd config set mount.home ~/mount

And the following:

    $ kd config set local.mounts.default ~/mount

Changes to:

    $ kd config set mount.exports.default ~/mount

It also makes it possible to change the default history length
produced by mount inspect command.

It can be overridden by the following command:

    $ kd config set mount.inspect.history 500

WARN: Do not set the history above 1e7 - the history uses container/ring
as its underlying implementation, and ring.New(1e7) can take
around 10s to complete, ring.New(1e8) takes more than 60s, which
makes the mount command to time out.

Also it is possible to change the default number of
synchronization goroutines, which are started for each
mount.

It can be overridden by the following command:

    $ kd config set mount.sync.workers 100

Last but not least, it is also possible to change the debug level
used in mount subsystem.

Currently available debug levels:

    <=0  - turns off debug logging
    1    - turns on debug logging for syncer events
    2-8  - reserved for future use
    >=9  - turns on debug logging for fuse events

It can be overridden by the following command:

    $ kd config set mount.debug 1

### DIFF
--- a/go/src/koding/kites/config/konfig.go
+++ b/go/src/koding/kites/config/konfig.go
@@ -69,35 +69,35 @@ func (e *Endpoints) Social() *Endpoint {
 	return e.Koding.WithPath("/api/social")
 }
 
-// Local describes configuration of local paths.
-type Local struct {
+// Mount describes configuration of mounts.
+type Mount struct {
 	// Mount is a default home path of mounted directories.
 	//
 	// If empty, defaults to ~/koding/mnt/.
-	MountHome string `json:"mountHome,omitempty"`
+	Home string `json:"home,omitempty"`
 
-	// Mounts maps named mounts to local paths.
+	// Exports maps named mounts to local paths.
 	//
-	// The Mounts["default"] export is used as
+	// The Exports["default"] export is used as
 	// a default one when caller does not specify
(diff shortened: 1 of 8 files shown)


go/src/koding/kites/config/konfig.go
```diff
@@ -69,35 +70,65 @@ func (e *Endpoints) Social() *Endpoint {
...
+
+	// Debug is a debug level used for logging within
+	// mounts.
+	//
+	//   <=0  - turns off debug logging
+	//   1    - turns on debug logging for syncer events
+	//   2-8  - reserved for future use
+	//   >=9  - turns on debug logging for fuse events
+	//
+	Debug int `json:"debug,omitempty"`
```
cihangir: in general, i am not in favor of using int for debug/verbose config. not objecting, just wanted to state :)

cihangir (approved):

PR #11150, opened 2017-05-09
NSE: Docs support

gokmen:
ps. this also sets new stack editor as default editor for stack templates. there is still option for using the old one.

### DIFF
--- a/client/new-stack-editor/lib/views/index.coffee
+++ b/client/new-stack-editor/lib/views/index.coffee
@@ -146,36 +146,38 @@ module.exports = class StackEditor extends kd.View
 
     { router, computeController } = kd.singletons
 
-    switch event
-      when Events.Menu.Logs
+    actions =
+      'Menu.Logs': =>
         @logs.unsetClass 'shake'
         unless @logs.isClosed()
           kd.utils.defer => @logs.setClass 'shake'
         @logs.resize { percentage: 40, store: yes }
-      when Events.Menu.Credentials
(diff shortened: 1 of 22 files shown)


client/new-stack-editor/lib/views/index.coffee
```diff
@@ -156,13 +158,14 @@ module.exports = class StackEditor extends kd.View
       when Events.Menu.MakeTeamDefault
         computeController.makeTeamDefault { template: @getData() }
       when Events.Menu.Clone
-        @toolbar.setBanner {
+        @toolbar.setBanner
           message  : 'Clonning...'
```
sinan: `Cloning…`

client/new-stack-editor/lib/views/sideview.coffee
```diff
@@ -50,17 +52,43 @@ module.exports = class SideView extends BaseView
...
+        else
+          @title.hide()
         @setClass item.cssClass
         item.view.show()
 
       else
         @unsetClass item.cssClass
         item.view.hide()
 
+    { expanded, collapsed = yes } = options
```
sinan: show as is (with the prev state) `unless options`
show expanded `if options.expanded`
show collapsed `unless options.expanded`

sinan (changes requested):
👍

gokmen:
@sinan all addressed and updated.

sinan (approved):

PR #11163, opened 2017-05-10
util: use email-validator pkg for email validation

### LINKED ISSUE #11019: The invitation form doesn't validate correctly with some email addresses.
## Current Behavior
image
- works fine with cartahealthcare.com
- working as expected when the same email address is imported as a CSV.

usirin:
fixes #11019


gokmen (approved):
there are some interesting issues with that package https://github.com/Sembiance/email-validator/issues but probably it's still a better one.

PR #11167, opened 2017-05-11
klient/machine: do not rebuild valid client more than once

ppknap:
Fixes: https://circleci.com/gh/koding/koding/2353

## Description
When a context is closed, supervisor logic should rebuild client. Tests are to check if rebuild was triggered. However, there is also test which checks if `Execker` is not executed, it won't be if build returns an error after the context close.

## Motivation and Context
Solve random test failures.

## How Has This Been Tested?
n/a


go/src/koding/klient/machine/mount/sync/supervised/sypervised_test.go
```diff
@@ -2,6 +2,7 @@ package supervised_test
```
rjeczalik: just noticed, supervised\_test.go

cihangir (approved):

go/src/koding/klient/machine/mount/sync/supervised/sypervised_test.go
ppknap: nice catch, it was here for months:O

PR #11169, opened 2017-05-11
Fix printconfig command

szkl:

### DIFF
--- a/config/generateRunFile.coffee
+++ b/config/generateRunFile.coffee
@@ -613,7 +613,8 @@ generateDev = (KONFIG, options) ->
 
     elif [ "$1" == "printconfig" ]; then
 
-      printconfig $@
+      shift
+      echo ${!1}
 
     elif [[ "$1" == "log" || "$1" == "logs" ]]; then
 


config/generateRunFile.coffee
```diff
@@ -613,7 +613,8 @@ generateDev = (KONFIG, options) ->
 
     elif [ "$1" == "printconfig" ]; then
 
-      printconfig $@
+      shift
+      echo ${!1}
```
cihangir: what does `${!1}` do? could not understand :/

szkl: This is called indirect expansion. It expands the variable named by expanding `PREFIX` variable first.

```
'${!PREFIX*}'
'${!PREFIX@}'
     Expands to the names of variables whose names begin with PREFIX,
     separated by the first character of the 'IFS' special variable.
     When '@' is used and the expansion appears within double quotes,
     each variable name expands to a separate word.
```

```
./run printconfig KONFIG_FOO
> 1=KONFIG_FOO
> ${!1}
> ${KONFIG_FOO}
```

cihangir: thx sir!

cihangir (approved):

PR #11172, opened 2017-05-11
object: fix nil check for flat stringers

rjeczalik:
When peeking for fmt.Stringer interface upgrade
the code assumed the underlying type can be nil.

This caused a panic, when trying to flat-string
a time.Duration.

### DIFF
--- a/go/src/koding/kites/kloud/utils/object/object.go
+++ b/go/src/koding/kites/kloud/utils/object/object.go
@@ -227,6 +227,13 @@ func (b *Builder) buildMapObject(v interface{}, obj Object, ignored ...string) {
 			continue
 		}
 
+		if s, ok := vv.(fmt.Stringer); ok && b.FlatStringers {
+			if !isNil(s) {
+				b.set(obj, key, s.String(), ignored...)
+			}
+			continue
+		}
+
 		child := flatten(vv)
 		if child == nil {
 			b.set(obj, key, vv, ignored...)
@@ -254,7 +261,7 @@ func (b *Builder) buildStruct(v interface{}, obj Object, ignored ...string) {
 		}
 
 		if s, ok := field.Value().(fmt.Stringer); ok && b.FlatStringers {
-			if !reflect.ValueOf(s).IsNil() {
+			if !isNil(s) {
 				b.set(obj, key, s.String(), ignored...)
 			}
 			continue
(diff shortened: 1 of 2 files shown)


go/src/koding/kites/kloud/utils/object/object.go
```diff
@@ -322,6 +329,15 @@ func isMapObject(v interface{}) bool {
 	return typ.Key().Kind() == reflect.String
 }
 
+func isNil(v interface{}) bool {
+	switch v := reflect.ValueOf(v); v.Type().Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Slice:
```
cihangir: TIL:

this works https://play.golang.org/p/q6NZKo5KUo

but 

this doesnt https://play.golang.org/p/zdOCtq4x9g

cihangir: not sure why tho

cihangir (approved):

rjeczalik:
The are two problems with https://play.golang.org/p/zdOCtq4x9g:

- you can't define method on an interface - Any.String()
- you can't initialise interface with literal (&Any{}), eventually you may
do "var a Any; &a"

cihangir:
https://github.com/koding/koding/pull/11172#issuecomment-300792547

My intention was to show, you cant define a method to an interface, but you can pass that interface to a function basing on this [here](https://golang.org/ref/spec#Method_declarations)

"The type of a method is the type of a function with the receiver as first argument."

My confusion comes after reading this 

"The type denoted by T is called the receiver base type; it must not be a pointer or interface type and it must be declared in the same package as the method. " 

If method receiver is the first argument to a function, why cant i define a method to an interface?

And all came from your switch statement - which didnt include reflect.Interface

rjeczalik:
> why cant i define a method to an interface?

I think this due to interface definition:

InterfaceType = "interface" "{" { MethodSpec ";" } "}" .


https://golang.org/ref/spec#Interface_types

Method set for an interface is defined within the interface{ ... } block
for an interface type, for non-interface types the method set is defined
outside type definition.

rjeczalik:
Also https://golang.org/ref/spec#Method_sets:

A type may have a method set associated with it. The method set of an
interface type is its interface. The method set of any other type T
consists of all methods declared with receiver type T.

PR #11177, opened 2017-05-13
Run user proxy in default configuration

szkl:

### DIFF
--- a/client/app/lib/util/proxifier.coffee
+++ b/client/app/lib/util/proxifier.coffee
@@ -35,11 +35,7 @@ module.exports = class Proxifier
     subdomain = if isInProduction then 'p' else 'dev-p'
 
     # create the base url
-    # Since proxy is a service from koding.com we need to use it
-    # for on-premise version as well. otherwise all the on-premises
-    # users needs to run proxy server on their own ~GG
-    baseURL = "#{protocol}//#{subdomain}.koding.com/-"
-    # baseURL = "#{protocol}//#{subdomain}.#{globals.config.domains.base}/-"
+    baseURL = "#{protocol}//#{subdomain}.#{globals.config.domains.base}/-"
 
     # if it's a tunnel given domain we need to do one more check
     # for tunnels since production tunnel proxy is different
--- a/config/workers.coffee
+++ b/config/workers.coffee
@@ -318,7 +318,7 @@ module.exports = (KONFIG, options, credentials) ->
         ]
 
     userproxies         :
-      group             : 'proxy'
+      group             : 'webserver'
       nginx             :
         websocket       : yes
         locations       : [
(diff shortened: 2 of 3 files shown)


rjeczalik (approved):

config/workers.coffee
```diff
@@ -318,7 +318,7 @@ module.exports = (KONFIG, options, credentials) ->
         ]
 
     userproxies         :
-      group             : 'proxy'
+      group             : if KONFIG.environment is 'default' then 'default' else 'proxy'
```
cihangir: you can add default into https://github.com/koding/koding/blob/master/deployment/grouptoenvmapping.coffee#L1 instead of this change

szkl: I wanted to avoid tunnel workers in the default configuration. They depend on Route53 right now. I thought they would break the health check. I can try again, though.

cihangir: Hm i forgot about tunnel workers - lets have this change then we can take a look later

cihangir (approved):

config/workers.coffee
szkl: Cool.

client/app/lib/util/proxifier.coffee
```diff
@@ -35,11 +35,7 @@ module.exports = class Proxifier
     subdomain = if isInProduction then 'p' else 'dev-p'
 
     # create the base url
-    # Since proxy is a service from koding.com we need to use it
-    # for on-premise version as well. otherwise all the on-premises
-    # users needs to run proxy server on their own ~GG
-    baseURL = "#{protocol}//#{subdomain}.koding.com/-"
-    # baseURL = "#{protocol}//#{subdomain}.#{globals.config.domains.base}/-"
+    baseURL = "#{protocol}//#{subdomain}.#{globals.config.domains.base}/-"
```
gokmen: if we are not going to run user proxy on dev env this may break it since they are going to provide some other domain here which will end up using managed vms in dev environments over tunnel.

szkl: I've forgotten about development environment.

gokmen: @cihangir suggested to running user proxy in dev environment as well, which solves this issue as well, in theory.

szkl: I agree. I'll update the pull request today.

szkl: I've split user proxy and tunnel proxy locations into separate groups. User proxy locations are in the web server group now. This is a much easier change to read in the `workers` module.

I've tested current changes in default and development environments.

gokmen (approved):

client/app/lib/util/proxifier.coffee
cihangir: Awesome solution, thanks!

cihangir (approved):

PR #11186, opened 2017-05-16
kd/mount/fuse: add symlink support

ppknap:
This PR enables symlink support in FUSE based mounts.

## Description
https://asciinema.org/a/7u4si3bv4v3cc4otigzpvkyhp

## How Has This Been Tested?
Manually

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/mount/notify/fuse/fuse.go
+++ b/go/src/koding/klient/machine/mount/notify/fuse/fuse.go
@@ -472,8 +472,84 @@ func (fs *Filesystem) ReleaseDirHandle(_ context.Context, op *fuseops.ReleaseDir
 	return err
 }
 
+// CreateSymlink creates a symlink to a given target.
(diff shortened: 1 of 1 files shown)


rjeczalik (approved):

cihangir:
I have one question that couldnt find an answer - I couldnt see in the gif and also there are not tests. do we support symlinks that are out of the mounted folder? As an example from your gif:
```
root_symlink -> /  
```

when i do `cd ./root_symlink` will it work properly?

cihangir:
Needs rebase or push an empty commit - test will fail due to a know bug where commit id is a NaN in JS.

ppknap:
> when i do cd ./root_symlink will it work properly?

Yes, it will.
[pic](https://asciinema.org/a/c5yid9ek49tmys4gv1tj4yhl3)

Initially, I was thinking about stripping/changing absolute symlink targets for mount<->local directories but this would break git(it would detect this as a file change) so I leave symlinks without target changes.

PR #11187, opened 2017-05-16
kites/config: don't alias default env

rjeczalik:


szkl (approved):
Build [fails](https://circleci.com/gh/koding/koding/2447).

PR #11194, opened 2017-05-16
go/scripts: update elb id

cihangir:


cihangir:
I've cancelled the build as it is unnecessary.

szkl (approved):

PR #11200, opened 2017-05-17
kloud/metadata: make it possible to merge user items

rjeczalik:
This PR makes it possible to modify the koding user,
that is running the Klient service on a remote vm.

In particular the following user_data will currently
make the template parsing to fail:

    user_data: |
      #cloud-config
      users:
      - name: ${var.koding_account_profile_nickname}
        groups:
        - docker

This PR will merge the koding user object with the
default one, defined by kloud. In effect, the resulting
user will be added to a docker group as well.

This technique can be also used e.g. to add an ssh key
for the koding user, like:

    user_data: |
      #cloud-config
      users:
      - name: ${var.koding_account_profile_nickname}
        ssh-authorized-keys:
        - ssh-rsa AAAAB3NzaC1yc2EA... koding-350298856


go/src/koding/kites/kloud/metadata/cloudinit.go
```diff
@@ -145,6 +154,34 @@ func mergeList(in, out []interface{}, path []string) ([]interface{}, error) {
...
+						continue
+					}
+
+					if err := merge(shallowCopy(obj, key), orig); err != nil {
+						return nil, err
+					}
+
+					out[i] = orig
+
+					continue loop
```
cihangir: is there any way to rewrite this without "continue loop" making really hard to read & understand the flow?

rjeczalik: Done.

gokmen (approved):

cihangir (approved):

Issue #11201, opened 2017-05-17
kd: panic


rjeczalik:
```
$ kd mount ls
metrics wont be collected:  timeout
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1ced75d]

goroutine 1 [running]:
main.run(0xc420010480, 0x3, 0x3)
        /Users/rjeczalik/src/github.com/koding/koding/go/src/koding/klientctl/main.go:1011 +0x8d7d
main.main()
        /Users/rjeczalik/src/github.com/koding/koding/go/src/koding/klientctl/main.go:70 +0x45
```

rjeczalik:
```
$ time kd ls
metrics wont be collected:  timeout
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1ced75d]

goroutine 1 [running]:
main.run(0xc42000c320, 0x2, 0x2)
        /Users/rjeczalik/src/github.com/koding/koding/go/src/koding/klientctl/main.go:1011 +0x8d7d
main.main()
        /Users/rjeczalik/src/github.com/koding/koding/go/src/koding/klientctl/main.go:70 +0x45

real    0m5.085s
user    0m0.063s
sys     0m0.027s
(rjeczalik-osxpro.local) {master} ruby-rails-sample $

**Comment by rjeczalik** [comment]
I think I know what's the problem - there's another kd process running:

```
$ ps ax | grep kd
65027 s002  S+     0:00.10 kd ssh purple_watermelon
65029 s002  S+     0:00.15 ssh -i /Users/rjeczalik/.ssh/twoja-stara -o StrictHostKeychecking=no -o UserKnownHostsFile=/dev/null -o ServerAliveInterval=300 -o ServerAliveCountMax=3 -o ConnectTimeout=7 -o ConnectionAttempts=1 rjeczalik@99.999.999.999
```

So starting another kd process times out on trying to get access to metrics.bolt. Probably it should be more fine-grained. Related https://github.com/koding/koding/pull/11159.

**Comment by rjeczalik** [comment]
Also it would be useful if there's an option to disable metrics (e.g. if someone wants to disable them in development environment), like:

```
$ kd config set noMetrics true
$ sudo kd restart
```

rjeczalik:
Already fixed.

PR #11202, opened 2017-05-17
kites/config: add disableMetrics setting

rjeczalik:
Allows for disabling metrics with:

    $ kd config set disableMetrics true

Works around #11201.

### DIFF
--- a/go/src/koding/kites/config/konfig.go
+++ b/go/src/koding/kites/config/konfig.go
@@ -169,8 +169,9 @@ type Konfig struct {
 	PublicBucketName   string `json:"publicBucketName,omitempty"`
 	PublicBucketRegion string `json:"publicBucketRegion,omitempty"`
 
-	LockTimeout time.Duration `json:"lockTimeout,omitempty"`
-	Debug       bool          `json:"debug,string,omitempty"`
+	LockTimeout    time.Duration `json:"lockTimeout,omitempty"`
+	DisableMetrics bool          `json:"disableMetrics,string,omitempty"`
+	Debug          bool          `json:"debug,string,omitempty"`
 }
 
 // KiteHome gives directory of the kite.key file.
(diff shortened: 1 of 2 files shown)


go/src/koding/kites/config/konfig.go
```diff
@@ -170,6 +170,7 @@ type Konfig struct {
 	PublicBucketRegion string `json:"publicBucketRegion,omitempty"`
 
 	LockTimeout time.Duration `json:"lockTimeout,omitempty"`
+	NoMetrics   bool          `json:"noMetrics,string,omitempty"`
```
cihangir: DisableMetrics sounds like a more fit.

ppknap: why just not:

```
$ kd config set metrics false
```

Generally, I always have problems when reading double negation (operator-name) like
`if !noSomething {`. 🤔

ppknap (approved):

go/src/koding/kites/config/konfig.go
rjeczalik: > why just not:

Typically in Go you try to make zero values useful - if you have a configuration struct, and default behaviour is that something is enabled, you need to resort to `NoSomething` or `DisableSomething` in order to make zero value meaningful.

ppknap: I'm not looking from Go developer perspective but rather from user one. `set noMetrics true` is less readable than `set metrics false`. Just a personal opinion. If you think this fits better, let it be.

rjeczalik: Yup, better idea, changed to `metrics false`. 👍

rjeczalik: Ok, now I remember why I did it that way - the configuration object merging relies on zero values - you can't have a default bool value which is not a zero value. There are two configuration values -  a default one and the one read from bolt. When they are merged into one, the effective field has the field's value of the object, which was non-zero.

That's why when default bool is true, during merging it will always be true. For now it's easier to stick with default values being zero values for bool types, instead of rewriting merging.

rjeczalik: Changed to `disableMetrics`.

cihangir (approved):
there is one more thing that we should implement in order to disable metrics but i will do it after merging this

PR #11204, opened 2017-05-18
kd/mount: prevent from syncing two changes of the same file concurrently

ppknap:
This PR solves synchronization races caused by more than one event producer. Eg:

 - FUSE produces edit change `C1` on file `F`
 - `C1` is added to anteroom and sent to worker `X`
 - Worker `X` runs rsync on `C1`
 - FUSE produces delete change `C2` on file `F`
 - `C2` is added to anteroom
 - Anteroom deprecates `C1` however, `C1` is already being processed
 - `C2` is sent to worker `Y`
 - Worker `Y` runs rsync on `C2`

With previous logic, we could not guess which change will be finished first.

## Motivation and Context
Remote changes support prerequirement.

## How Has This Been Tested?
Unit tests.

### DIFF
--- a/go/src/koding/klient/machine/mount/anteroom.go
+++ b/go/src/koding/klient/machine/mount/anteroom.go
@@ -21,10 +21,14 @@ type Anteroom struct {
 	closed bool          // set to true when the object was closed.
 	stopC  chan struct{} // channel used to close queue dispatching go-routine.
 
-	mu     sync.Mutex
-	evs    map[string]*msync.Event // Change name to change event map.
-	synced int64                   // How many events are processing.
-	idle   *subscribers            // Subscribers waiting for idle signals.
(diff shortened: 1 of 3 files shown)


go/src/koding/klient/machine/mount/anteroom.go
```diff
@@ -184,13 +189,54 @@ func (a *Anteroom) dequeue() {
 		wakeupTick = time.NewTicker(30 * time.Second)
 	)
 
+	// Pop event from the queue. If received event is present in current
+	// processed map, the event must be stopped until already sent one is
+	// executed. This prevents situations where two paraller workers executes
```
cihangir: s/paraller/parallel

cihangir (approved):

rjeczalik (approved):

PR #11212, opened 2017-05-18
Auto generate terraformer and kloud worker secret keys

szkl:

### DIFF
--- a/config/credentials.default.coffee
+++ b/config/credentials.default.coffee
@@ -1,3 +1,5 @@
+randomstring = require 'randomstring'
+
 module.exports = (options) ->
   kiteHome = '$KONFIG_PROJECTROOT/generated/kite_home/koding'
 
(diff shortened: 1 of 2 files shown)


gokmen (approved):
This change is ok for me but somehow while testing this I had to cleanup everything after each `./configure` otherwise it didn't work for me. Not sure if it's an issue on my end, just fyi.

szkl:
@gokmen I just tested and I had no issue after configure and restarting workers. I've only tested creating a new stack template with a new credential and building a stack out of it. What issue did you experience?

rjeczalik (approved):

gokmen:
@szkl it was failing to connect with the updated `secretKey` and it started to work again after cleanup and build go binaries. dunno maybe there was a running service that needs to be stopped and restarted again.

Issue #11068, opened 2017-04-26
kd/mount: use simpler identifiers for mounts.


ppknap:
Drop UUIDv4 and use name logic similar to UNIX `mount` command. Mount id is also used as cache directory identifier, so this needs to be taken into account in order to avoid cache collisions.

PR #11215, opened 2017-05-19
kd: add mount sync pause/resume options.

ppknap:
This PR adds `--pause` and `--resume` options to `kd mount sync` subcommand. `kd sync` was itself moved to `kd mount` namespace. Calling `kd mount sync` on paused mount is allowed.

[pic](https://asciinema.org/a/9bkl4vcwlfsjcqp9ku1fzpbzf)

## Motivation and Context
`kd live` requirement

## How Has This Been Tested?
Manually

## Types of changes


go/src/koding/klient/machine/mount/anteroom.go
```diff
@@ -302,6 +328,22 @@ func (a *Anteroom) Unsync(path string) {
 	a.cursMu.Unlock()
 }
 
+// Pause prevets Anteroom from sending new events.
+func (a *Anteroom) Pause() {
+	atomic.StoreInt64(&a.paused, 1)
```
rjeczalik: Can we also a.idle.done(false)? If we wait for kd sync and pause concurrently, the original kd will gang indefinitely.

go/src/koding/klientctl/main.go
```diff
@@ -743,6 +744,26 @@ error opening: %s
...
+				Description: cmdDescriptions["mount-sync"],
+				Action:      ctlcli.ExitErrAction(MachineSyncMount, log, "sync"),
+				Flags: []cli.Flag{
+					cli.DurationFlag{
+						Name:  "timeout, t",
+						Usage: "Maximum time to wait.",
+						Value: time.Minute,
+					},
+					cli.BoolFlag{
+						Name:  "pause",
```
rjeczalik: What about making them (pause + resume) subcommands?

go/src/koding/klientctl/main.go
```diff
@@ -863,16 +884,6 @@ error opening: %s
 			},
 		}},
 	}, {
-		Name:   "sync",
```
rjeczalik: What about aliasing kd sync to kd mount sync?

rjeczalik (commented):
Feedback on WaitIdle + Pause interoperability.

go/src/koding/klientctl/main.go
ppknap: Initially, my goal was to do like you wrote. However, there are two issues with this approach:

1) Subcommand `pause`/`resume` and directory `pause`/`resume` ambiguity:

If we have `./pause` directory, calling `kd mount sync pause` could indicate either we want to pause synchronization or trigger sync on `./pause` directory. This can be solved by some heuristics that takes into account that mounts cannot be nested but it's still confusing. 

2) Subcommand `pause`/`resume` and mount name ambiguity:

If #11068 is done, calling `kd mount sync pause` may be either pausing synchronization for mount in WD or pausing synchronization for mount with name `pause`.

Generally, I think that the subcommand syntax over option one will be very hard to understand for the potential users.

go/src/koding/klientctl/main.go
ppknap: 👍 , Although, I think that `sync` command is only specific to mounts and should not be put in the global namespace. It's easier to add it later if frequently requested, but will be much harder to remove without breaking changes if it causes users' confusion.

go/src/koding/klientctl/main.go
rjeczalik: > If we have ./pause directory, calling kd mount sync pause could indicate either we 

I think we can require that relative paths must begin with `./` or `../`, go tool works that way so there's no ambiguity if the argument is a package name or directory.

> If #11068 is done, calling kd mount sync pause may be either pausing synchronization for mount in WD or pausing synchronization for mount with name pause

This can be handled at mount name generator phase - add list of reserved mount names and generate one that does not exist in the list.

rjeczalik: I still think they should be subcommands and not flags. Can we pick up this issue elsewhere?

rjeczalik (approved):

PR #11218, opened 2017-05-20
Send metrics to countly

cihangir:


go/src/glide.lock
```diff
@@ -552,7 +552,7 @@ imports:
 - name: github.com/koding/ec2dynamicdata
   version: 56f69f128b37c6bdfd61a6cac4b4b87e517fa2ee
 - name: github.com/koding/eventexporter
-  version: 8c411a9ed989426b2aa5bbcc6c9515cb3c76caf5
+  version: 0c3422129727d2c751c8a57e9570321b8d01217a
```
ppknap: Could you please update glide lock file. Commit `8c411a9ed989426b2aa5bbcc6c9515cb3c76caf5` is master's head in `eventexporter` repository.

go/src/koding/kites/kloud/metrics/metrics.go
```diff
@@ -69,6 +73,12 @@ func (p *Publisher) Publish(r *kite.Request) (interface{}, error) {
 		return nil, err
 	}
 
+	go func() {
+		if _, err := http.Post(p.metricsEndpoint, "application/json", bytes.NewReader(r.Args.Raw)); err != nil {
```
rjeczalik: http.DefaultClient has no request timeout set and I think it should.

Also I think Publisher could have `Client *http.Client` field and default to http.DefaultClient if nil.

go/src/koding/kites/metrics/metrics.go
```diff
@@ -42,9 +42,14 @@ func NewWithPath(boltPath, app string) (*Metrics, error) {
 // NewWithDB creates a new metric collector with given bolt db instance. See New
 // func.
 func NewWithDB(db *bolt.DB, app string) (*Metrics, error) {
-	storage, err := NewBoltStorageWithDB(db)
+	var storage Storage
+	var err error
+	if db != nil {
+		storage, err = NewBoltStorageWithDB(db)
```
rjeczalik: This would make we would silently lose metrics.

If user `kd ssh some_machine`, the kd process is going to take a lock on metrics.bolt as long as the ssh session is alive.

If kd is executed concurrently (e.g. by atom or user in other terminal), it will fallback to in-memory store which would affectively drop any metrics created during its runtime.

Or I'm missing something?

go/src/koding/kites/kloud/metrics/metrics.go
cihangir: I have used default client since i wasnt handling error cases but as you said would be better to have a configured client in the case where the target is down/slow to respond

go/src/koding/kites/metrics/metrics.go
cihangir: We have that losing issue on [here](https://github.com/koding/koding/pull/11218#discussion_r118312089) as well. That is in my list to handle them gracefully.

rjeczalik (approved):

ppknap (approved):

Issue #11067, opened 2017-04-26
kd/machine: automatically set new address to machine when it changes its IP


ppknap:
Currently, we store machine IP addresses after `kd list` along with kite address. When machine is restarted its IP may change. This is detected in client and reconnection is performed. However, IP used in SSH calls is not updated. This causes logic which use ssh to fail.

## Expected Behavior
Klient uses correct IP addresses for machines which changes their IP.

## Current Behavior
Klient needs `kd list` in order to obtain new addresses.

## Possible Solution
Dynamic client already uses kite addresses and has detection logic for disconnected clients. Thus, It can use kite address after machine is online again to obtain its new IP and then store it to addresses structure.

PR #11233, opened 2017-05-24
kd/machine: automatically set new address to machine when it changes its IP

ppknap:
Resolves: #11067

Before:
[pic](https://asciinema.org/a/12yeusjtgqel6ou28i75ks70v)

After:
[pic](https://asciinema.org/a/e3k7h5iv5a5g9bb67zl40o3fy)

## How Has This Been Tested?
Manually

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/client/dynamic.go
+++ b/go/src/koding/klient/machine/client/dynamic.go
@@ -25,6 +25,12 @@ type Builder interface {
 	// Build builds new client which will connect to machine using provided
 	// address.
 	Build(ctx context.Context, addr machine.Addr) Client
+
+	// IP lookups for IP address to machine pointed by provided argument. If
+	// argument's network is already "ip", this method should be no-op. Non-nil
+	// errors should be returned when it's not possible to find IP address using
+	// given value.
+	IP(addr machine.Addr) (machine.Addr, error)
 }
 
 // DynamicOpts are the options used to configure dynamic client.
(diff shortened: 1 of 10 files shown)


go/src/koding/klient/machine/client/dynamic.go
```diff
@@ -146,7 +162,30 @@ func (dc *Dynamic) cron() {
...
+
+	// Lookup for IP address and set if changed.
+	setIP := func(a machine.Addr) {
+		// Do not lookup tunneled connections.
+		if _, err := dc.opts.AddrFunc("tunnel"); err == nil {
+			return
+		}
+
+		ipA, err := dc.opts.Builder.IP(a)
+		if err != nil || ipA.Network != "ip" {
```
rjeczalik: Can `ipA.Network` be either `ip4` or `ip6`?

rjeczalik (approved):

go/src/koding/klient/machine/client/dynamic.go
ppknap: it should be possible to handle IPv6 too. And this logic does that(although I have never seen IPv6 in kite and I don't know if it handles it). IPv6 support also requires `-6` flag to `ssh` command and square brackets in rsync (`username@[IPv6]:~/dir`). We may want to create a separate issue for IPv6 support since it was never supported by kd.

PR #11234, opened 2017-05-24
NotificationWorker for Realtime updates

gokmen:
## Description

This PR provides in house realtime notification support for all the existing events that we use with PubNub. 

## Motivation and Context

Currently we don't have realtime update support for `default` environment which is provided for on premise installs. To be able to have realtime support on on premise Koding you need to have a paid PubNub subscription but with this PR we're providing same feature set in house, by using existing stack that we have; rabbitmq, mongo and sockjs used for implementation.

Notification workers are designed to handle all the requirements to fill this gap and they are horizontally scalable for web socket usage. For long polling load balancer needs to be configure to have sticky session support. see https://github.com/koding/koding/commit/7bb93076049ca00773557eeb9f06d73f50124c85

Notification Workers does following;
 - [x] provides a REST api for drop-in-replacement with socialApi's existing dispatcher
 - [x] provides SockJS server for clients to connect for realtime updates
 - [x] consumes RabbitMQ for publishing and consuming realtime events
 - [x] on client side it has a simple sock-js client implementation with auto reconnect support with backoff

Can be shown as;

It provides user and team (group) notifications via different routes, to have same ability with existing channel implementation;

There is a simple status endpoint which shows the connection statuses that can be disabled from nginx configuration;

It can auto re-connect in case of connection loss;

**It doesn't require a change on existing implementation and it works well for the existing events; with a full scenario screencast can be watched from [here](http://recordit.co/7mzvJlywm3).**

## How Has This Been Tested?

It's tested manually for now, there will be integration tests for the worker and client side implementation in the near feature.

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/client/app/lib/nodenotificationclient.coffee
+++ b/client/app/lib/nodenotificationclient.coffee
@@ -62,7 +62,8 @@ module.exports = class NodeNotificationClient extends kd.Object
       debug 'got auth ok, ready to go!'
       @emit 'ready'
     else
-      debug 'an unknown message received', message
+      debug 'got new message', message
+      @emit 'message', message
 
 
   _send: (obj) ->
--- a/client/app/lib/nodenotificationclient.coffee
+++ b/client/app/lib/nodenotificationclient.coffee
@@ -53,9 +53,10 @@ module.exports = class NodeNotificationClient extends kd.Object
     debug 'initiating the socket connection...'
     @_socket = new SockJS SUBSCRIBE_ENDPOINT
 
-    @_socket.onmessage = @bound '_handleNewMessage'
-    @_socket.onclose   = @bound '_handleClose'
-    @_socket.onopen    = @bound '_handleOpen'
+    @_socket.addEventListener 'message', @bound '_handleNewMessage'
+    @_socket.addEventListener 'close',   @bound '_handleClose'
+    @_socket.addEventListener 'error',   @bound '_handleError'
+    @_socket.addEventListener 'open',    @bound '_handleOpen'
 
     return this
 
@@ -85,6 +86,10 @@ module.exports = class NodeNotificationClient extends kd.Object
     @_bo.backoff()  if event?.code > 1000
 
 
+  _handleError: (error) ->
+    debug 'got an error', error
+
+
   _handleNewMessage: (message) ->
 
     return  unless message = @_parse message
@@ -106,10 +111,7 @@ module.exports = class NodeNotificationClient extends kd.Object
 
   _send: (obj) ->
 
-    if @_socket
-      @_socket.send JSON.stringify obj
-    else
-      debug "it's too early to send a message, connect first!"
+    @_socket.send JSON.stringify obj
 
 
   _parse: (message) ->
--- a/client/app/lib/nodenotificationclient.coffee
+++ b/client/app/lib/nodenotificationclient.coffee
@@ -12,6 +12,7 @@ isLoggedIn = require 'app/util/isLoggedIn'
 module.exports = class NodeNotificationClient extends kd.Object
 
   SUBSCRIBE_ENDPOINT = '/notify/subscribe' # this can be taken from config ~ GG
+  [ AUTH, MESSAGE, OK ] = ['auth', 'message', 'ok']
 
 
   constructor: (options = {}, data) ->
@@ -106,19 +107,20 @@ module.exports = class NodeNotificationClient extends kd.Object
 
     return  unless message = @_parse message
 
-    if message.auth is 'ok' and not message.__to?
+    if message.type is AUTH and message.status is OK
       debug 'got auth ok, ready to go!'
       @emit 'ready'
 
-    else
-      if message.__to.user?
-        debug "got new message for #{message.__to.user} user", message
+    else if message.type is MESSAGE
+
+      if user = message.recipient.user
+        debug "got new message for #{user} user", message
         to = 'message'
       else
-        debug "got new message for #{message.__to.group} group", message
+        debug "got new message for #{message.recipient.group} group", message
         to = 'group:message'
 
-      @emit to, message
+      @emit to, message.content
 
 
   _send: (obj) ->
@@ -131,8 +133,7 @@ module.exports = class NodeNotificationClient extends kd.Object
     try
       message = JSON.parse message.data
     catch e
-      debug 'got a corrupted message', message
-      @close internal = yes
+      debug 'got a corrupted message, dropping to the floor', message, e
       return null
 
     return message
--- a/client/app/lib/nodenotificationclient.coffee
+++ b/client/app/lib/nodenotificationclient.coffee
@@ -6,9 +6,7 @@ SockJS = require 'sockjs-client'
 backoff = require 'backoff'
 
 kookies = require 'kookies'
-globals = require 'globals'
-
-nick = require 'app/util/nick'
+isLoggedIn = require 'app/util/isLoggedIn'
 
 
 module.exports = class NodeNotificationClient extends kd.Object
@@ -19,7 +17,6 @@ module.exports = class NodeNotificationClient extends kd.Object
   constructor: (options = {}, data) ->
 
     super options, data
-
 
     @_bo = backoff.exponential
       initialDelay : 1200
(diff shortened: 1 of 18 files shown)


client/app/lib/realtimecontroller.coffee
```diff
@@ -61,58 +66,52 @@ module.exports = class RealtimeController extends KDController
     # https://www.pivotaltracker.com/story/show/88210800
     @pbNotification = PUBNUB.init options
 
+    fetchServerTime = (callback = kd.noop) =>
 
-  syncTime: ->
+      @pubnub.time (timestamp) =>
+        return callback Date.now() * 10000  unless timestamp
```
szkl: Would it have any side-effect to set browser time as `@lastSeenOnline` property?

workers/notification/lib/notification/notificationworker.coffee
```diff
@@ -0,0 +1,335 @@
...
+    return  unless @_connections[connection.id]
+
+    @_verifiedConnections = @_verifiedConnections.filter (connectionId) ->
+      connection.id isnt connectionId
+
+    @_verifiedRoutes = @_verifiedRoutes.filter (route) ->
+      [ ... , connectionId ] = route
+      connection.id isnt connectionId
+
+    delete @_connections[connection.id]
```
szkl: @cihangir L310?

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+
+    debug 'closing the connection'
+    @_socket?.close()
+
+
+  _handleOpen: ->
+
+    debug 'connection established, sending auth message...'
+    @_send @_getAuth()
+    kd.utils.wait 5000, =>
```
gokmen: if it ends up with connecting to sockjs server and sending the message sometimes I encountered no response issue that just happens. to prevent possible race condition there I'm waiting for 5 sec. and checking for the connection status again. otherwise backoff is handling retry part.

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+
+
+  _connect: ->
+
+    debug 'initiating the socket connection...'
+    @_socket = new SockJS SUBSCRIBE_ENDPOINT
+
+    @_socket.onmessage = @bound '_handleNewMessage'
+    @_socket.onclose   = @bound '_handleClose'
+    @_socket.onopen    = @bound '_handleOpen'
```
szkl: I think this should subscribe to `error` event too.

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+
+  _handleClose: (event) ->
+
+    debug 'connection closed by remote.'
+
+    # this is due to readyState implementation in kd.Object ~ GG
+    @readyState = 0
+    @once 'ready', => @readyState = 1
+
+    @_bo.backoff()  if event?.code > 1000
```
gokmen: I'll add as a comment, but it's generic knowledge I thought :) it's simply this https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent `0-999` is not used, `1000` is normal close if some how if we implement stop notifications we can call `close` manually or if connection closed by remote (if your session is invalid etc.) we don't need/want to retry to connect. But in any other cases `> 1000` we will try to re-connect.

client/app/lib/nodenotificationclient.coffee
szkl: An alternative way to bind event handlers would be calling [`EventTarget::addEventListener`](https://github.com/sockjs/sockjs-client/blob/master/lib/event/eventtarget.js#L11-L22) method. That way can easily address issues like [this](https://github.com/koding/koding/pull/11234/files#r118319648).

szkl (changes requested):

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+
+
+  _handleNewMessage: (message) ->
+
+    return  unless message = @_parse message
+
+    if message.auth is 'ok' and not message.__to?
```
gokmen: I thought the same but I didn't want to conflict with existing notification schema that is provided by social api dispatcher. currently there are different channels for handling `__to` part, and I tried to find a name for it in the message as much as possible to prevent conflict. but I can make something `message.type = 'notification:auth', message.status = ok` for auth handling as suggested.

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+      debug "it's too early to send a message, connect first!"
+
+
+  _parse: (message) ->
+
+    try
+      message = JSON.parse message.data
+    catch e
+      debug 'got a corrupted message', message
+      @close internal = yes
```
gokmen: I did the same on the worker side too, but I'm not sure too. I just tried to prevent sending non standard messages (ie: JSON) between server-client. I can only log that.

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+      return null
+
+    return message
+
+
+  _getAuth: ->
+
+    return {
+      auth : {
+        clientId  : kookies.get('clientId')
```
gokmen: yup, username and groupName was only for verification but you're right we don't need that query to go through db.

workers/notification/lib/notification/notificationworker.coffee
```diff
@@ -0,0 +1,335 @@
...
+    @_connections = {}
+    @_verifiedConnections = []
+    @_verifiedRoutes = []
+
+
+  start: ->
+
+    @log 'starting worker...'
+
+    @_connectToRabbitMQ()
```
gokmen: we're consuming but we're not sending them through before making the bongo connection https://github.com/koding/koding/pull/11234/files#diff-8367197550fb4e5f4852ba02179f6710R70

package.json
```diff
@@ -96,6 +96,7 @@
     "sendmail": "1.1.1",
     "shortid": "2.2.6",
     "simple-recaptcha": "0.0.3",
+    "sockjs": "^0.3.18",
```
gokmen: yeah why it didnt 🤔 I'll update in any case.

client/app/lib/realtimecontroller.coffee
gokmen: have no idea, just moved it out to same block with init flow. any idea @cihangir ?

client/app/lib/nodenotificationclient.coffee
gokmen: errors are not handled if they don't end with close so I ended up with using close only, what should we do when error happened but connection is still open?

Didn't your last comment though :) how it will help to address the issue?

package.json
szkl: https://github.com/koding/koding/pull/11239

client/app/lib/nodenotificationclient.coffee
```diff
@@ -0,0 +1,136 @@
...
+      else
+        debug "got new message for #{message.__to.group} group", message
+        to = 'group:message'
+
+      @emit to, message
+
+
+  _send: (obj) ->
+
+    if @_socket
```
szkl: This can be improved by also checking the ready state of SockJS client.

```coffeescript
if @_socket and @_socket.readyState is SockJS.OPEN
```

szkl (changes requested):

client/app/lib/nodenotificationclient.coffee
szkl: We can simply log the errors first and deal with handling them as we see any.

If connection ready state is not open yet then just bind an event handler and send the message in the inline callback function. Otherwise, messages need to be stored in a queue.

gokmen: ah ok I got it now, you're right.

client/app/lib/realtimecontroller.coffee
cihangir: users might have invalid times, that creates issue while fetching the message history

gokmen:
all the issues resolved cc/ @cihangir @szkl

szkl (approved):

cihangir (approved):

Issue #11191, opened 2017-05-16
Switch to another team modal is broken


cihangir:
http://snpy.in/SG97OZ

PR #11238, opened 2017-05-24
Change Team Modal UI fix

hakankaradis:
http://recordit.co/Po7Qmos86A
Fixes: #11191

### DIFF
--- a/client/app/lib/changeteam/index.coffee
+++ b/client/app/lib/changeteam/index.coffee
@@ -29,6 +29,6 @@ module.exports = class ChangeTeamView extends kd.ModalView
 
     controller = new ChangeTeamController()
     @addSubView list = controller.getView()
-    controller.on 'ItemsLoaded', ->
+    controller.on 'AllItemsAddedToList', ->
       # delay is needed to let custom scroll view set their css classes on the list
       kd.utils.defer -> list.setClass 'loaded'
--- a/client/app/lib/sidebar/components/resources.coffee
+++ b/client/app/lib/sidebar/components/resources.coffee
@@ -4,7 +4,7 @@ List = require 'app/components/list'
 
 OwnedResourcesList = require './ownedresourceslist'
 SharedResourcesList = require './sharedresourceslist'
-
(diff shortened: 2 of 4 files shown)


client/app/lib/sidebar/components/resources.coffee
```diff
@@ -4,7 +4,7 @@ List = require 'app/components/list'
 
 OwnedResourcesList = require './ownedresourceslist'
 SharedResourcesList = require './sharedresourceslist'
-
+SidebarGroupDisabled = require './sidebargroupdisabled'
```
hakankaradis: @usirin

client/app/lib/changeteam/index.coffee
```diff
@@ -29,6 +29,6 @@ module.exports = class ChangeTeamView extends kd.ModalView
 
     controller = new ChangeTeamController()
     @addSubView list = controller.getView()
-    controller.on 'ItemsLoaded', ->
+    controller.on 'AllItemsAddedToList', ->
```
hakankaradis: kdlistviewcontroller doesn't emit `ItemsLoaded` 
@gokmen according to testfile
https://github.com/koding/koding/blob/master/client/app/test/kodinglistcontroller.test.coffee#L329
It should emit `ItemsLoaded`. Am I missing something or this usage is deprecated?

client/app/lib/sidebar/components/resources.coffee
usirin: :+1:

usirin (approved):

client/app/lib/changeteam/index.coffee
gokmen: @hakankaradis yes, you're right it was emiting that before I removed it in here https://github.com/koding/koding/pull/11150/files#diff-28c8fbaa26775fa401a9d6974d2f1232L122 ✌️ but I didn't understand how the test you provided passed in my PR when I delete that. @szkl, @usirin any ideas?

PR #11279, opened 2017-06-02
klientctl: cobra CLI infrastructure implementation

ppknap:
This PR introduces `spf13/cobra` CLI to `kd`

It defines all middlewares that are required for correct KD behavior.

⚠️ This PR doesn't change current `kd` CLI so it can be merged and used for testing. In order to use `cobra` cli `cobra` tag is required during `klientctl` installation/compilation.

Currently, there is only one command implemented(as a POC) `kd version`.

## Architecture

Cobra commands usually are stored in `commands` directory. Which contains command constructors and implementation.

Inside `commands` directory, we store `cli` package(similar to `docker` but with proper package tree). The `cli` package defines command *middlewares* - modifiers to the `RunE` functions that adds common functionality. Despite middlewares, `cli` package also contains `CLI` *type*, which object is passed to all commands and stores data streams, loggers, metrics client and all other data structures that might be used by command implementations. `CLI` type is like `context.Context` but it stores persistent(per process) objects.

`main.go` is kept as small as possible. All other functionality is added per commands via *middlewares*.

Moreover:
* there are no global objects/flags used by CLI.
* any input/output goes through `*cli.CLI` object. No need to run separate process during tests.
* any mocking should be done by modifying `*cli.CLI` object. 
* aliased commands should have its `New*CommandAlias` constructor to work properly with `metric` middleware.

The architecture is based on `cobra`'s best practices and takes some ideas from `github.com/rkt/rkt`, `github.com/moby/moby`(docker), and `github.com/spf13/hugo` packages.

## Motivation and Context
Better `kd` UX and pre-requirement for command autocompletion.

## How Has This Been Tested?
Manually.

https://asciinema.org/a/f23l0giazbucqf2676vrlfr3z

## Types of changes


go/src/koding/klientctl/commands/cli/cli.go
```diff
@@ -0,0 +1,145 @@
...
+	if w == nil {
+		w = ioutil.Discard
+	}
+
+	handler := logging.NewWriterHandler(w)
+
+	// Make handler writer accept all incoming log messages.
+	handler.SetLevel(logging.DEBUG)
+
+	logger := logging.NewLogger("kd")
```
cihangir: what about NewCustom?

go/src/koding/klientctl/commands/cli/cli.go
```diff
@@ -0,0 +1,145 @@
...
+func (c *CLI) Log() logging.Logger {
+	if c.log != nil {
+		return c.log
+	}
+
+	return newLogger(nil)
+}
+
+// Metrics returns metrics client or nil if not enabled.
+func (c *CLI) Metrics() *metrics.Metrics {
```
cihangir: Seems like we can export c.m as c.Metrics

go/src/koding/klientctl/commands/version/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		v.Latest, _ = config.LatestKDVersionNum()
+		if opts.jsonOutput {
+			cli.PrintJSON(c.Out(), v)
+			return nil
+		}
+
+		fmt.Fprintf(c.Out(), "Installed Version: %s\n", getReadableVersion(v.Installed))
+		fmt.Fprintf(c.Out(), "Latest Version: %s\n", getReadableVersion(v.Latest))
```
cihangir: should we omit this line if Latest is empty?

cihangir (approved):
LGTM 👍

rjeczalik (approved):

go/src/koding/klientctl/commands/cli/cli.go
ppknap: Yes, we can. However, I'm still not sure if this becomes an interface (like in Docker).

Moreover, I wanted to not add a public member to the structure which has only getters(consistency).

PR #11318, opened 2017-06-19
Update time attributes of files restored from cache

szkl:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,6 +137,13 @@ jobs:
             - go-{{ .Branch }}
             - go-master
 
+      - run:
+          name: touch files restored from cache
+          command: |
+            timestamp=$(date --date="@$(($(date +%s) + 60))" +%m%d%H%M)
+            find go/bin -exec touch {} -t $timestamp \;
+            find go/pkg -exec touch {} -t $timestamp \;
+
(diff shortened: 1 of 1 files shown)


rjeczalik (approved):

cihangir (approved):

cihangir:
why is this required?

rjeczalik:
Circle caches pkg/ and bin/,  once unzipped go rebuilds them either way
since source tree has mtime > the archived ones.

cihangir:
I see go build run again for this PR, could we say that does not solve the rebuilding issue?

rjeczalik:
Maybe there's issue with caching? (e.g. single cache file and some feature
brach overwrites it?)

szkl:
@cihangir you're right. There are still some packages rebuilt. I've one more idea to try which might make this pull request obsolete. However, this change helps with the excessive memory allocation by the build process. It can be evaluated as a band-aid to achieve ✅  in the CI.

cihangir:
@rjeczalik i was assuming it was always the cache being an issue from the beginning 

> There are still some packages rebuilt.

@szkl what about running that command under src folder too? if the issue is 

> mtime > the archived ones.

it is worth trying

szkl:
🏁

cihangir (approved):

PR #11316, opened 2017-06-19
klientctl: add bash autocomplete for machines and mounts.

ppknap:
~Depends on #11314~

PR #11321, opened 2017-06-22
klientctl: switch CLI to cobra

ppknap:
~Depends on: #11316~


ppknap:
Tests will be broken until #11329 is merged.

rjeczalik (approved):

PR #11326, opened 2017-06-23
Introduce Circle CI workflow

szkl:


.circleci/config.yml
```diff
@@ -127,33 +132,33 @@ jobs:
...
           command: |
             timestamp=$(date --date="@$(($(date +%s) + 120))" +%m%d%H%M)
             base=origin/master
             head=$(git rev-parse HEAD)
             git diff --name-only $base...$head go/ | xargs --no-run-if-empty touch -t $timestamp
-
       - run: env GOGC=10 go/build.sh
-
       - save_cache:
           key: go-{{ .Branch }}-{{ .Revision }}-{{ epoch }}
```
cihangir: what about just storing {{ .Branch }} and not caring about other stuff ie revisions?

cihangir (approved):
AWESOME!1!11!

.circleci/config.yml
szkl: I think it'd be fine. The reason I've added revision variants is to eliminate the possibility of a conflict during a rebuild of an older build which would restore the cache from a newer build.

Let's experiment without revision variants. I'll update the pull request.

PR #11334, opened 2017-06-28
Run nginx as a worker

szkl:

### DIFF
--- a/.ebextensions/004_files_config_updaters.config
+++ b/.ebextensions/004_files_config_updaters.config
@@ -130,22 +130,11 @@ files:
       #!/bin/bash
       source /etc/profile
 
-      cd $EB_APP_DEPLOY_DIR
-
       echo "stopping httpd if exists"
       service httpd stop || echo "httpd is stopped"
 
-      echo "updating ngnix configuration"
-      echo "moving nginx config"
-      cp deployment/generated_files/nginx.conf /etc/nginx/
-      echo "updated nginx config"
-
       echo 'koding:$apr1$K17a7D.N$vuaxDfc4kJvHAg7Id43wk1' >/etc/nginx/conf.d/.htpasswd
       echo 'USER3489:$apr1$52wVqnxi$eVUrgOn8oRfPQRG8vU8jE0' >>/etc/nginx/conf.d/.htpasswd
-
-      echo "restarting ngnix with current config"
-      service nginx restart
-      echo "restarted nginx"
 
 
   "/opt/elasticbeanstalk/hooks/appdeploy/post/080_configure_supervisord.sh":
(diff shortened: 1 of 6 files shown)


gokmen (approved):

config/workers.coffee
```diff
@@ -8,6 +8,11 @@ module.exports = (KONFIG, options, credentials) ->
   nodeProgram = if options.watchNode then './watch-node' else 'node'
 
   workers =
+    nginx:
+      group: 'webserver'
+      supervisord:
+        command: 'nginx -c %(ENV_KONFIG_PROJECTROOT)s/nginx.conf'
```
cihangir: s/nginx.conf/deployment\/generated_files/nginx.conf

also could you remove the `fs.writeFileSync "./nginx.conf", KONFIG.nginxConf` line from configure? any idea why we are duplicating it?

cihangir: could you also change the "stopsignal" to QUIT for supervisord conf of this worker? we will make sure that nginx is restarted/stopped [gracefully](http://github.perusio.org/nginx-manpage/#SIGNALS).

.ebextensions/004_files_config_updaters.config
```diff
@@ -122,26 +122,22 @@ files:
       chown -R $EB_APP_USER:$EB_APP_USER $EB_APP_DEPLOY_DIR
 
 
-  "/opt/elasticbeanstalk/hooks/appdeploy/post/070_configure_supervisord.sh":
+  "/opt/elasticbeanstalk/hooks/appdeploy/post/070_configure_nginx.sh":
```
cihangir: old files might stay during updates if machine is not get recycled

cihangir (changes requested):
better to have stopsignal changed to QUIT

config/workers.coffee
szkl: I don't recall. I suppose it was just to have a shorter path.

Stop signal is a good catch. Also, that reminded me an old observation. I think `supervisorctl shutdown` kills the subprocesses because it is much faster than `supervisorctl stop all`. I'll dig into it.

.ebextensions/004_files_config_updaters.config
szkl: I should add current file names to overrides configuration.

szkl:
Updated.

config/workers.coffee
szkl: I'll remove redundant configuration output files in a separate pull request.

deployment/supervisord.coffee
```diff
@@ -51,7 +51,7 @@ generateWorkerSection = (app, options = {}) ->
     autorestart             : yes
     startsecs               : 10
     startretries            : 5
-    stopsignal              : 'TERM'
+    stopsignal              : options.stopsignal or 'TERM'
```
cihangir: this is not needed, line 65 already handles those

cihangir (approved):

PR #11338, opened 2017-06-30
Service installation for Kubernetes

mertaytore:
Initial version of the installation process for Kubernetes.

## Description
I added two new commands to the auto-generated ./run script: `./run k8s start` and `./run k8s stop`. Start runs to check whether the dependencies have been previously installed or not. If they have, workers pod is created according to the pod definition in containers.yaml.template
Stop command stops the running Kubernetes cluster (minikube)

## Motivation and Context
A step for Koding on K8s. After having a copy of this repository, `./configure` is needed to be executed to have a ./run script. If the dependencies regarding Kubernetes are fulfilled, `./run k8s start` builds up a local Kubernetes cluster. As a result, Koding runs on team stack for development <your-hostname>:8090

## How Has This Been Tested?
Services' connectivities are tested via their respective check_connectivity functions. 

## Types of changes


szkl (changes requested):
I think it's best to create a new file which sets related environment variables for a service. `k8s` command would source that file. It is possible to have connectivity checks that way.

config/generateRunFile.coffee
```diff
@@ -442,6 +442,72 @@ generateDev = (KONFIG, options) ->
...
+
+        echo ""
+        exit 1
+      ;;
+      esac
+
+      if [ "$param" == "start" ]; then
+        command -v VirtualBox       >/dev/null 2>&1 || { echo >&2 "I require curl but it's not installed.  Aborting."; exit 1; }
+        command -v minikube         >/dev/null 2>&1 || { echo >&2 "I require a Kubernetes cluster. To install minikube: \
+                (curl -Lo minikube https://storage.googleapis.com/minikube/releases/v0.20.0/minikube-darwin-amd64 && chmod +x minikube && mv minikube /usr/local/bin/)"; exit 1;}
```
cihangir: if the naming scheme is consistent in download links, you could infer the the os and replace linux/darwin with that

config/generateRunFile.coffee
```diff
@@ -442,6 +442,72 @@ generateDev = (KONFIG, options) ->
...
+        command -v kubectl          >/dev/null 2>&1 || { echo >&2 "I require kubectl. To install kubectl: \
+                (curl -L -O https://storage.googleapis.com/kubernetes-release/release/v1.6.4/bin/linux/amd64/kubectl && chmod +x kubectl && mv kubectl /usr/local/bin/)"; exit 1;}
+        minikube start
+
+        if [[ `uname` == 'Darwin' ]]; then
+          command -v helm           >/dev/null 2>&1 || { echo >&2 "I require kubernetes-helm but it's not installed. (brew install kubernetes-helm maybe?)"; exit 1; }
```
cihangir: you can advise this https://github.com/kubernetes/helm/blob/master/docs/install.md#from-script

mertaytore:
ready for another review, updated the PR description.

config/generateRunFile.coffee
```diff
@@ -442,6 +426,116 @@ generateDev = (KONFIG, options) ->
...
+        command -v helm             >/dev/null 2>&1 || { echo >&2 "I require helm but it's not installed. To install helm: \
+                (curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get > get_helm.sh && chmod 700 get_helm.sh && ./get_helm.sh)"; exit 1; }
+
+        # minikube must be deleted and recreated if a disk resize is needed
+        minikube start --disk-size 50g
+
+        export MINIKUBE_IP=$(minikube ip)
+        eval $(minikube docker-env)
+
+        docker image inspect baseinstall:v1 >/dev/null 2>&1 && echo yes || echo no >/tmp/response.log 2>&1
```
cihangir: you dont need to write to a file to get the response, you could have it in the exported value

```
➜  cat test.sh
#!/bin/bash

function fail() { return 1; }
function succ() { return 0; }

succ && echo "succeeded"
fail || echo "failed"

# set the default behaviour
export RESULT="no";

succ && RESULT="yes"
echo $RESULT

fail || RESULT="no"
echo $RESULT

➜ ./test.sh
succeeded
failed
yes
no

```

config/generateRunFile.coffee
```diff
@@ -442,6 +426,116 @@ generateDev = (KONFIG, options) ->
...
+
+        export IMAGE_RESPONSE=$(tail -1 /tmp/response.log)
+
+        if [[ "$IMAGE_RESPONSE" == "no" ]]; then
+          echo 'base installation docker image for koding was not found, building the image:\n'
+          docker build -t baseinstall:v1 -f Dockerfile-k8s-base .
```
cihangir: lets not build the image in user machines, that might create security issues where we can add sensitive information that the codebase has.

config/generateRunFile.coffee
```diff
@@ -442,6 +426,116 @@ generateDev = (KONFIG, options) ->
...
+        fi
+
+        minikube ssh "docker pull mongo:3.2.8"
```
cihangir: these operations should not be required, kubernetes handles the image pull ops

config/generateRunFile.coffee
```diff
@@ -442,6 +426,116 @@ generateDev = (KONFIG, options) ->
...
+        minikube ssh "docker pull koding/postgres:latest"
+        #minikube ssh "docker pull koding/countly-server:latest"
+        minikube ssh "docker pull rabbitmq:3-management"
+        minikube ssh "docker pull redis"
+        minikube ssh "docker pull koding/base"
+
+        kubectl create -f $KONFIG_PROJECTROOT/deployment/kubernetes/backend/backend-deployment.yaml
+
+        export BACKEND_POD_NAME=$(kubectl get pods --namespace default -l "app=backend" -o jsonpath="{.items[0].metadata.name}")
```
cihangir: * better to use a specific namespace instead of the default one
* app=backend is  not a good selector where many other test apps can have "backend" in the default namespace - not an issue once we switch to a dev namespace

mertaytore: made changes to all reviews accordingly; created a namespace named service-dev (replacing default namespace) and renamed app label to service-installations.

config/generateRunFile.coffee
szkl: Base images should be generally available already and also must not contain any installation-specific files.

config/generateRunFile.coffee
```diff
@@ -424,6 +424,124 @@ generateDev = (KONFIG, options) ->
...
+        echo "Commands:  "
+        echo "  start : install the services in Kubernetes pods"
+        echo "  stop  : stop the currently running Kubernetes cluster"
+
+        echo ""
+        exit 1
+      ;;
+      esac
+
+      if [ "$param" == "start" ]; then
```
szkl: This is a very big block. Let's split it into a function.

config/generateRunFile.coffee
```diff
@@ -424,6 +424,124 @@ generateDev = (KONFIG, options) ->
...
+          echo ' timed out!'
+          exit 255
+        fi
+
+        echo -n '.'
+
+        sleep $interval
+        duration=$((duration + interval))
+
+        declare response_code=$(kubectl get pods --namespace service-dev -o jsonpath="{.items[0].status.phase}")
```
szkl: `declare` is redundant here.

config/generateRunFile.coffee
```diff
@@ -424,6 +424,124 @@ generateDev = (KONFIG, options) ->
...
+      echo ' succeeded!'
+    }
+
+    function check_pod_ready () {
+      declare interval=5
+      declare timeout=600
+      declare duration=0
+      declare response_code=$(kubectl get pods --namespace service-dev -o jsonpath="{.items[0].status.phase}")
+
+      until [[ $response_code != *"Pending"* ]]; do
```
szkl: How about invoking a function which basically returns 0 or 1 depending on the response code?

config/generateRunFile.coffee
```diff
@@ -659,6 +777,9 @@ generateDev = (KONFIG, options) ->
     elif [ "$1" == "switchclient" ]; then
       switch_client_version $2
 
+    elif [ "$1" == "k8s" ]; then
+      k8s $2
```
szkl: For forward-compatibility, you can shift the first parameter and pass the rest to the `k8s` function. E.g.:

```shell
...
elif [ "$1" == "k8s" ]; then
  shift
  k8s "$@"
...
```

deployment/kubernetes/workers-pod/containers.yaml
```diff
@@ -0,0 +1,109 @@
...
+        value: "6379"
+      - name: GOPATH
+        value: $(KONFIG_PROJECTROOT)/go
+      - name: GOBIN
+        value: $(GOPATH)/bin
+      ports:
+        - containerPort: 8090
+          hostPort: 8090
+      workingDir: /opt/koding
+      command: [ "scripts/bootstrap-container", "--config", "dev", "--projectRoot", "/opt/koding", "--publicPort", "8090", "--host", "$(hostname)", "--hostname", "$(hostname)" ]
```
szkl: --host option should be `$(hostname):8090`.

deployment/kubernetes/workers-pod/containers.yaml
```diff
@@ -0,0 +1,109 @@
...
+      image: redis
+      ports:
+        - containerPort: 6379
+          hostPort: 6379
+      workingDir: /opt/koding
+
+    - name: landing
+      image: koding/base
+      workingDir: /opt/koding
+      command: [ "make", "-C", "client/landing", "dev" ]
```
szkl: I'd prefer --directory instead of -C, personally.

szkl (changes requested):

szkl (approved):

cihangir:
🎉

PR #11364, opened 2017-07-18
provider: xss fix

mertaytore:
## Description
A fix to resolve the XSS regarding provider authorization, with XSSEncode.

### DIFF
--- a/workers/social/lib/social/error.coffee
+++ b/workers/social/lib/social/error.coffee
@@ -4,9 +4,11 @@ KnownErrors =
 
 module.exports = class KodingError extends Error
 
+  encoder   = require 'htmlencode'
+
   constructor: (message, name, errorObject) ->
 
-    return new KodingError message  unless this instanceof KodingError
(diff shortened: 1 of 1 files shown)


cihangir (approved):

workers/social/lib/social/error.coffee
```diff
@@ -4,9 +4,11 @@ KnownErrors =
 
 module.exports = class KodingError extends Error
 
+  encoder   = require 'htmlencode'
```
sinan: unless you have a valid reason, please take `require` line on top of the file and fix spacing between the variable name and the equal sign.

sinan (changes requested):
looks good only a code styling issue.

workers/social/lib/social/error.coffee
mertaytore: did the changes as requested.

sinan (approved):

PR #11365, opened 2017-07-20
Switch using stripe.js/v3 - stripe elements

usirin:
stripe-elements

/cc @cihangir

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -35,17 +35,9 @@ savePaymentToken = (token) ->
   global._payment or= {}
   _payment.token = token
 
-saveCardInfo = ({ number, exp_month, exp_year }) ->
+saveCardElement = (cardElement) ->
   global._payment or= {}
-
-  _payment.card =
-    exp_month: exp_month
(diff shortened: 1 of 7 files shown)


sinan (approved):
👍

sinan:
tests 👀

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -557,6 +557,14 @@ module.exports = utils = {
       else resolve response.token.id
 
 
+  submitWithDummyToken: (client) ->
+    STRIPE_TEST_TOKEN = 'tok_visa'
```
usirin: /cc @cihangir since there is [no way to pre-fill sensitive data](https://stripe.com/docs/stripe.js#element-options) using Stripe elements, and the new way of creating token is [by passing a instance](https://stripe.com/docs/stripe.js#stripe-create-token) of a `StripeElement` to `stripe.createToken` function, they now provide test tokens to bypass this process.

This eliminates the need to make a call for creating a token.

usirin: Not sure why this comment looks as an outdated comment :/

PR #11381, opened 2017-08-01
Remove JView and Pistachio dependency

gokmen:
In `kd.js` we've moved back Pistachio functionality thus there is no need another abstraction here in Koding Main App.

## Description
From now on (once https://github.com/koding/kd/pull/174 merged) one can use Pistachio templates on any `kd.View` instance. If needed to use like in `kd.ListItemView` we can use `kd.View::viewAppended` there too.

## Motivation and Context
Since the functionality moved away from `kd.js`, there were no template functionality in the framework, so, I've put back the functionality to the framework and removed all customisations from main app instead.

## How Has This Been Tested?
I've tested manually all the views that has usage of either `JView`, `JCustomHTMLView` or `Pistachio` independently.

## Checklist:
- [x] All new and existing tests passed. 🤞


gokmen:
Had to close previous PR #11380 since CircleCI got messed up with tests but it seems it's not ok with this one as well 🤕 @szkl halp 🖐

cihangir:
client build fails at

```
[09:33:37] Finished 'libs' after 132 ms
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Parsing file /koding/client/landing/node_modules/kd.js/index.coffee: Unexpected token (1:25)
```

https://circleci.com/gh/koding/koding/3991

gokmen:
@cihangir that package has been updated but I still couldn't convince circleci to get the new one :)

szkl:
🦄

sinan (approved):
👍

PR #11382, opened 2017-08-02
Support for npm 5.x and node 8.x

gokmen:
Fixes npm5 issue with symlinked node modules, with current `node_modules_koding` and existing post/pre install scripts in package.json, `npm5` was deleting some of the packages after install. To be able to fix that I've reverted symlinks before installing packages then created them back after install.

Right now with following changes I can run koding with following;

```
(npm5) $ node -v
v8.2.1
(npm5) $ npm -v
5.3.0
```

ps. For now I've ignored the `package-lock.json` since there are still issues with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,7 @@ elektron/node_modules
 
 .env.sh
 .env.json
+package-lock.json
 website/apidocs/lang
 
 credential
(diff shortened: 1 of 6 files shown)


szkl (approved):

cihangir:
Well i was waiting to update all the images before merging this one...

PR #11388, opened 2017-08-03
generate-bongo-schema: unsuccessful termination of schema generation

mertaytore:
Changed the termination status of generate-bongo-schema script. With this fix, the script won't let a makefile, who utilizes this script, to proceed its execution if the given file for schema is invalid.


gokmen (approved):

szkl (changes requested):
Test results are missing.

PR #11420, opened 2017-09-05
Fallback to a generic error view when the card is declined

usirin:
fixes the problem of https://twitter.com/cdinu/status/901548830621040641

@sinan I hunted down the response he got in stripe, and recorded this video especially with the same error object which is:

```js
{
    "message": "Your card was declined.",
    "type": "card_error",
    "code": "card_declined",
    "decline_code": "do_not_honor"
}
```

landing-billing-generic-errors

Also generic error messages are parsed from https://stripe.com/docs/declines/codes and replaced the sensitive error messages with generic ones.

### DIFF
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -522,7 +522,7 @@ module.exports = utils = {
       data: JSON.stringify { source: { token }, email }
       success: resolve
       error: (err) ->
-        reject JSON.parse err.responseText
+        reject JSON.parse err.responseJSON?.description
 
     $.ajax options
 
--- a/client/landing/site.landing/coffee/core/utils.coffee
+++ b/client/landing/site.landing/coffee/core/utils.coffee
@@ -522,7 +522,10 @@ module.exports = utils = {
       data: JSON.stringify { source: { token }, email }
       success: resolve
       error: (err) ->
-        reject JSON.parse err.responseJSON?.description
+        try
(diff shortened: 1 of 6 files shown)


usirin:
Btw, the reason for errors not showing up before this PR is that stripe changed their response format again, i needed to parse it to make it work again.

https://github.com/koding/koding/pull/11420/commits/3109c160bc5aa2a274921d938f7bef1844f181bb fixes that problem.

szkl (approved):

client/landing/site.landing/coffee/core/utils.coffee
```diff
@@ -522,7 +522,7 @@ module.exports = utils = {
       data: JSON.stringify { source: { token }, email }
       success: resolve
       error: (err) ->
-        reject JSON.parse err.responseText
+        reject JSON.parse err.responseJSON?.description
```
sinan: trust but verify :) try/catch for json.parse.

sinan (changes requested):
🙏 sorry forgot to review this one. 👍 looks good thx, two minor things tho:
 - [ ] when failed the action should read: "Use a different card"
 - [ ] and the commented try/catch

sinan:
ping @usirin

usirin:
@sinan sorry that i missed the comments, updated now.

usirin:
@szkl dunno why but builds are failing constantly (go part is not even starting), can you take a look?

sinan (approved):
👍

Issue #10961, opened 2017-04-08
klient/machine: add label support for commands


rjeczalik:
```
$ kd machine ls
ID                        ALIAS           TEAM    STACK                  PROVIDER  LABEL           OWNER  AGE      IP  STATUS
58e5faf04e81359e50d221a0  green_date      barfoo  Poliwrath Aws Stack    aws       aws-instance.1         2d3h         offline (2d3h)
58e5faf04e81359e50d2219f  maroon_tomato   barfoo  Poliwrath Aws Stack    aws       aws-instance.0         2d3h         offline (2d3h)
5849a2ebca97813764f80894  navy_jackfruit  foobar  Sandslash Azure Stack  azure     azure-instance         120d17h      offline (120d17h)
58e5faf04e81359e50d221a1  red_orange      barfoo  Poliwrath Aws Stack    aws       aws-instance.2         2d3h         offline (2d3h)
```

It should be possible to use labels in `kd machine *` commands, like mount:

```
$ kd machine mount azure-instance:~/ /mnt/azure-instance
```

Also remove aliases or at least hide them from default output - using labels over aliases has the advantage they're sticky - an alias change when a stack is reinited, while label does not.

If some lookup / caching would be needed in order to use Kloud API / remote.api, extend kloud methods to accept also labels instead of identifiers, to spare time on lookups, cache invalidation etc.

TODO

- [ ] kd/machine: hide / remove aliases
- [ ] kd: add labels argument support
- [ ] klient/machine: add support for labels
- [ ] kloud: add support for labels

/cc @gokmen @sinan

ppknap:
@rjeczalik generally, it would be nice to use machine labels only, it definitely simplifies UX of KD so I'm :+1:. However, there is one problem with this approach - what are you/we going to support shared machines with the same label? eg.:

```
(archpk) ~$ kdev machine list
ID                        ALIAS           TEAM    STACK                PROVIDER  LABEL           OWNER        AGE     IP             STATUS
58ea7146879075600209c33f  olive_apple     mount   Rapidash Aws Stack   aws       aws-instance-1  pawelknap88  4m6s    52.57.74.32    online (2m22s)
58d830a068e72e4d19f7e76d  red_quince      koding  Default stack        aws       koding-vm-0                  13d20h                 offline (13d20h)
58d831373619795d1971221d  red_watermelon  mount   Magmar Aws Stack     aws       aws-instance                 13d20h  52.59.160.118  offline (3d20h: Machine is Stopped)
[...]

PR #11423, opened 2017-09-07
klient: add support for machine labels

ppknap:
## Description
This PR adds support for machine labels as machine identifiers. The syntax: `[owner@]label`, where `owner` is specified as shown by `kd list` ( current user of kd is identified by empty string `""` ).

Machines which have the same label and the same user are not supported. Error in form:
```
identifier "identifier" is ambiguous (matches: machineID1, machineID2...)
```
will be returned in such cases.

Auto-completion support for labels was is also included in this PR.

Aliases are hidden in `kd list` response but are still supported and visible in `kd list --json`. They will be dropped later.

/cc @gokmen 

## Motivation and Context
Solves: #10961

## How Has This Been Tested?
Unit tests, manually

https://asciinema.org/a/nUsI8qoGqEScXyK3jb75bMQxK

## Types of changes

### DIFF
--- a/go/src/koding/klient/machine/machine.go
+++ b/go/src/koding/klient/machine/machine.go
@@ -46,3 +46,9 @@ func (ids IDSlice) String() string {
 
 	return strings.Join(strs, ", ")
 }
+
+// Metadata stores additional information about single machine.
+type Metadata struct {
+	Owner string `json:"owner"`
+	Label string `json:"label"`
+}
--- a/go/src/koding/klient/machine/machine.go
+++ b/go/src/koding/klient/machine/machine.go
@@ -34,18 +34,25 @@ func (ids IDSlice) Less(i, j int) bool {
 	return string(ids[i]) < string(ids[j])
 }
 
-// String implements fmt.Stringer interface. It pretty prints machine IDs.
-func (ids IDSlice) String() string {
-	strs := make([]string, len(ids))
+// StringSlice converts ID to string slice.
+func (ids IDSlice) StringSlice() []string {
+	ss := make([]string, len(ids))
 	for i := range ids {
-		strs[i] = string(ids[i])
+		ss[i] = string(ids[i])
 	}
 
-	return strings.Join(strs, ", ")
+	return ss
+}
+
+// String implements fmt.Stringer interface. It pretty prints machine IDs.
+func (ids IDSlice) String() string {
+	return strings.Join(ids.StringSlice(), ", ")
 }
 
 // Metadata stores additional information about single machine.
 type Metadata struct {
 	Owner string `json:"owner"`
 	Label string `json:"label"`
+	Stack string `json:"stack"`
+	Team  string `json:"team"`
 }
--- a/go/src/koding/klient/machine/machine.go
+++ b/go/src/koding/klient/machine/machine.go
@@ -2,7 +2,6 @@ package machine
 
 import (
 	"errors"
-	"sort"
 	"strings"
 
 	"koding/klient/config"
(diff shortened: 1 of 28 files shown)


rjeczalik:
Still wip?

go/src/koding/klient/machine/machine.go
```diff
@@ -29,6 +30,13 @@ type ID string
 // IDSlice represents a set of machine IDs.
 type IDSlice []ID
 
+// Sort sorts machine IDs in lexicographical order.
+func (ids IDSlice) Sort() {
+	sort.Slice(ids, func(i, j int) bool {
+		return string(ids[i]) < string(ids[j])
+	})
+}
+
```
rjeczalik: Wouldn't it be better to have just `func (IDSlice) Less(i, j int) bool` and then do `sort.Slice(ids, ids.Less)` externally? You wouldn't need to alloc a func each time. Just an idea.

ppknap:
yep, just for manual tests will do in an hour(must leave now). You can review it now but please not merge

rjeczalik (approved):

go/src/koding/klient/machine/machine.go
ppknap: I can change this. Just for clarification - I don't allocate a new slice here. I'm reusing existing one.

go/src/koding/klient/machine/machinegroup/id.go
```diff
@@ -24,22 +26,36 @@ type IDResponse struct {
...
 //  - machine alias.
+//  - machine label in form [owner@]label.
 //  - machine IP address.
 func (g *Group) ID(req *IDRequest) (*IDResponse, error) {
 	if req == nil {
 		return nil, errors.New("invalid nil request")
 	}
 
+	owner, identifier := "", req.Identifier
+	if toks := strings.SplitN(identifier, "@", 2); len(toks) > 1 {
```
cihangir: just curious:

do we allow `@` in identifier?

asking because toks[1] might have more `@`s in it

cihangir (approved):

go/src/koding/klient/machine/machinegroup/id.go
rjeczalik: I believe we intentionally treat tok[1] (`label`) as an arbitrary string, so it can have any characters. Later lookup will just fail with 404.

rjeczalik:
```
#!/bin/bash -eo pipefail
./run exec scripts/api-generator.coffee --check
/bin/bash: ./run: No such file or directory
Exited with code 1
```
/cc @szkl

szkl:
This needs to be rebased on top of koding/koding#master branch.

go/src/koding/klient/machine/machinegroup/id.go
ppknap: this was discussed [here](https://github.com/koding/koding/issues/10961#issuecomment-292802212).

It doesn't cover all cases eg. `unix@machine` label which will fail due to `unix` token parsed as username. However, we still are able to workaround this by `@unix@machine`.

gokmen: why not searching for all occurrences and ask for the which one if there are multiple? So, you have a `machine` and tried to use `foo@machine`, search for `machine` and `foo@machine` respectively; if only one of them found use it as is, if there are multiple result ask for what to do. like in here https://github.com/koding/koding/issues/10961#issuecomment-298421459 `@unix@machine` is completely weird 👎

ppknap: @gokmen the problem with asking is that we cannot use this approach in non-interactive kd calls (eg atom IDE calls `kd mount <labal>`).

rjeczalik:
Is the PR ready to review?

ppknap:
> Is the PR ready to review?

🚀

go/src/koding/klient/machine/machinegroup/id.go
```diff
@@ -16,35 +18,96 @@ type IDRequest struct {
...
+
+	sort.Slice(ids, ids.Less)
+
+	return ids
+}
+
+// ItemDesc returns description string for each item stored in response.
+func (res *IDResponse) ItemDesc() (ds []string) {
+	ids := res.IDSlice()
+	for _, id := range ids {
```
rjeczalik: `for _, id := range res.IDSlice()`

rjeczalik (approved):

rjeczalik:
💯

PR #11433, opened 2017-09-22
Add nginx options to KONFIG object

szkl:

### DIFF
--- a/deployment/nginx.coffee
+++ b/deployment/nginx.coffee
@@ -40,6 +40,11 @@ allowInternal = '''
         deny                  all;
 '''
 
+real_ip_options =
+  from: '0.0.0.0/0'
+  header: 'X-Forwarded-For'
+  recursive: 'on'
+
 createWebLocation = ({ name, locationConf, cors }) ->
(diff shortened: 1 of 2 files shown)


deployment/nginx.coffee
```diff
@@ -460,3 +462,18 @@ createHttpsRedirector = (KONFIG) ->
       # use generic names, do not hardcode values
       return 301 https://$host$request_uri;
     }"""
+
+createHttpOptions = (KONFIG) ->
```
cihangir: Could you combine this  and createListenDirective together to have a unified functionality? (we can set the defaults for options.nginx for createListenDirective )

cihangir (changes requested):
unification needed

deployment/nginx.coffee
szkl: real_ip module directives are set in `http` block. It applies to everywhere functionally. Is this what you are suggesting?

cihangir: There is a function in this file called "createListenDirective" which has pretty similar body.

szkl:
Updated.

cihangir (approved):

PR #11441, opened 2017-10-02
klient: improve opening boltdb / fix wrong team default

rjeczalik:
This PR:

- fixes "invalid team name or user does not belong to the team", wrong default
- fixes "bad file descriptor" error, which is a regression; previously we ensured in kd's main that boltdb file exists; this is no longer true, that's why function which opens BoltDB now is responsible for ensuring the file exists, when database is being open in read mode.

### DIFF
--- a/go/src/koding/kites/config/cache.go
+++ b/go/src/koding/kites/config/cache.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"os/user"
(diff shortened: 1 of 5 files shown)


cihangir (approved):

cihangir:
@rjeczalik Could  you please rebase and push again? tests are failing and restarting did not help :/

rjeczalik:
@cihangir Done.

PR #11447, opened 2017-10-05
Start kite server immediately

szkl:


go/src/koding/kites/kloud/main.go
```diff
@@ -58,6 +58,8 @@ func main() {
 		log.Fatal(err)
 	}
 
+	k.Kite.Run()
```
rjeczalik:
```
go k.Kite.Run()
```

This is blocking operation, so it needs to be started on a separate gouroutine.

rjeczalik (approved):

PR #11466, opened 2017-10-19
Configure with inheriting options flags

mertaytore:
Options flags can now change default values of credentials values for dev environment.

Tests and necessary additions to circleci are also included.

### DIFF
--- a/config/inheritOptionFlags.test.coffee
+++ b/config/inheritOptionFlags.test.coffee
@@ -1,0 +1,83 @@
+traverse = require 'traverse'
+inherit = require './inheritOptionFlags'
+
+{ expect }  = require '../workers/social/testhelper'
+
+test_options =
+  level1a       :
+    level2a     : 'test-level2a-value'
+    level2b     :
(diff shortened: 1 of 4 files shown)


config/inheritOptionFlags.test.coffee
```diff
@@ -0,0 +1,68 @@
+traverse = require 'traverse'
+inherit = require './inheritOptionFlags'
+
+{ expect }  = require '../workers/social/testhelper'
+
+test_options =
+  level1a       :
+    level2a     : 'test-level2a-value'
+    level2b     :
+      level3a   : 'test-level3a-value'
```
cihangir: Could you make this `level3b` with that way we can ensure that we preserve the existing ones while adding a new property in "level 3" items.

cihangir (changes requested):
just one small test case

szkl (approved):

cihangir (approved):

cihangir:
👍

PR #11472, opened 2017-10-25
NSE: remove Safari workaround since the issue fixed on upstream

gokmen:
Flex usage fixed in Safari 11 which is the stable version of macOS now
https://bugs.webkit.org/show_bug.cgi?id=137730

screen shot 2017-10-25 at 15 39 51

## Types of changes


sinan (approved):
👍

